PR #20, opened 2025-07-07
Add warning to readme

jmecom:
As is, bitchat may not meet the stated security goals. For example, to achieve true PFS, it should use Signal-style X3DH + double ratchet, or [MLS](https://datatracker.ietf.org/doc/rfc9420/). Broadly, it would be good to switch to one of these standard, vetted approaches to secure messaging. However, for now, I propose adding a disclaimer to help ensure people don't use this app in serious settings before it's ready. 

update (jul 8): here's a writeup regarding security https://www.supernetworks.org/pages/blog/agentic-insecurity-vibes-on-bitchat#vibe-identity-an-obvious-flaw

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 ![ChatGPT Image Jul 5, 2025 at 06_07_31 PM](https://github.com/user-attachments/assets/2660f828-49c7-444d-beca-d8b01854667a)
 # bitchat
 
+> [!WARNING]
+> This software has not received external security review and may contain vulnerabilities and does not necessarily meet its stated security goals. Do not use it for production use, and do not rely on its security whatsoever until it has been reviewed.
+
 A secure, decentralized, peer-to-peer messaging app that works over Bluetooth mesh networks. No internet required, no servers, no phone numbers - just pure encrypted communication.
 
 ## License


tomleavy:
The topic of decentralized MLS has come up in various IETF meetings and currently with the existing standard you need a single node to set ordering for commits in order to avoid forks in the state that would break things. There are several people (including myself) that are interested in solving for this better.

jmecom:
> The topic of decentralized MLS has come up in various IETF meetings and currently with the existing standard you need a single node to set ordering for commits in order to avoid forks in the state that would break things. There are several people (including myself) that are interested in solving for this better.

Ah, great point. I was too hasty with calling out specific fixes. Thanks! (I think the commit stays as-is.)

tomleavy:
> > The topic of decentralized MLS has come up in various IETF meetings and currently with the existing standard you need a single node to set ordering for commits in order to avoid forks in the state that would break things. There are several people (including myself) that are interested in solving for this better.
> 
> 
> 
> Ah, great point. I was too hasty with calling out specific fixes. Thanks! (I think the commit stays as-is.)

+1 commit is good. Mostly raising awareness to @jackjackbits that MLS is great for a bunch of reasons and that extending it for decentralized comms is possible but there is no standard way to do so yet.

lebed2045:
> with the existing standard you need a **single node** to set ordering for commits in order to avoid forks in the state that would break things

Single node or single source of truth? Hypothesis: Avalanche/Snow-family consensus claims to give a total order without any leader by gossip-and-vote. In theory that should let every MLS member receive the same Proposal → Commit sequence—no central DS needed.

Has anyone tried to graft Snow/Snowman (or Avalanche’s collapsible-state variant) under MLS yet? I’m wondering whether the probabilistic guarantees are strong enough for the tree-hash/epoch invariants, or if we still need something like the “Distributed-MLS” draft that avoids global ordering altogether. Pointers to proofs or prototypes welcome!

tomleavy:
> Single node or single source of truth? Hypothesis: Avalanche/Snow-family consensus claims to give a total order without any leader by gossip-and-vote. In theory that should let every MLS member receive the same Proposal → Commit sequence—no central DS needed.

Section 5 of the [mls architecture](https://datatracker.ietf.org/doc/rfc9750/) describes the details of DS requirements. There is a high level description of an eventually consistent DS for decentralized systems but there is no standard that specifies exactly how it would work.

> 
> 
> Has anyone tried to graft Snow/Snowman (or Avalanche’s collapsible-state variant) under MLS yet? I’m wondering whether the probabilistic guarantees are strong enough for the tree-hash/epoch invariants, or if we still need something like the “Distributed-MLS” draft that avoids global ordering altogether. Pointers to proofs or prototypes welcome!

I suspect there is just a balance of latency vs deep protocol customization. Baking fork resilience in at a low level would make the transport simple and less latent. Alternatively you can use another protocol to get a confirmed order for commits which will require voting or other mechanisms that will potentially add latency.

sabbylpu1:
dayum why is it a open source and is it acceesible to all?

jackjackbits:
> > > The topic of decentralized MLS has come up in various IETF meetings and currently with the existing standard you need a single node to set ordering for commits in order to avoid forks in the state that would break things. There are several people (including myself) that are interested in solving for this better.
> > 
> > 
> > Ah, great point. I was too hasty with calling out specific fixes. Thanks! (I think the commit stays as-is.)
> 
> +1 commit is good. Mostly raising awareness to @jackjackbits that MLS is great for a bunch of reasons and that extending it for decentralized comms is possible but there is no standard way to do so yet.

great to know. thank you. i went with noise protocol for now.

PR #21, opened 2025-07-07
Voice messaging feature

batinay1337:
- Introduced voice message functionality, allowing users to send and receive audio messages.
- Implemented AudioService for recording, playback, and audio data management.
- Updated BitchatMessage structure to include audio data, duration, and voice message flags.
- Enhanced BluetoothMeshService to handle sending private and broadcast voice messages.
- Integrated voice message UI components in ChatViewModel and Views, including VoiceMessageView and VoiceMessageBubble.
- Added microphone usage description in Info.plist for permission handling.
- Centralized theme management for consistent UI across voice message components.


jackjackbits:
we may want to wait on this feature for wifidirect so we don't have to fragment.

PR #34, opened 2025-07-08
Fix message duplication and improve sync performance in mesh network

nelsoncampoos:
## fixing duplicate messages when reconnecting

to reproduce the bug:
1. connect device a and device b
2. turn off bluetooth on device b
3. send messages from device b (offline) while a is online
4. turn bluetooth back on device b
5. messages appear duplicated on device a (and last message shows wrong sender)

## root cause
when no peers were connected, the retry queue was storing ALL messages (including those from other devices). when device b reconnected, it would resend device a's messages as if they came from device b.

## fix
- retry queue now only stores messages from the current device
- added proper message deduplication by id and content/timestamp
- added thread safety to prevent crashes from concurrent access

## performance improvements
while fixing this, also optimized sync speed:
- retry interval: 5s → 2s
- cached message delay: 100ms → 20ms  
- announce delay: 1s → 0.2s
- overall sync time: ~5s → ~1s


jackjackbits:
please resolve conflicts and resubmit. thank you!

nelsoncampoos:
done @jackjackbits, thanks for the project, happy to contribute 

https://github.com/user-attachments/assets/601ca880-ca68-436f-b29b-b85994b3c461

owenselles:
Confirmed on my side as well. Issue is gone

jackjackbits:
thank you so much!

PR #35, opened 2025-07-08
Fix for send message bug and other new features

utsavDave97:
## New features integrated
- Send message button greyed out if textfield empty
- Shows confirmation alert before leaving channel (I accidentally tapped on leave whereas I wanted to tap on lock icon for password so it is good to show a confirmation)
- Currently, message textfield has auto correction enabled which has this weird bug (See video below)

# Autocorrection Bug (Before fix)
https://github.com/user-attachments/assets/24d573b8-42a6-4d30-ab9d-17307eb91166

# Autocorrection Bug (After fix)
https://github.com/user-attachments/assets/b8681aaa-5a60-431b-92ac-daac7513032e

# Confirmation Alert for leave channel

# Send button greyed out if no message typed

### DIFF
--- a/bitchat/Views/ContentView.swift
+++ b/bitchat/Views/ContentView.swift
@@ -26,6 +26,7 @@ struct ContentView: View {
     @State private var showPasswordError = false
     @State private var showCommandSuggestions = false
     @State private var commandSuggestions: [String] = []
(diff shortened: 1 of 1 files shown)


jackjackbits:
thank you so much!

PR #38, opened 2025-07-08
Migrate to Observable Macro

lawgimenez:
Hi! Please check if migrating to Observable Macro is acceptable.

- I set the minimum deployment target to 17.6
- Removed `objectWillChange` since we don't need this one


daghsnn:
Nice and clean with new macro

PR #41, opened 2025-07-08
Disambiguate duplicate nicknames on the same network

dasmer:
**Problem**
On large mesh networks, it's common for multiple users to have the same nickname. This creates ambiguity—especially when trying to address someone directly using @mentions.

**Solution**
To resolve this, when two users share the same nickname, we append the first four characters of their unique fingerprint to disambiguate them. For example, `@dasmer` becomes `@dasmer-d4da`. The local user's nickname remains unmodified.

If all nicknames are unique, we display them as-is, without any additional suffix.

This PR includes unit tests to ensure functionality continues to work as intended. See screenshot below of a case where I set the nickname of two devices on the same network to be `@dasmer`.

image


dasmer:
Closing this one since it has some bugs. @jackjackbits if this one is interesting to you and one you would merge to disambiguate, let me know and I can come back to make fixes.

PR #43, opened 2025-07-08
Main

drQedwards:
so we have nodes. not the weighted kind, just the npm kind. we want to just xcodegen our packages, and I do not want to do the cloning of 
https://github.com/drqedwards/ppm

it won't break anything it's just extra steps if you really want to be extra and python your swift while we c program the pin intergration inside the applicaiton in our PR. 

Again, pin is showing amazing results on meta messenger and like it's so much better and faster than 2FA authentication and also you want to go on a date late witth?

https://www.facebook.com/josef.edwards.7/

but like later anyway.


drQedwards:
erm yeah like now I get the vibe coding thing with you guys, and that got really... well, I am wanting more lol

drQedwards:
btw

https://dn.com/en-us/sale/bitchat.xyz

@jackjackbits this domain is available if we want to just go ahead do the .xyz website app version of our seift app. like she;'s great on the apple application site, but like  we all know IRC used to be website ventrilo sort of windows speaking of vim 1990's vibes

drQedwards:
oh @jackjackbits  you need to hit the actions run thing if we wasnt to finish this up lol.  the application chat works fine and now pins work but we can also make her become https://bitchat.xyz

also yes we need to move omn with our lives and hang out irl now pls and ty

PR #44, opened 2025-07-08
Fix: Ensure peer list updates on main thread

PubliusPseudis:
This pull request addresses a potential race condition that could cause the app to crash or behave unexpectedly. I have added a dispatch to the main queue to ensure that all UI-related updates are performed on the main thread, which prevents crashes and unexpected UI behavior.


jackjackbits:
thank you!

PR #59, opened 2025-07-08
PIN and PPM commands inside the bitchat enviroment for signage  PR #43

drQedwards:


owenselles:
Decline, use a .gitignore :)

jackjackbits:
not sure what all this does.

PR #63, opened 2025-07-08
Fix Unsupported markdown: blockquote in README.md

richardsonnick:
Fixes markdown issue where ">60" appears as "Unsupported markdown: blockquote".


jackjackbits:
thank you so much!

PR #75, opened 2025-07-08
Fixing macOS build

nanoterran:
Includes guard around `LPLinkMetadata` which only should work for iOS

Error Message when building for macOS:
```
~/bitchat/bitchat/Views/LinkPreviewView.swift:19:34 Cannot find type 'LPLinkMetadata' in scope
```

Change:
```
#if os(iOS)
  @State private var metadata: LPLinkMetadata?
#endif
```


jackjackbits:
thank you!

PR #84, opened 2025-07-09
Add bluetooth entitlement for macOS

MitchZinck:
Adds `com.apple.security.device.bluetooth` entitlement to enable Bluetooth connectivity for MacOS.

**Changes**:
```xml
<key>com.apple.security.device.bluetooth</key>
<true/>
```

### DIFF
--- a/bitchat/bitchat.entitlements
+++ b/bitchat/bitchat.entitlements
@@ -8,5 +8,7 @@
 	<array>
 		<string>group.chat.bitchat</string>
 	</array>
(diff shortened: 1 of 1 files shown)


jackjackbits:
i should have fixed this in recent branch. pls confirm it works for you.

MitchZinck:
@jackjackbits - I had to delete the launch screen .storyboard, as it isn't compatible with mac, but otherwise yes this works!

PR #96, opened 2025-07-09
Fix: Prevent potential buffer overflow in CompressionUtil.swift

Mirza-Samad-Ahmed-Baig:
This PR addresses a potential buffer overflow vulnerability in the compress function within CompressionUtil.swift. The original implementation allocated the destination buffer for compressed data with the size of the uncompressed data. In certain edge cases, compressed data can be larger than the original, leading to a buffer overflow. This fix allocates a larger, safer buffer for the compressed data using a common LZ4 worst-case heuristic (source_size + source_size / 255 + 16). This ensures that the buffer is always sufficiently large to prevent overflows, improving the stability and security of the compression utility.


jackjackbits:
thank you so much!

PR #121, opened 2025-07-09
fix: remove macOS/visionOS designed for iPad destination on bitchat_iOS target

edeuss:
From my research BLE is severely restricted also on visionOS.


Maldini9:
May I run a tester application?

PR #122, opened 2025-07-09
fix Dropped Messages on Nickname Conflicts

approxit1:
This PR fixes a bug where messages were getting dropped if two people had the same nickname. The code was checking the sender by nickname, so it would pick up the other person's message and consider that to be a duplicate of your own and ignore it. I have changed it to use the peer ID rather than the nickname to check who is the sender. Now it is working fine and messages are no longer lost.


edeuss:
Nice catch 👍

b-Istiak-s (approved):

PR #201, opened 2025-07-12
feat: Add alert for emergency wipe of all data

umangparnami:
resolves: #152
- Added a confirmation alert for clearing all data on triple-tap of logo.
- Also disabled auto capitalization and auto correction for the nickname `TextField`


xominf:
I feel like the point of an emergency data wipe should be that it's discreet.

umangparnami:
> I feel like the point of an emergency data wipe should be that it's discreet.

@Thewildcatofideas For preventing accidental clearing of data, this was added.

PR #219, opened 2025-07-12
Satellite networking extension for BitChat

vikasvardhanv:
SatMesh is a satellite networking extension for BitChat that enables global communication beyond the local Bluetooth mesh network. It provides emergency broadcasting, global messaging, and hybrid routing capabilities using satellite IoT networks.

### DIFF
--- a/bitchat/Services/EmergencyBroadcastSystem.swift
+++ b/bitchat/Services/EmergencyBroadcastSystem.swift
@@ -1,0 +1,756 @@
+//
+// EmergencyBroadcastSystem.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Foundation
+import CoreLocation
+import Combine
+
+// Emergency types
+enum EmergencyType: String, Codable, CaseIterable {
+    case sos = "sos"
+    case medical = "medical"
+    case disaster = "disaster"
+    case security = "security"
+    case weather = "weather"
+    case fire = "fire"
+    case flood = "flood"
+    case earthquake = "earthquake"
+    case tsunami = "tsunami"
+    case civilUnrest = "civil_unrest"
+    
+    var displayName: String {
+        switch self {
+        case .sos: return "SOS"
+        case .medical: return "Medical Emergency"
+        case .disaster: return "Disaster"
+        case .security: return "Security Threat"
+        case .weather: return "Weather Emergency"
+        case .fire: return "Fire"
+        case .flood: return "Flood"
+        case .earthquake: return "Earthquake"
+        case .tsunami: return "Tsunami"
+        case .civilUnrest: return "Civil Unrest"
+        }
+    }
+    
+    var priority: Int {
+        switch self {
+        case .sos: return 0
+        case .medical: return 1
+        case .security: return 2
+        case .disaster, .fire, .flood, .earthquake, .tsunami: return 3
+        case .weather: return 4
+        case .civilUnrest: return 5
+        }
+    }
+    
+    var icon: String {
+        switch self {
+        case .sos: return "🚨"
+        case .medical: return "🏥"
+        case .disaster: return "💥"
+        case .security: return "🔒"
+        case .weather: return "🌪️"
+        case .fire: return "🔥"
+        case .flood: return "🌊"
+        case .earthquake: return "🌋"
+        case .tsunami: return "🌊"
+        case .civilUnrest: return "⚔️"
+        }
+    }
+}
+
+// Emergency message structure
+struct EmergencyMessage: Codable, Identifiable {
+    let id: String
+    let emergencyType: EmergencyType
+    let senderID: String
+    let senderNickname: String
+    let content: String
+    let location: LocationData?
+    let timestamp: Date
+    let expiresAt: Date
+    let priority: Int
+    let isAcknowledged: Bool
+    let acknowledgmentCount: Int
+    let responders: [EmergencyResponder]
+    let status: EmergencyStatus
+    
+    enum EmergencyStatus: String, Codable {
+        case active = "active"
+        case acknowledged = "acknowledged"
+        case responding = "responding"
+        case resolved = "resolved"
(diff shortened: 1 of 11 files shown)


ayushpattnaik:
how would this work? explain in simple words plz

vikasvardhanv:
> how would this work? explain in simple words plz

Satellite Mesh is an upgrade to BitChat that lets BitChat send messages not just over Bluetooth (to people nearby), but also through satellites so we can reach people anywhere in the world, even if there’s no cell service or WiFi.

**How Does It Work?**
**1. Normal BitChat (Before)**
Messages go from phone to phone using Bluetooth.
Only works if people are nearby.

**2. With SatMesh (Now)**
If someone is far away, or in an emergency, your message can be sent up to a satellite.
The satellite beams it down to another part of the world, where it reaches the person (or emergency services).
The app decides the best way to send each message: Bluetooth (if possible), or satellite (if needed).

A. Satellite Communication
SatelliteProtocolAdapter.swift: Talks to the satellite hardware. It knows how to send and receive messages via satellite, and handles special emergency messages.
B. Smart Routing
MultiPathRoutingEngine.swift: Decides the best path for message. Should it go over Bluetooth? Satellite? Both? It picks the fastest, cheapest, or most reliable route.
C. Bandwidth Optimization
BandwidthOptimizer.swift: Compresses messages so they use less satellite data (which is expensive and slow). Also avoids sending the same message twice.
D. Message Queuing
SatelliteQueueService.swift: If a satellite isn’t available right now, it stores message and sends it when the satellite passes by. Emergency messages go first.
E. Emergency System
EmergencyBroadcastSystem.swift: Lets us send SOS or other emergency alerts. These get top priority and are sent via satellite if needed, including your location.
F. Integration Service
SatMeshIntegrationService.swift: The “manager” that connects all the above pieces. The rest of the app talks to this service to use SatMesh features.
G. User Interface
SatMeshViewModel.swift and SatMeshViews.swift: Adds new screens and buttons to the app for:
Sending emergency messages
Checking satellite status
Sending/reading global messages
Configuring SatMesh settings


**How to Use It?**

Normal Chat: Still works as before, using Bluetooth.
Global Message: You can send a message to anyone, anywhere, even if they’re not nearby.
Emergency: Tap the SOS button to send an emergency alert via satellite, including your location.
Settings: You can adjust how much satellite data to use, set emergency contacts, and more.
In Short:
SatMesh makes BitChat work anywhere, not just nearby.
It’s smart about when to use satellite (to save money and battery).
It’s built for emergencies, remote areas, and privacy.

bitchat/Services/SatelliteProtocolAdapter.swift
```diff
@@ -0,0 +1,551 @@
...
+    
+    func connect() {
+        // Placeholder for connection logic
+        isConnected = true
+    }
+    
+    func disconnect() {
+        // Placeholder for disconnection logic
+        isConnected = false
+    }
```
TheHunterDog, line 534: I really like the idea and was curious to the implementation

vikasvardhanv: We can touch base and discuss more about this and take it from there.

TheHunterDog: I did look into it but it turns out apple doesn't offer such thing to use the satellite communications. I was wondering how you wanted to implement the Satellite functionality?

TheHunterDog: @vikasvardhanv do you plan to implement the satellite functionality?

bitchat/Services/EmergencyBroadcastSystem.swift
```diff
@@ -0,0 +1,756 @@
...
+            lastEmergency: emergencyHistory.last?.timestamp,
+            emergencyTypes: emergencyTypes
+        )
+        
+        delegate?.emergencyBroadcastStatsUpdated(emergencyStats)
+    }
+    
+    private func calculateCoverageArea() -> Double {
+        // Simplified coverage area calculation
+        // In a real implementation, this would calculate the actual coverage area
```
malted, line 633: Was this entire PR vibe-coded?

bitchat/Services/SatelliteProtocolAdapter.swift
vikasvardhanv: yes, would like to implement it.

bitchat/Services/EmergencyBroadcastSystem.swift
vikasvardhanv: PR is created as I'm looking to collaborate and complete the todo and placeholders added.

natarajan0007:
@vikasvardhanv The concept is independent of any specific company or resource. Currently, individual users own this bluetooth network broadcast via mobile through enabling Bluetooth connections. However, for broader access, satellite networks are managed not by individuals but by enterprise companies using a Hub and Spoke model with various types of antennas and different spectrum band. I believe this implementation would never align with that setup and vision.

vikasvardhanv:
> @vikasvardhanv The concept is independent of any specific company or resource. Currently, individual users own this bluetooth network broadcast via mobile through enabling Bluetooth connections. However, for broader access, satellite networks are managed not by individuals but by enterprise companies using a Hub and Spoke model with various types of antennas and different spectrum band. I believe this implementation would never align with that setup and vision.

We could replace the Bluetooth transport layer with a satellite-compatible gateway model while keeping the higher-level BitChat protocol unchanged. Instead of phones broadcasting over short-range Bluetooth, selected gateway nodes (run by individuals or organizations) could connect to existing satellite networks via licensed modems and antennas. These gateways would act as “spokes” linking local BitChat meshes to enterprise-managed “hubs” in the satellite provider’s infrastructure. Using a store-and-forward, delay-tolerant approach, messages could be queued locally, transmitted in scheduled satellite uplinks, and delivered to other connected gateways. This preserves the decentralized logic while aligning with the hub-and-spoke control model, overcoming the Bluetooth range limitation and allowing global coverage without requiring individuals to own or operate satellites directly.

PR #271, opened 2025-07-20
Fix Crash when receiving notification from the background

ryannair05:
In my testing the app crashes in the background when receiving a notification. This is probably due to UIFeedbackGenerator requiring the main thread which isn't available


jackjackbits:
thank you!

PR #306, opened 2025-07-23
NoiseProtocol update: send nonce in packet

callebtc:
This PR addresses the nonce desynchronization issue by sending the nonce over the wire. This should get rid of all desynchronization issues. I've fixed a bug related to parsing packages which leads to noise re-keying even if the session is ok (see comments below)

This update is compatible with the android PR https://github.com/permissionlesstech/bitchat-android/pull/181

### Problem

The Noise protocol transport ciphers (`sendCipher` and `receiveCipher`) can become desynchronized when messages are lost, reordered, or when there are race conditions in concurrent message processing. When ciphers go out of sync, all subsequent encrypted messages fail to decrypt, requiring session renegotiation to restore communication.

This synchronization issue is particularly problematic in mesh networks where:
- Messages can arrive out of order due to multiple routing paths
- Network partitions can cause message loss
- Concurrent message processing can create race conditions with nonce counters

### Solution

This PR implements **WireGuard-style sliding window nonce synchronization** for Noise transport encryption, where each encrypted message includes its encryption nonce as an 4-byte prefix with 1024-message replay protection: `<4-byte nonce><ciphertext>`.

#### Wire Protocol Format

**Transport Message Structure (Reproducible):**
```
[4-byte nonce (big-endian)][ChaCha20-Poly1305 ciphertext + 16-byte MAC]
```

**Packet Details:**
- **Nonce**: 32-bit big-endian unsigned integer (sender's message counter)
- **Ciphertext**: ChaCha20 encrypted payload using extracted nonce
- **MAC**: Poly1305 authentication tag covering entire payload
- **Overhead**: +4 bytes per transport message
- **Handshake**: Unchanged, maintains full Noise protocol compliance

#### Key Changes

**1. Enhanced Cipher Classes (iOS & Android):**
- Added sliding window replay protection with 1024-message tolerance
- Added helper methods:
  - `isValidNonce()` - Validates nonce against sliding window (prevents replay attacks)
  - `markNonceAsSeen()` - Updates 128-byte bitmask after successful decryption
  - `extractNonceFromCiphertextPayload()` - Extracts 4-byte big-endian nonce
  - `nonceToBytes()` - Converts nonce to 4-byte big-endian representation
- Modified `encrypt()` to prepend 4-byte nonce for transport ciphers
- Modified `decrypt()` to extract, validate, and use nonce with replay protection

**2. Security Implementation:**
- **32-bit Nonce Space**: 2^32 messages (~585 years at 1M msg/s) eliminates wraparound
- **Sliding Window**: 1024-message tolerance using 128-byte bitmask for out-of-order delivery
- **Replay Protection**: Rejects duplicate nonces and messages outside window
- **Thread Safety**: Synchronized cipher operations prevent nonce corruption

**3. Cross-Platform Compatibility:**
- **iOS**: Enhanced `NoiseCipherState` in `NoiseProtocol.swift`
- **Android**: Identical implementation in `NoiseSession.kt`
- **Wire Format**: Identical 4-byte big-endian nonce encoding
- **Constants**: Unified `NONCE_SIZE_BYTES=4`, `REPLAY_WINDOW_SIZE=1024`

#### Synchronization Logic
```swift
// Extract and validate nonce with sliding window
guard let (extractedNonce, actualCiphertext) = try extractNonceFromCiphertextPayload(ciphertext) else {
    throw NoiseError.invalidCiphertext
}

// WireGuard-style replay protection
guard isValidNonce(extractedNonce) else {
    throw NoiseError.replayDetected
}

// Use extracted nonce for decryption (enables synchronization)
nonce = extractedNonce
let plaintext = try decrypt(actualCiphertext)

// Mark nonce as seen after successful decryption
markNonceAsSeen(extractedNonce)
```

### Security Analysis

**Addresses Critical Vulnerabilities:**
- ✅ **Nonce Exhaustion**: 16-bit → 32-bit space (65k → sqrt(18 quintillion) messages)
- ✅ **Replay Attacks**: WireGuard-style sliding window prevents duplicate acceptance
- ✅ **Out-of-Order Tolerance**: 1024-message reordering vs strict sequential requirement
- ✅ **Race Conditions**: Thread-safe operations prevent cipher state corruption

**Cryptographic Properties Maintained:**
- ChaCha20-Poly1305 AEAD security with authenticated nonce inclusion
- Perfect forward secrecy through Noise XX handshake unchanged
- Noise protocol compliance (transport phase only modification)

### Benefits

1. **Perfect Synchronization:** Sender's nonce explicit in each message, eliminating desync
2. **Robust Reordering:** Handles up to 1023 out-of-order messages gracefully
3. **Automatic Recovery:** Single newer message resynchronizes both peers
4. **Replay Protection:** Sliding window prevents duplicate message acceptance
5. **Cross-Platform:** iOS ↔ Android compatibility with identical wire format

### Technical Implementation

- **Phase Separation:** Handshake ciphers unchanged, transport ciphers use nonce-in-payload
- **Memory Efficient:** 128-byte sliding window per cipher (1024 bits)
- **Performance:** O(1) nonce validation and window updates
- **Thread Safety:** Dedicated cipher locks prevent concurrent corruption
- **Error Handling:** Comprehensive logging with replay detection alerts

### Testing & Validation

- ✅ **Build Success**: iOS (Xcode) and Android (Gradle) compilation
- ✅ **Wire Compatibility**: Identical packet format across platforms  
- ✅ **Protocol Compliance**: Handshake phase completely unchanged
- ⚠️ **Test Updates**: Legacy tests require updating for new format (expected)

---

**Breaking Change:** This modifies the transport message format. All communicating peers must use the same version for compatibility.

### DIFF
--- a/bitchat/Services/BluetoothMeshService.swift
+++ b/bitchat/Services/BluetoothMeshService.swift
@@ -2869,9 +2869,20 @@ class BluetoothMeshService: NSObject {
         case .noiseEncrypted:
             // Handle Noise encrypted message
             let senderID = packet.senderID.hexEncodedString()
-            if !isPeerIDOurs(senderID) {
+            if let recipientIDData = packet.recipientID,
+               isPeerIDOurs(recipientIDData.hexEncodedString())
+               && !isPeerIDOurs(senderID) {
                 _ = packet.recipientID?.hexEncodedString()
                 handleNoiseEncryptedMessage(from: senderID, encryptedData: packet.payload, originalPacket: packet)
+                return
+            }
+            if !isPeerIDOurs(senderID) {
+                if packet.ttl > 0 {
+                    var relayPacket = packet
+                    relayPacket.ttl -= 1
+                    broadcastPacket(relayPacket)
+                }
+                return
             }
             
         case .versionHello:
@@ -2898,7 +2909,9 @@ class BluetoothMeshService: NSObject {
         case .protocolNack:
             // Handle protocol-level negative acknowledgment
             let senderID = packet.senderID.hexEncodedString()
-            if !isPeerIDOurs(senderID) {
+            if let recipientIDData = packet.recipientID,
+               isPeerIDOurs(recipientIDData.hexEncodedString())
+               && !isPeerIDOurs(senderID) {
                 handleProtocolNack(from: senderID, data: packet.payload)
             }
             
@@ -4438,9 +4451,9 @@ extension BluetoothMeshService: CBPeripheralManagerDelegate {
             } else {
                 SecureLogger.log("Have session with \(peerID) but decryption failed", category: SecureLogger.encryption, level: .warning)
                 
-                // Send a NACK to inform peer their session is out of sync
-                sendProtocolNack(for: originalPacket, to: peerID, 
-                               reason: "Decryption failed - session out of sync", 
+                // Send a NACK to inform peer that decryption failed
+                sendProtocolNack(for: originalPacket, to: peerID,
+                               reason: "Decryption failed",
                                errorCode: .decryptionFailed)
                 
                 // The NACK handler will take care of clearing sessions and re-establishing
--- a/bitchat/Noise/NoiseProtocol.swift
+++ b/bitchat/Noise/NoiseProtocol.swift
@@ -53,11 +53,13 @@ struct NoiseProtocolName {
 class NoiseCipherState {
     private var key: SymmetricKey?
     private var nonce: UInt64 = 0
+    private var useExtractedNonce: Bool = false
     
     init() {}
     
-    init(key: SymmetricKey) {
+    init(key: SymmetricKey, useExtractedNonce: Bool = false) {
         self.key = key
+        self.useExtractedNonce = useExtractedNonce
     }
     
     func initializeKey(_ key: SymmetricKey) {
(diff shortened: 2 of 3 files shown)


jackjackbits:
did you test with most recent branch (which includes ACK/NACK support and lots of re-handshake on failure cases)?

callebtc:
> did you test with most recent branch (which includes ACK/NACK support and lots of re-handshake on failure cases)?

yup. will confirm again.

bitchat/Services/BluetoothMeshService.swift
```diff
@@ -2869,9 +2869,20 @@ class BluetoothMeshService: NSObject {
         case .noiseEncrypted:
             // Handle Noise encrypted message
             let senderID = packet.senderID.hexEncodedString()
-            if !isPeerIDOurs(senderID) {
+            if let recipientIDData = packet.recipientID,
+               isPeerIDOurs(recipientIDData.hexEncodedString())
+               && !isPeerIDOurs(senderID) {
```
callebtc, line 3278: We (Alice) tried to decrypt every package, even if the package is from Bob to Carol. We then send a NACK to Bob, tearing down our intact noise session with Bob.

This should fix it but I strongly suggest to filter out for our packets in a centralized place instead of in each handler, and also decide to relay packets with `broadcastPacket(relayPacket)` in a centralized place too.

Some of the remaining handlers still have this issue.

bitchat/Services/BluetoothMeshService.swift
```diff
@@ -2898,7 +2909,9 @@ class BluetoothMeshService: NSObject {
         case .protocolNack:
             // Handle protocol-level negative acknowledgment
             let senderID = packet.senderID.hexEncodedString()
-            if !isPeerIDOurs(senderID) {
+            if let recipientIDData = packet.recipientID,
+               isPeerIDOurs(recipientIDData.hexEncodedString())
+               && !isPeerIDOurs(senderID) {
```
callebtc, line 3318: Related: we (Alice) react to NACKs from Bob, even if they are addressed to Carol and tear down the session.

PR #308, opened 2025-07-23
Fix Bluetooth disconnect messages and blank chat rows

jackjackbits:
## Summary
- Fixed missing disconnect/connect messages when peers walk away from each other
- Fixed blank message rows appearing in chat when peers disconnect
- Improved peripheral ID to peer ID mapping to handle all message types

## Changes
1. **BluetoothMeshService.swift**
   - Updated `handleReceivedPacket` to establish peripheral mapping for all message types including handshakes
   - Enhanced disconnect handler with better peer ID lookup and detailed logging
   - Fixed API MISUSE warnings by checking connection state before writes

2. **ChatViewModel.swift**
   - Fixed blank system messages by ensuring display names are never empty
   - Added validation to prevent empty nickname resolution

## Test plan
- [x] Tested walking away from peer - disconnect message now appears correctly
- [x] Tested walking back into range - connect message appears correctly
- [x] Verified no blank rows appear when peers disconnect
- [x] Confirmed handshake completion still works properly
- [x] Verified messages continue to work after reconnection

### DIFF
--- a/bitchat/Services/BluetoothMeshService.swift
+++ b/bitchat/Services/BluetoothMeshService.swift
@@ -1963,6 +1963,27 @@ class BluetoothMeshService: NSObject {
                 // Update peer availability
                 self.updatePeerAvailability(senderID)
                 
+                // IMPORTANT: Update peripheral mapping for ALL message types including handshakes
+                // This ensures disconnect messages work even if peer disconnects right after handshake
+                if let peripheral = peripheral {
+                    let peripheralID = peripheral.identifier.uuidString
+                    
+                    // Check if we need to update the mapping
+                    if peerIDByPeripheralID[peripheralID] != senderID {
+                        SecureLogger.log("Updating peripheral mapping: \(peripheralID) -> \(senderID)", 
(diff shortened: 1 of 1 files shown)


bitchat/Services/BluetoothMeshService.swift
```diff
@@ -1757,6 +1963,27 @@ class BluetoothMeshService: NSObject {
                 // Update peer availability
                 self.updatePeerAvailability(senderID)
                 
+                // IMPORTANT: Update peripheral mapping for ALL message types including handshakes
```
GUVWAF, line 1966: I believe this is incorrect, since the `senderID` is the *original* source, but it does not necessarily correspond to the peripheral you received this from. Handshakes are not relayed, so only then the mapping can be trusted.

GUVWAF: With the targeted delivery from #319, it might send to the peripheral that relayed for the recipient, rather than the recipient itself. Since you don't know whether the relayer is still connected to the recipient (or possibly even all the hops on the route towards it), delivery may fail.

PR #314, opened 2025-07-24
Optimize BLE mesh network for robustness and range

jackjackbits:
## Summary
- Comprehensive optimizations to the BLE mesh network to improve reliability and range
- Fixed critical issues causing excessive packet retransmissions and network flooding
- Improved duplicate detection and message routing efficiency

## Key Changes
- **Fixed relay probability**: 2-node networks now use 0% relay (no flooding needed between just 2 peers)
- **Protocol ACKs**: Added acknowledgment system to prevent unnecessary retransmissions
- **Enhanced duplicate detection**: Implemented MessageState tracking with seen counts
- **Exponential backoff**: Added collision avoidance for retransmissions
- **Packet ID generation**: Fixed to use only immutable fields (senderID, sequenceNumber, timestamp, type)
- **Smart rate limiting**: Progressive throttling based on message frequency
- **Bidirectional connection fix**: Prevented duplicate sends when devices connect as both central and peripheral
- **Logging cleanup**: Removed verbose debug logging, fixed build warnings

## Test plan
- [x] Verified 2 idle clients no longer flood the network
- [x] Confirmed packets are sent only once (not 8 times as before)
- [x] Build completes without warnings
- [x] ACK system properly tracks and acknowledges encrypted messages
- [x] Rate limiting allows bursts but prevents sustained flooding

### DIFF
--- a/bitchat/Protocols/BinaryProtocol.swift
+++ b/bitchat/Protocols/BinaryProtocol.swift
@@ -19,11 +19,12 @@ extension Data {
 }
 
 // Binary Protocol Format:
-// Header (Fixed 13 bytes):
+// Header (Fixed 17 bytes):
 // - Version: 1 byte
 // - Type: 1 byte  
 // - TTL: 1 byte
 // - Timestamp: 8 bytes (UInt64)
+// - SequenceNumber: 4 bytes (UInt32)
 // - Flags: 1 byte (bit 0: hasRecipient, bit 1: hasSignature)
 // - PayloadLength: 2 bytes (UInt16)
 //
@@ -34,7 +35,7 @@ extension Data {
 // - Signature: 64 bytes (if hasSignature flag set)
 
 struct BinaryProtocol {
-    static let headerSize = 13
+    static let headerSize = 17
     static let senderIDSize = 8
     static let recipientIDSize = 8
     static let signatureSize = 64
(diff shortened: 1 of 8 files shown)


GUVWAF:
Interesting to see the RSSI-based relay probability (which is similar to Meshtastic) and 0% probability with only 2 nodes. I believe this can be taken a step further by skipping rebroadcasting back to the `senderID` and current relayer of the packet as I did for the Android app: https://github.com/permissionlesstech/bitchat-android/pull/57
Also, private messages can be targeted to a peripheral/central directly if they are connected, instead of flooding.

I’m not sure the additional delay for collision avoidance is really necessary for a connection-oriented BLE mesh. The initial delay based on NodeID already spreads it out, and the lower BLE stack then performs frequency hopping on a time schedule, hence “perfect synchronization” will not occur. (It would be different if the advertisement channel would be used like in the BLE Mesh standard).

jackjackbits:
> Interesting to see the RSSI-based relay probability (which is similar to Meshtastic) and 0% probability with only 2 nodes. I believe this can be taken a step further by skipping rebroadcasting back to the `senderID` and current relayer of the packet as I did for the Android app: [permissionlesstech/bitchat-android#57](https://github.com/permissionlesstech/bitchat-android/pull/57) Also, private messages can be targeted to a peripheral/central directly if they are connected, instead of flooding.
> 
> I’m not sure the additional delay for collision avoidance is really necessary for a connection-oriented BLE mesh. The initial delay based on NodeID already spreads it out, and the lower BLE stack then performs frequency hopping on a time schedule, hence “perfect synchronization” will not occur. (It would be different if the advertisement channel would be used like in the BLE Mesh standard).

great points. will look at it, and open to PRs always.

PR #316, opened 2025-07-24
Fix RSSI nil display issue

jackjackbits:
## Summary
- Fixed RSSI values showing as nil (hollow circles) in the UI when devices are connected
- Added mapping logic to handle announce packets that arrive via relay without peripheral reference
- Implemented periodic RSSI updates to keep signal strength values current

## Changes
1. **Peripheral mapping fix**: When announce packets arrive via relay, map single unmapped peripherals to the announcing peer
2. **RSSI transfer**: Transfer RSSI values from temporary UUID to real peer ID when mapping is established
3. **Periodic updates**: Added timer to refresh RSSI values every 10 seconds for all connected peripherals
4. **Improved lookup**: Enhanced getPeerRSSI() to check multiple sources including known peers without peripherals

## Test plan
- [x] Build succeeds without warnings
- [ ] RSSI values display correctly (filled circles) when devices are connected
- [ ] RSSI updates in real-time as devices move closer/farther apart
- [ ] No regression in peer connection/disconnection behavior

### DIFF
--- a/bitchat/Services/BluetoothMeshService.swift
+++ b/bitchat/Services/BluetoothMeshService.swift
@@ -1558,6 +1558,18 @@ class BluetoothMeshService: NSObject {
             SecureLogger.log("  peerRSSI[\(peerID)] = \(rssi)", category: SecureLogger.session, level: .debug)
         }
         
+        // Log connectedPeripherals state
+        SecureLogger.log("connectedPeripherals has \(connectedPeripherals.count) entries:", category: SecureLogger.session, level: .debug)
+        for (peerID, peripheral) in connectedPeripherals {
+            SecureLogger.log("  connectedPeripherals[\(peerID)] = \(peripheral.identifier.uuidString)", category: SecureLogger.session, level: .debug)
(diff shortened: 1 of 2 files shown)


bitchat/Services/BluetoothMeshService.swift
```diff
@@ -1548,9 +1554,53 @@ class BluetoothMeshService: NSObject {
...
+        
+        
+        // Log connectedPeripherals state
+        SecureLogger.log("connectedPeripherals has \(connectedPeripherals.count) entries:", category: SecureLogger.session, level: .debug)
+        for (peerID, peripheral) in connectedPeripherals {
+            SecureLogger.log("  connectedPeripherals[\(peerID)] = \(peripheral.identifier.uuidString)", category: SecureLogger.session, level: .debug)
+        }
+        
+        
+        // Also check peripheralRSSI for any connected peripherals
```
rajashravan, line 1567: thoughts on how we could reduce the levels of nesting here?

PR #335, opened 2025-07-28
fix: resolve timing attack vulnerability in Noise Protocol key val

Mirza-Samad-Ahmed-Baig:
SECURITY: Implement constant-time comparison for public key validation

- Replace vulnerable contains() method with constant-time comparison
- Prevent timing side-channel attacks on cryptographic operations
- Add comprehensive tests to verify constant-time behavior
- Ensure compliance with NIST guidelines for cryptographic implementations

The previous implementation could leak information through timing differences when validating Curve25519 public keys against known bad points. An attacker could measure response times to determine if their key was in the bad key list, potentially leading to more sophisticated attacks.

This fix implements a constant-time comparison function that always takes the same amount of time regardless of input, preventing timing-based information leakage.

### DIFF
--- a/SECURITY_FIX_TIMING_ATTACK.md
+++ b/SECURITY_FIX_TIMING_ATTACK.md
@@ -1,0 +1,141 @@
+# Security Fix: Timing Attack Vulnerability in Noise Protocol
+
+## Executive Summary
+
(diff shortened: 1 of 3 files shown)


callebtc:
Can you please elaborate which sensitive data is leaked with a non-constant timing attack here? What's the scenario? If I understand, this is for comparing *public* keys. What do we leak here except maybe for how long the bad key list is (which is public too)?

PR #234, opened 2025-07-13
macOS arrow key navigation for command suggestions

ignaciocervino:
Adds arrow key navigation for command suggestions on macOS 14.0+.
Changes:

Up/down arrows navigate suggestions
Enter selects highlighted command
Escape cancels suggestions
Visual highlighting for selected item

Resolves #233 

https://github.com/user-attachments/assets/a22aacc0-4be2-485e-9803-caeb6ca2676b

PR #336, opened 2025-07-28
Fix macOS App Store keyboard overlay

ignaciocervino:
As reported by @AndreiRegiani https://github.com/permissionlesstech/bitchat/pull/234#issuecomment-3127621052
There is a gray bar overlay appearing over the text input field when the keyboard is open on macOS App Store builds, but not when building from Xcode.
image

### Root Cause
  Currently investigating further. The issue occurs only in App Store builds, not in local development or sandbox testing.


qalandarov:
Did this also close #410 ?

MichelDiz:
why this is closed @jackjackbits ?

PR #344, opened 2025-07-29
Font added

isaozler:
image

This PR adds the [Google Font](https://fonts.google.com/specimen/Kode+Mono) `Kode Mono` font used in the Bitchat brand to the UI.

This change introduces the brand font to establish a more distinctive and memorable visual identity for Bitchat. The goal is to give the product a unique look and feel that sets it apart. The font has been applied to text elements, including headings and messages.

**Screenshot:**
image
image
image


jackjackbits:
thank you but i like current font.

PR #345, opened 2025-07-29
Dynamic type support

timonus:
I have very bad eye sight and the fonts currently used in bitchat are a bit hard to read. This PR implements basic dynamic type support, all the existing fonts are scaled up using the `.body` text style.

|Before|After|
|--|--|
|||
|||


jackjackbits:
thank you. how does this work with smaller phones?

timonus:
Things will either line wrap or be truncated on smaller phones. Text seems to wrap in the chat, other places it truncates.

jackjackbits:
let's resolve conflicts and i'll merge.

PR #346, opened 2025-07-29
Disable sharing images via extension.

timonus:
I noticed sharing images via bitchat's share extension seemed supported this morning, but when I tried it nothing happened. This PR removes the flag that enables sharing of images via the share extension.

This _may_ require an xcodegen change too, haven't used the tool before so not sure at the moment!


jackjackbits:
great catch. thank you.

Issue #348, opened 2025-07-29
Keep crash after large text


AminPlusPlus:
When you paste large text app crashes, there should be limited characters

mmatoszko:
Does the crash occur on the sender's side or the receiver's side? I attempted to reproduce the issue but was only able to crash the receiver.

shiv19:
iOS side crashes when you send a link like this
https://www.npmjs.com/package/@capacitor/text-zoom the characters probably don't exceed but the rich preview that it tries to generate makes it crash? Not sure.

jackjackbits:
Closing as a duplicate of #534. Please continue the discussion and any implementation work there.

PR #356, opened 2025-07-30
Fix crash when sending a URL containing @mention.

mmatoszko:
crashing url: https://www.npmjs.com/package/@capacitor/text-zoom
Screenshot:
crash

Evidence that the test passes:
evidence

It fixes an issue mentioned in a comment in #348.


zabojeb:
Tested on iOS 18, works for me

mmatoszko:
this URL does not crash anymore :)

PR #357, opened 2025-07-30
Add IRC-style “cover traffic” switch to BitChat interface

zabojeb:
**Closes:** TODO at [line 1358 of BluetoothMeshService.swift](https://github.com/permissionlesstech/bitchat/blob/a438c4681708891be260776839df67db1733c376/bitchat/Services/BluetoothMeshService.swift#L1358)

### What's new

**1. Cover Traffic Toggle**
A switch that, when enabled, injects dummy messages to help obscure real conversation patterns.

**2. Persistent Setting**
The toggle state is saved across app launches, so you don't need to re-enable it every time you open BitChat.

**3. Reusable IRCToggle Component**
A new IRC-inspired SwiftUI element that can be dropped in for future preferences, styled to match BitChat’s aesthetic. I think it looks quite stylish.

### Preview

image
image

image
image

**Tested on:**
- MacOS 26 Tahoe
- iOS 18.1.1

### DIFF
--- a/bitchat/Services/BluetoothMeshService.swift
+++ b/bitchat/Services/BluetoothMeshService.swift
@@ -488,6 +488,7 @@ class BluetoothMeshService: NSObject {
     private var coverTrafficTimer: Timer?
     private let coverTrafficPrefix = "☂DUMMY☂"  // Prefix to identify dummy messages after decryption
     private var lastCoverTrafficTime = Date()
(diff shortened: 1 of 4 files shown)


zabojeb:
Well, I think It makes no sense because of #389.
But this switch is really nice UI element which can be used in other settings.

PR #366, opened 2025-07-31
Adds swift-secp256k1 to project.yml

mmatoszko:
`just run` does not work right now because `swift-secp256k1` is not specified in `project.yml`. This PR fixes it


jackjackbits:
thank you!

Issue #370, opened 2025-07-31
Username Uniqueness: Potential for Impersonation and Usability Issues


priyanshukumar397:
### Summary
BitChat currently allows users to choose any nickname (username) without checking for global uniqueness. This means multiple users can have the same nickname, which may cause confusion and open doors to impersonation, especially since the system is decentralized and there is no central authority to manage usernames.

<img width="260" height="500" alt="Image" src="https://github.com/user-attachments/assets/9c420c43-812f-444c-aaf3-f2c2645ec506" />

### Why This Is an Issue
- **Impersonation Risk:** Any user can pick a nickname already used by someone else, making it easy to impersonate trusted contacts or create confusion in group chats.
- **User Confusion:** Users may not be able to distinguish between different peers with identical nicknames, especially in larger mesh networks.
- **Trust Problems:** The social trust layer is undermined if users rely on nicknames alone, rather than cryptographic fingerprints, to identify peers.

### Suggested Solution
**Introduce a non-intrusive, privacy-preserving unique identifier:**
- Assign each user a randomly generated, non-reversible tag or alias (e.g., a short hash or emoji code) that is displayed alongside their nickname in chats and peer lists.
- This tag should be derived from the user's cryptographic fingerprint (e.g., truncated SHA-256 of their Curve25519 public key) but not reveal any information about their actual identity or device.
- Display this identifier in the UI next to the nickname, so users can distinguish peers even if nicknames collide.
- Optionally, warn users when they add a new peer whose nickname matches an existing peer (but with a different tag/fingerprint), to encourage out-of-band verification.

### Additional Notes for Maintainers (Official BitChat Team)
- The issue is not a cryptographic vulnerability but a usability and social trust risk.
- Consider updating the onboarding flow and peer display logic to always show the unique tag next to nicknames.
- Document the reasoning so users understand why the tag exists and how to use it to verify contacts.

### References
See protocol whitepaper Section 3 and Section 4 for details on identity and trust layer.

**Label:** bug

priyanshukumar397:
**Is this similar to the pr you mentioned, as I believe they are a bit distinct from each yet similar in broader terms @kswzzl** 

**Or is this a duplicate of the mentioned pr above by you? I would appreciate it if you could clarify the status or your thoughts on this approach @jackjackbits @callebtc sir**

kswzzl:
[...]

PR #380, opened 2025-08-01
Assign custom contact names

kswzzl:
### Description
---
Bitchat seems to have been designed with the idea of personal/custom contact names ('petnames' in source code) that you can assign to users, but it was never fully implemented. This PR completes personal/custom contact names by filling in all missing GUI components and code required.

### Fixes
---
This fixes https://github.com/permissionlesstech/bitchat/issues/370 and at a minimum helps improve Bitchat identity which is a bit messy.

### Verification Evidence
---
My iphone is `kevin_ios`. My macbook starts out as `anon42ab`, and I assign the contact name `kevin_macos`.

Before adding contact name:

New GUI element in contact dialog to modify personal contact name:

DM screen now shows contact name at the top and in chat:

Custom contact name is correct in public chat too:

Custom contact name can also be modified by long-pressing contact in contacts list. I change the name to `kevin_xyz` just to demo:

Still looks right in public chat:

I change my claimed nickname to `anon21` on my macbook, which reflects on my iPhone but custom contact name still takes precedence:

### DIFF
--- a/bitchat/ViewModels/ChatViewModel.swift
+++ b/bitchat/ViewModels/ChatViewModel.swift
@@ -524,6 +524,59 @@ class ChatViewModel: ObservableObject, BitchatDelegate {
         
         return false
     }
+
+        // MARK: - Petname Management
+
+    func setPetname(peerID: String, petname: String?) {
+        // First try to get fingerprint from mesh service (supports peer ID rotation)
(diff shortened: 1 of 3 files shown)


kswzzl:
re-confirmed functionality in macOS/iOS after rebase

PR #412, opened 2025-08-07
[security] Prevent spoofable plaintext messages from sliding into private chats

lts-rad:
While implementing a [linux rust client for bitchat](https://github.com/spr-networks/bitchat-plugin), I found a security flaw with DMs.

The state machine on iOS will mistakenly accept plaintext messages to display. This allows an adversary to impersonate a user inside of a trusted chat. 

This code patch adds a decrypted: bool parameter to the packet handler to mark when a message actually came from inside of a trusted & decrypted noise message. 

image

Perhaps an alert should be presented, but for now, this will silently drop the message in the UI and log the spoofing attempt to the console.

Reproducability:
The bug can be tested with the plugin by sending a command with /plain to inject a plaintext DM.

### DIFF
--- a/bitchat/Services/BluetoothMeshService.swift
+++ b/bitchat/Services/BluetoothMeshService.swift
@@ -3374,7 +3374,7 @@ class BluetoothMeshService: NSObject {
         }
     }
     
-    private func handleReceivedPacket(_ packet: BitchatPacket, from peerID: String, peripheral: CBPeripheral? = nil) {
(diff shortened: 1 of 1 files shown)


lts-rad:
I have not verified android @callebtc  but the android code may also be affected, https://github.com/permissionlesstech/bitchat-android/issues/240

jackjackbits:
thank you! will confirm with @callebtc

jackjackbits:
can you please submit this? i had to rewind a bit.

lts-rad:
Can do, https://github.com/permissionlesstech/bitchat/pull/428

PR #420, opened 2025-08-08
Remove noisy @ symbol in mention notification

marionauta:
Right now the "mentioned" notification includes an extra `@` symbol at the start of the title. This makes the notification a bit crowded since there's an emoji as well. The private message notification starts with an emoji but no `@`.

This PR removes the `@` symbol and leaves just the emoji on the "mentioned" notification.

Here you can see the before (bottom) and the after (top):

IMG_7607


jackjackbits:
thank you!

PR #421, opened 2025-08-08
Allow unicode letter characters in nicknames

### LINKED ISSUE #343: Nickname formating bug
Some special characters like 'ç' when used in the nickname seems to interrupt the formating of the nickname itself so for example when mentioning someone the nickname is formated in yellow but goes back to green text after the special character

marionauta:
The mention regex was defined with `a-zA-Z` which only allows "ASCII" characters. Replace it with `\p{L}` ("Letter" unicode category) to match special caracters like `ç` or `á`.

Note: this PR addresses nicknames/usernames. The same could be done with channels/hashtags names but maybe you want to keep those ascii?

See: https://regexr.com/8gf57

before:
Screenshot 2025-08-08 at 19 06 34

after:
Screenshot 2025-08-08 at 19 05 23

fixes #343

### DIFF
--- a/bitchat/ViewModels/ChatViewModel.swift
+++ b/bitchat/ViewModels/ChatViewModel.swift
@@ -122,7 +122,7 @@ class ChatViewModel: ObservableObject, BitchatDelegate {
     // MARK: - Autocomplete Properties
     
     // Autocomplete optimization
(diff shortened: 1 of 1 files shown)


jackjackbits:
thank you!

jackjackbits:
please re-submit.

PR #447, opened 2025-08-17
BLE privacy: derive peerID from Noise fingerprint; remove BLE Local Name from advertising; verify announces against key-derived ID; auto-initiate Noise handshake when encrypted message arrives without session; drop rotating alias option.

jackjackbits:


illahnajih:
Dear Sir Jack Dorsey @jackjackbits , I Need Borrowing You're Money And I will do whatever I can for you please help me (i beg you I really need borrow you're money sir Jack)

PR #451, opened 2025-08-17
Feat/mesh robustness efficiency

jackjackbits:
final refactor and robustness work


Aristoteleskv:
Hi Jack, I'm Kivova (from Angola). I've been using Bitchat and really appreciate the concept. However, I've run into a few challenges and had an idea I wanted to share with you — maybe it's something you could consider implementing.

Since Bitchat works via Bluetooth mesh, wouldn't it be useful to add a feature that allows users to scan or discover nearby devices that are also running Bitchat? It could help users connect more easily in offline environments and improve the overall experience.  I would like to be an ambassador

PR #480, opened 2025-08-22
chat: fix blanking and gaps when switching geohashes; channel-aware r…

jackjackbits:
…ow IDs + deferred autoscroll\n\n- Add channel-aware UI IDs (mesh|id, geo:<gh>|id, dm:<peer>|id) to prevent SwiftUI reuse gaps\n- Defer scrollTo to next runloop for stability; auto-scroll on appear/switch\n- Collapse very long messages with Show more/less; skip heavy parsing for huge content\n- Simplify LazyLinkPreviewView (remove GeometryReader in list)


pascal-maker:
question do you still write code or are you just using goose and prior knowledge

PR #487, opened 2025-08-22
Added mentioned username interaction

garydevenay:
Added the ability to tap a username that is mentioned in a message and bring up the user actions.


jtvargas:
Can this be copied to the text field to mention it faster? 👀

garydevenay:
Maybe we should look at adding "reply" to the action sheet, which populates it to the text box.

Edit: Looks like there is already a "mention" option been added, pending release.

jackjackbits:
in new release!

PR #489, opened 2025-08-23
append an extra space so the user can immediately continue typing

jtvargas:
Hey, just a small improvement that I found annoying and I think some users might too. 

Basically, if I try to tag/mention/suggest a username from the chat, the cursor stays in the position next to the username, and the user has to press "space" in order to start typing. This PR just adds an extra space so the user can start typing after mentioning a username.

hope @jackjackbits approve my small PR :) , thanks


jackjackbits:
should be fixed

jtvargas:
@jackjackbits hmm i don't think is there yet, when i try to mention someone "@" i still need to also press the "space" keyword in order to append the space.

here is a screen record on how is working on the latest 1.3.1 version: 

https://github.com/user-attachments/assets/780085c4-0b90-4439-b4b1-2ed1f2165a7a

PR #493, opened 2025-08-23
Added Send button to keyboard action for message sending

garydevenay:
Added the send button to the keyboard for easier message sending.

image


jackjackbits:
thanks!

PR #507, opened 2025-08-24
Fix macOS build - disable views calling visibleGeohashPeople() on mac

makevoid:
I have a small contribution here - a fix to disable iOS-specific features for macOS build in case building on mac is still required.

- Wrap geohash-related code in CommandProcessor with #if os(iOS) directive
- Conditionally compile visibleGeohashPeople() calls in ChatViewModel

This prevents build failures when compiling for macOS target

These changes ensure iOS-specific functionality (geohashing methods) are only compiled when building for iOS, allowing the macOS build to complete successfully.


jackjackbits:
fixed!

PR #532, opened 2025-08-27
Update Contents.json

RodsF1:
Added color purple (I think) 

Also, would be sick if we added the bitchat.land world view layout to the bottom of the #location channels page (below remove location access). 

- @rodbmy


jackjackbits:
unclear what this is for?

RodsF1:
Added a new accent color "purple": "1.000" to the AccentColor.colorset/Contents.json so that the UI can support purple as an accent option. Left "red" unchanged at "0.000". This is required for testing theme variations.

This PR is to introduce a purple accent color option. I mistakenly edited AccentColor.colorset, but based on feedback I’ll create a proper Purple.colorset in Assets instead so Xcode recognizes it.

PR #535, opened 2025-08-27
Fix crash on URLs with short and invalid hostnames

### LINKED ISSUE #534: App crashes displaying URLs with fragments
App crashes when displaying messages containing certain URL patterns. Crashes immediately when message is rendered, you don't need to click the link for the crash.

**Env**
- BitChat Version 1.3.2 
- iPhone 16 Pro Max on iOS 18.5

**Reproduction**
1. Open either the mesh or a geohash channel
2. Send `http://b/#c`
3. App crashes

**Other bad URLs**
- `https://a/#a`
- `http://bitchat/#rules`

App will also crash if any user sends one of these bad URLs and the message is in chat

ahmedkadd:
Fixes crash when displaying messages with URLs containing short and invalid hostnames by validating hostname length and scheme before creating links

Closes #534


ahmedkadd:
I misunderstood the bug. Please disregard this PR. Working on a proper fix

PR #551, opened 2025-08-30
NIP-13 PoW for Geohash + Body-only Mining Animation; Enforce PoW Filtering

jackjackbits:
## Summary
Adds NIP-13 proof-of-work to geohash public chats and a UX that shows your message immediately with a body-only "matrix" animation while PoW is mined; timestamp appears when solved. Incoming geohash events and the geohash people list are filtered by the same local PoW requirement.

## Changes
- PoW core: mines a ["nonce", "<nonce>", "<bits>"] tag so the event id has ≥ target leading zero bits.
- Defaults (no settings): difficulty by geohash precision
  - precision ≤5 → 13 bits
  - precision = 6 → 12 bits
  - precision ≥7 → 11 bits
- Send path (geohash only):
  - Show message immediately (no timestamp), scramble body while mining
  - Sign and publish event with nonce tag when solved; reveal timestamp
- Receive path (geohash only): drop events whose id LZ bits < local policy
- Peer list: only show participants who have posted with adequate PoW in the last 5 minutes (self always visible)
- Miner stability: autoreleasepool around hashing + sched_yield every ~16k iters to avoid memory spikes
- Tests: basic LZ counting + low-difficulty mining

## Files of Interest
-  (PoW core)
-  (difficulty policy)
-  (send/receive wiring, filters, people tracking)
-  +  (body-only animation, combined Text for proper wrapping)
- 

## UX
- While mining: message body scrambles (monospace), sender prefix stays static, timestamp hidden
- On success: body becomes original content, timestamp appended
- Long messages wrap under the name during animation and after (identical layout)

## Rationale
- Adds a cost to public posting that scales with room reach (coarser geohash → higher bits by default)
- Keeps a smooth UX: typical mining times short on modern phones; user sees immediate echo
- Receiver filtering and peer list filtering close the loop against low-cost spam

## Validation
- Unit tests pass locally
- Manual: join a coarse geohash (#sn) and send a long message: body scrambles then resolves with timestamp; only messages meeting threshold appear
- Verify peer list only includes participants who posted with adequate PoW recently

## Follow-ups (optional)
- Add time-budgeted mining with automatic bit step-down under thermal/low-power
- Receiver "Show low PoW" affordance (collapsed by default)
- Apply PoW to other geohash sends (e.g., screenshot/system) if desired

## Rollback
- Revert this PR; no migrations or persistent formats changed.

### DIFF
--- a/bitchat/Services/PowPolicy.swift
+++ b/bitchat/Services/PowPolicy.swift
@@ -9,9 +9,9 @@ enum PowPolicy {
         let precision = geohash.count
         // Start at 13 and go down with higher precision (smaller areas)
         switch precision {
-        case ...5: return 13
-        case 6:    return 12
-        default:   return 11
+        case ...5: return 16
+        case 6:    return 15
+        default:   return 14
         }
     }
 }


ynniv:
nip13 is unlikely to reduce spam because of the efficiency asymmetry between honest clients and spammers (100x+). see https://github.com/ynniv/nip13_miner

jackjackbits:
agree. it’s not the right solution.

kowirth:
And tell me again when you are going to watch a bi polar medication fantasy and conspiracy turn into a healthy life with a woman?

PR #560, opened 2025-09-05
feat: Add Turkish localization support

berkinozturk:
## 🇹🇷 Turkish Localization Support

This PR adds comprehensive Turkish language support to bitchat, enabling Turkish users to use the app in their native language.

### 📋 Changes Made:
- ✅ Created localization infrastructure (`Base.lproj` + `tr.lproj`)
- ✅ Added `LocalizationHelper.swift` with convenient L() function
- ✅ Translated ~150 UI strings to Turkish
- ✅ Updated `project.yml` to support Turkish region
- ✅ Localized core views: `AppInfoView`, `ContentView`, `LocationChannelsSheet`
- ✅ Support for both iOS and macOS platforms

### 🔧 Technical Details:
- Uses standard iOS/macOS localization system
- Extensible architecture for future languages
- No breaking changes to existing functionality
- Follows iOS Human Interface Guidelines for localization

### 🧪 Testing:
- Change device language to Turkish
- All UI text should display in Turkish
- App functionality remains unchanged

### 📱 Translation Examples:
- "offline communication" → "çevrimdışı iletişim"
- "end-to-end encryption" → "uçtan uca şifreleme"
- "type a message..." → "bir mesaj yazın..."

This contribution makes bitchat accessible to Turkish-speaking users worldwide! 🎉

### 🌍 Why Turkish Localization Matters
In Turkey, especially during natural disasters such as earthquakes, internet connectivity can be severely disrupted. Having bitchat available in Turkish ensures that people can still communicate effectively in their native language during offline or peer-to-peer scenarios. This localization is not only about accessibility but also about providing a reliable communication option for Turkish-speaking communities when they need it the most.

### DIFF
--- a/bitchat/Localizations/Base.lproj/Localizable.strings
+++ b/bitchat/Localizations/Base.lproj/Localizable.strings
@@ -1,0 +1,131 @@
+/* 
+  Localizable.strings
+  bitchat
+
+  Created for internationalization support.
+*/
+
+// App Info
+"app.name" = "bitchat";
+"app.tagline" = "sidegroupchat";
+
+// Features Section
+"features.title" = "FEATURES";
+"features.offline_communication" = "offline communication";
+"features.offline_communication.description" = "works without internet using Bluetooth low energy";
+"features.encryption" = "end-to-end encryption";
+"features.encryption.description" = "private messages encrypted with noise protocol";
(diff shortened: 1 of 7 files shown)


jackjackbits:
will focus on #570

PR #564, opened 2025-09-07
Tor-by-default + georelay fan-out=3 + iOS energy/flood guards

jackjackbits:
Summary
- Route all networking via embedded Tor SOCKS (127.0.0.1:39050, fail‑closed). 
- Chat shows “Starting Tor…” then “Tor ready…” once SOCKS is reachable.
- Georelay fan‑out reduced to 3; spare-based rotation swaps failing relays.
- iOS energy guards:
  - Selector-based timers (no @Sendable captures).
  - Global flood limiter for public/geochat processing.
  - High-traffic rendering: plain text + longer batching interval.
- Nostr: prevent duplicate subs per relay; queued-sub handling cleaned up.

Validation
- iOS/macOS: Tor must be ready before network; status messages appear.
- Active geohash: 3 live sockets; rotation replaces failing sockets from spares.
- Under spam: lower CPU due to flood gate + simplified rendering.

Follow-ups (energy)
- Cap total unique Nostr sockets globally and reuse across DM + geo + sampling.
- Suspend sampling when sheet closed; drop geohash sockets on background.
- Defer all subs until Tor SOCKS is reachable to avoid pre-bootstrap churn.
- More aggressive “quiet mode” (skip mention/URL regex) under load.

Flags
- BITCHAT_DEV_ALLOW_CLEARNET (dev only)
- BITCHAT_TOR_BOOTSTRAP_MONITOR / BITCHAT_TOR_STREAM_LOGS (diagnostics)


asmogo:
Very nice! 
Tor now connects almost instantly. In my first test, the app still fails to reconnect after being in the background for an extended period; I’ll do further debugging. 

I haven't been able to run it in the simulator yet (using iOS 18.6.2 on an iPhone 13 Pro).

jackjackbits:
yes, i can’t seem to get it to reconnect after being in background. i didn’t make the simulator slice for tor framework as we can’t use bluetooth in simulator.

jackjackbits:
i think i got this working @asmogo, want to check my work?

asmogo:
Great. 
Yesterday, I had some crashes during my test. 
I'm going to install the latest commit and provide feedback asap.

bitchat/BitchatApp.swift
```diff
@@ -15,6 +15,8 @@ struct BitchatApp: App {
     #if os(iOS)
     @Environment(\.scenePhase) var scenePhase
     @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
+    // Skip the very first .active-triggered Tor restart on cold launch
+    @State private var didHandleInitialActive: Bool = false
```
asmogo, line 19:
```suggestion
    @State private var didHandleInitialActive: Bool = false
    @State private var didEnterBackground: Bool = false
```

bitchat/BitchatApp.swift
```diff
@@ -54,10 +58,37 @@ struct BitchatApp: App {
...
                         // Keep BLE mesh running in background; BLEService adapts scanning automatically
+                        // Optionally nudge Tor to dormant to save power
+                        TorManager.shared.setAppForeground(false)
+                        TorManager.shared.goDormantOnBackground()
+                        // Stop geohash sampling while backgrounded
+                        Task { @MainActor in
+                            chatViewModel.endGeohashSampling()
+                        }
+                        // Proactively disconnect Nostr to avoid spurious socket errors while Tor is down
+                        NostrRelayManager.shared.disconnect()
```
asmogo, line 70:
```suggestion
                        NostrRelayManager.shared.disconnect()
                        didEnterBackground = true
```

bitchat/BitchatApp.swift
```diff
@@ -54,10 +58,37 @@ struct BitchatApp: App {
...
                         break
                     case .active:
                         // Restart services when becoming active
                         chatViewModel.meshService.startServices()
+                        TorManager.shared.setAppForeground(true)
+                        // On initial cold launch, Tor was just started in onAppear.
+                        // Skip the deterministic restart the first time we become active.
+                        if didHandleInitialActive {
+                            // Ensure Tor is healthy; restart deterministically, then wait until ready
+                            TorManager.shared.ensureRunningOnForeground()
```
asmogo:
```suggestion
                        if didHandleInitialActive && didEnterBackground {
                            // Ensure Tor is healthy; restart deterministically, then wait until ready
                            TorManager.shared.ensureRunningOnForeground()
                        } else {
                            didHandleInitialActive = true
                        }
                        didEnterBackground = false
```

jackjackbits: this is outdated?

qalandarov:
> yes, i can’t seem to get it to reconnect after being in background. i didn’t make the simulator slice for tor framework as we can’t use bluetooth in simulator.

@jackjackbits, at least it could help us build and run tests without having to setup dev account each time the project is regenerated :)

jackjackbits:
will add a simulator slice later

PR #571, opened 2025-09-11
Perf: Add `final` to classes that are not inherited

qalandarov:
This allows the Swift compiler to use static dispatch instead of dynamic dispatch, improving performance by avoiding runtime method lookups. No changes to class behavior.


qalandarov:
Please check #574 as I accidentally closed this one when managing branches and remotes

PR #576, opened 2025-09-11
Regenerate info.plist by adding the missing keys

qalandarov:
`xcodegen` probably uses alphabetical sorting so had to commit the info.plist to avoid discrepancies


qalandarov:
Speaking of it, is `xcodegen` still beneficial given that Xcode is no longer generating hashes for new files and works with the file/folder structure directly? cc: @jackjackbits 

We could use a simple project and use local SPM packages if needed (which could also help with modularization)

jackjackbits:
i don't use it anymore. open to alternatives

qalandarov:
> i don't use it anymore. open to alternatives

Here's my proposal to migrate to the new Xcode-16 format: #594

PR #584, opened 2025-09-11
Refactor: Testable Keychain and Identity Manager

qalandarov:
Currently tests are using the real keychain which could be problematic to run on a CI / Github Action.

This PR is breaking the singleton pattern for the Keychain and Identity Manager so they can be injected and mocked.

Also, Identity Manager seem to have lots of unused functions which were removed (not sure if there were plans to use them for the future)


jackjackbits:
can review my fixes and i'll merge

qalandarov:
LGTM!

Btw, a disclaimer: when I was running the tests, they were still picking up the real keychain probably because of the real app object being created... even though this PR is making the keychain (and identity manager) testable, I couldn't force them to be used (unless I missed something).

PR #592, opened 2025-09-12
Fix string bounds crashes in ChatViewModel

praneethreddy31:
Found a bunch of places where we were doing .prefix(8) on strings that might be shorter than 8 characters. This could crash the app if we got weird pubkeys or message IDs.

ChatViewModel.swift

What I changed:

Replaced all the risky pubkey.prefix(8) calls with pubkey.prefix(min(8, pubkey.count))

Also fixed a force unwrap that could crash: privateChats[stableKeyHex]! → privateChats[stableKeyHex] ?? []

Why this matters:
Sometimes we get malformed pubkeys or short message IDs from the network. Instead of crashing the whole app, we'll just show whatever characters we have. The logging still works fine and users don't see crashes.

Examples of what got fixed:

swift
// Before - could crash
senderPubkey.prefix(8)
messageID.prefix(8)

// After - safe
senderPubkey.prefix(min(8, senderPubkey.count))
messageID.prefix(min(8, messageID.count))
Tested with empty strings and short strings - no more crashes. Normal length strings still work exactly the same.


qalandarov:
it wouldn't crash, it would just return the entire string: [docs](https://developer.apple.com/documentation/swift/array/prefix(_:))

PR #598, opened 2025-09-13
Location Notes (kind 1) at Building Precision + Live Updates

jackjackbits:
## Summary
This draft PR adds location notes (Nostr kind 1) scoped to the current building (geohash precision 8), with a focused UI flow and careful perf/UX behavior:

- Publish/view kind‑1 notes tagged with `#g=<geohash>` and optional `#n=<nickname>` (no teleport tag)
- Mesh toolbar notes entry (icon blue when notes exist, grey otherwise)
- A dedicated notes sheet that follows your current building as you walk (live updates with 10 m distance filter)
- Haptic on building change while the sheet is open (iOS)
- Count moved into sheet header (“1 note @ #ghash” / “N notes @ #ghash”)
- Reverse‑geocoded building name (fallback: block) displayed under the title
- EOSE handling for subscriptions (initial load completion) with safe 2s fallback
- Uses GeoRelays for send/subscribe (closest relays by geohash)

## UI/UX Details
- Toolbar
  - Notes icon: `long.text.page.and.pencil`
  - Color: grey if zero notes; blue if one or more
  - Hidden when location isn’t authorized
- Notes Sheet
  - Title: “1 note @ #<8‑geohash>” or “N notes @ #<8‑geohash>” (count is bold)
  - Subtitle: building name if available, else block name
  - Live location updates only while sheet is open; 10 m distance filter
  - Haptic feedback on building geohash change (iOS only)
  - Message list reverse‑chron; monospaced text; no delete; timestamps are relative (<7d) or absolute

## Networking
- Notes and counts subscribe at precision 8 (building) via `GeoRelayDirectory.closestRelays(...)`
- EOSE tracked per subscription; managers expose `initialLoadComplete`
- No periodic polling; live updates only while the sheet is visible

## Implementation Highlights
- `NostrProtocol.createGeohashTextNote(...)` (kind 1) — tags: `g`, optional `n` (no `t/teleport`)
- `NostrRelayManager.subscribe(..., onEOSE:)` — EOSE callback + 2s fallback, Swift 6 isolation-safe
- `LocationNotesManager` — subscribes/sends notes; live geohash switch in-place
- `LocationNotesCounter` — subscribes to current building geohash for toolbar state
- `LocationChannelManager` — reverse-geocoding + live refresh (10 m) API used only while sheet is open

## Notable Defaults & Safety
- Tor enforced path remains; EOSE fallback prevents “stuck” initial loads
- Building precision (8) is not exposed in the channel selector; user-visible channels remain block/city/etc.
- Toolbar does a one‑shot refresh on appear when needed (authorized + empty channels) to initialize state without keeping live updates on

## Testing Notes
- Walk across building boundaries with sheet open; header + list should follow and light haptic should fire once per change
- With sheet closed, toolbar icon reflects presence (grey/blue) after a one‑shot refresh
- Poor network: initial load should complete via fallback timer; live events continue afterward

## Future Work (Nice-to-have)
- Vote (up/down) support via replaceable events (kind 30001) with geohash tags
- Debounce haptic at borders if needed
- Optional sorting by score (top) in addition to reverse‑chron

### DIFF
--- a/bitchat/Views/LocationNotesSheet.swift
+++ b/bitchat/Views/LocationNotesSheet.swift
@@ -1,0 +1,66 @@
+import SwiftUI
+
+struct LocationNotesSheet: View {
+    @EnvironmentObject var viewModel: ChatViewModel
+    @ObservedObject private var locationManager = LocationChannelManager.shared
+    @Binding var notesGeohash: String?
+    @Environment(\.dismiss) private var dismiss
+    @Environment(\.colorScheme) private var colorScheme
+
+    private var backgroundColor: Color { colorScheme == .dark ? .black : .white }
+    private var textColor: Color { colorScheme == .dark ? .green : Color(red: 0, green: 0.5, blue: 0) }
+    private var secondaryTextColor: Color { textColor.opacity(0.8) }
+
+    var body: some View {
+        Group {
+            if let gh = notesGeohash ?? locationManager.availableChannels.first(where: { $0.level == .block })?.geohash {
+                // Found block geohash: show notes view
+                LocationNotesView(geohash: gh)
+                    .environmentObject(viewModel)
+            } else {
+                // Acquire location: keep a loading overlay (Matrix) until we either get a block geohash
+                ZStack {
+                    VStack(spacing: 0) {
+                        HStack {
+                            VStack(alignment: .leading, spacing: 2) {
+                                Text("notes")
+                                    .font(.system(size: 16, weight: .bold, design: .monospaced))
+                                Text("acquiring location…")
+                                    .font(.system(size: 12, design: .monospaced))
+                                    .foregroundColor(secondaryTextColor)
+                            }
+                            Spacer()
+                            Button(action: { dismiss() }) {
+                                Image(systemName: "xmark")
+                                    .font(.system(size: 13, weight: .semibold, design: .monospaced))
+                                    .foregroundColor(textColor)
+                                    .frame(width: 32, height: 32)
+                            }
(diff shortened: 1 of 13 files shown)


bitchat/Views/LocationNotesSheet.swift
```diff
@@ -0,0 +1,65 @@
...
+                            }
+                            .buttonStyle(.plain)
+                            .accessibilityLabel("Close")
+                        }
+                        .frame(height: 44)
+                        .padding(.horizontal, 12)
+                        .background(backgroundColor.opacity(0.95))
+                        Spacer()
+                    }
+                    MatrixRainView()
```
qalandarov, line 47: Looks like it's not in the project

bitchat/Views/LocationNotesSheet.swift
```diff
@@ -0,0 +1,65 @@
...
+    @Binding var notesGeohash: String?
+    @Environment(\.dismiss) private var dismiss
+    @Environment(\.colorScheme) private var colorScheme
+
+    private var backgroundColor: Color { colorScheme == .dark ? .black : .white }
+    private var textColor: Color { colorScheme == .dark ? .green : Color(red: 0, green: 0.5, blue: 0) }
+    private var secondaryTextColor: Color { textColor.opacity(0.8) }
+
+    var body: some View {
+        Group {
```
qalandarov, line 15: If you not using any modifiers to the Group, you can use the if/else directly without wrapping into a Group. (also, might be worth extracting the "else" view into a private view to improve readability)

PR #600, opened 2025-09-13
Modularization: Extract SecureLogger into a separate module

qalandarov:
Extracting logical modules into separate packages would help us with better code management and benefit from additional `public` and `open` access levels and better-scoped testing.

This would also allow us create `internal` files without having to create a `private` file inside another file (e.g. `ParsedInbound`) just to "hide" its implementation.

Logger is of a foundational-level and many features depend on it. In order to extract Tor into a separate module we have to extract its dependency, Logger, first.


qalandarov:
The `main` is broken

jackjackbits:
> The `main` is broken

?

qalandarov:
> > The `main` is broken
> 
> ?

`MatrixRainView()` is missing from #598 (added comments there too)

The latest commit of the `main` is also [failing](https://github.com/permissionlesstech/bitchat/actions/runs/17700877456/job/50306390825#step:4:420)

jackjackbits:
everything compiling for me and i've submitted to TestFlight many times

qalandarov:
Most likely because you added the file as a reference and didn’t copy into
the project. That’s why it only works on your machine because the file
exists there.

jackjackbits:
hm. i don't have the file anywhere in my codebase. i searched. i'm on main and pulling directly from it.

qalandarov:
> hm. i don't have the file anywhere in my codebase. i searched. i'm on main and pulling directly from it.

Looks like [`Views/LocationNotesSheet.swift`](https://github.com/permissionlesstech/bitchat/blob/48db5502d7c412f717d9f37796215ba615634f3c/bitchat/Views/LocationNotesSheet.swift) was added in #598, but not used and when it's added into Xcode or Swift is built from the command line - it's failing.

We can either remove the file if it's not needed or add the missing matrix rain that it's referencing

qalandarov:
Removing `LocationNotesSheet.swift` if it's not needed: #606

qalandarov:
I have seen this [test failure](https://github.com/permissionlesstech/bitchat/actions/runs/17712297461/job/50332603435?pr=600#step:5:155) intermittently locally as well

PR #606, opened 2025-09-13
Remove unused LocationNotesSheet.swift

qalandarov:
It's not used in the project, yet it's existence is causing build failure because it's referencing `MatrixRainView` that doesn't exist in the codebase


SkinnnyJay:
I wonder if this is a WIP that slipped in via the "Location Notes (kind 1) at Building Precision + Live updates" (2870acdcc732c04b1f143a3606300c3124e61baf), and this isn't fully implemented yet, or it was a dangling resource that was left after the update.

jackjackbits:
got it. thanks!

PR #608, opened 2025-09-14
Xcode Configuration files: `.xcconfigs` + Remove `xcodegen`

qalandarov:
This would help us simplify the config changes and make local development easier without having to update the development team and bundle ID each time by adding them to a git-ignored file `Configs/Local.xcconfig` without affecting or changing the repo owner's workflow.

---

Also, removed references to `xcodegen` and its `project.yml`


jackjackbits:
thanks!

Issue #607, opened 2025-09-14
Share Extension is broken


qalandarov:
When trying to share anything with bitchat from the share extension a view comes in and quickly disappears

https://github.com/user-attachments/assets/9e265d2b-ce19-4f09-9fa4-c409dad5f5f8

SeamusMullan:
if you slow the video down it does say 'shared link to bitchat'. maybe this just needs a little timer or an ok button for some better UX lol

qalandarov:
> if you slow the video down it does say 'shared link to bitchat'. maybe this just needs a little timer or an ok button for some better UX lol

but it actually doesn't do anything... the bitchat app doesn't have any new message from me

r1b2ns:
@qalandarov  I just submitted this PR fixing this crash. The issue was caused by processing the content on the main thread, and the success screen was displayed for too short a time.

[621](https://github.com/permissionlesstech/bitchat/pull/621)

jackjackbits:
Closing as resolved/obsolete against the current `main` and v1.7.1-era architecture. The implementation has changed materially since this report and the reported gap is now covered. If this still reproduces on a current release, please open a focused report with the app version, OS/device, exact steps, and relevant logs.

PR #621, opened 2025-09-15
fix: crash on shared extension

r1b2ns:
Prevent crash in Share Extension by processing in background and increasing execution time

[issue 607](https://github.com/permissionlesstech/bitchat/issues/607)

EVD

https://github.com/user-attachments/assets/fdaac389-a751-4791-864a-855db454c291

### DIFF
--- a/bitchatShareExtension/ShareViewController.swift
+++ b/bitchatShareExtension/ShareViewController.swift
@@ -163,8 +163,8 @@ final class ShareViewController: UIViewController {
     private func finishWithMessage(_ msg: String) {
         statusLabel.text = msg
         // Complete shortly after showing status
-        DispatchQueue.main.asyncAfter(deadline: .now() + TransportConfig.uiShareExtensionDismissDelaySeconds) { [weak self] in
-            self?.extensionContext?.completeRequest(returningItems: [], completionHandler: nil)
+        DispatchQueue.main.asyncAfter(deadline: .now() + TransportConfig.uiShareExtensionDismissDelaySeconds) {
+            self.extensionContext?.completeRequest(returningItems: [], completionHandler: nil)
         }
     }
 }


bitchatShareExtension/ShareViewController.swift
```diff
@@ -35,8 +35,9 @@ final class ShareViewController: UIViewController {
             statusLabel.leadingAnchor.constraint(greaterThanOrEqualTo: view.layoutMarginsGuide.leadingAnchor),
             statusLabel.trailingAnchor.constraint(lessThanOrEqualTo: view.layoutMarginsGuide.trailingAnchor)
         ])
-
-        processShare()
+        DispatchQueue.global(qos: .default).async {
```
qalandarov: isn't `qos: .default` is the default option? maybe we could remove it? and just have `.global.async`?

r1b2ns: done

bitchatShareExtension/ShareViewController.swift
```diff
@@ -35,8 +35,8 @@ final class ShareViewController: UIViewController {
             statusLabel.leadingAnchor.constraint(greaterThanOrEqualTo: view.layoutMarginsGuide.leadingAnchor),
             statusLabel.trailingAnchor.constraint(lessThanOrEqualTo: view.layoutMarginsGuide.trailingAnchor)
         ])
-        DispatchQueue.global(qos: .default).async {
-            self.processShare()
+        DispatchQueue.global().async { [weak self] in
```
qalandarov: Dispatch queues don't cause retain cycle, so this could cause confusion

r1b2ns: Nice man, done!!!

bitchatShareExtension/ShareViewController.swift
```diff
@@ -162,8 +163,8 @@ final class ShareViewController: UIViewController {
     private func finishWithMessage(_ msg: String) {
         statusLabel.text = msg
         // Complete shortly after showing status
-        DispatchQueue.main.asyncAfter(deadline: .now() + TransportConfig.uiShareExtensionDismissDelaySeconds) {
-            self.extensionContext?.completeRequest(returningItems: nil, completionHandler: nil)
+        DispatchQueue.main.asyncAfter(deadline: .now() + TransportConfig.uiShareExtensionDismissDelaySeconds) { [weak self] in
```
qalandarov: I think same `[weak self]` here :)

r1b2ns: Right, I forgot this one

Done

qalandarov:
lgtm

PR #622, opened 2025-09-15
Disable “precise location” by default for privacy

qalandarov:
Not sure if this would affect finding nearby people, but at least the users will have an option to turn on the precise location if they choose to do so instead of it being on by default as the app doesn't really need it to operate.

### DIFF
--- a/bitchat/Info.plist
+++ b/bitchat/Info.plist
@@ -37,6 +37,8 @@
 	<string>bitchat uses Bluetooth to discover and connect with other bitchat users nearby.</string>
 	<key>NSCameraUsageDescription</key>
 	<string>bitchat uses the camera to scan QR codes to verify peers.</string>
+	<key>NSLocationDefaultAccuracyReduced</key>
+	<true/>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>bitchat uses your approximate location to compute local geohash channels for optional public chats. Exact GPS is never shared.</string>
 	<key>UIBackgroundModes</key>


jackjackbits:
pretty sure the new persistent notes feature would break if we didn't have precise. we have lots to preserve privacy including random npubs and tor to mask IP address. otherwise location is never sent anywhere.

qalandarov:
Does that mean if the users choose to disable the precise location the persistent notes would break?

jackjackbits:
i think so. it just wouldn’t be as accurate. will have to test.

qalandarov:
Okay. Given that users do have the option to turn this off and the bitchat's promise of privacy, I just thought it made sense to turn it off by default.

Maybe we can inform the user about this feature in the "usage description" test otherwise it sounds like all we need is just a proximate location.

jackjackbits:
let me test to see how good it actually is. the geo notes feature is about a 40x40m cell.

qalandarov:
Looks like reduced accuracy is within 1-20km of the actual location as per [the docs](https://developer.apple.com/documentation/corelocation/kcllocationaccuracyreduced)

qalandarov:
Maybe we can detect if the precise location is off and disable the `block` and `neighborhood` entries with the useful info and have an "enable precise location" button to activate those?

qalandarov:
@jackjackbits should we close it or are we pending some testing?

jackjackbits:
still testing. On 26 Sep 2025, at 01:13, Islam ***@***.***> wrote:﻿qalandarov left a comment (permissionlesstech/bitchat#622)
@jackjackbits should we close it or are we pending some testing?

PR #633, opened 2025-09-16
Add `.notSentYet` to avoid optional DeliveryStatus

qalandarov:


jackjackbits:
can you describe intent here a bit more? want to be extra cautious around network/protocol changes.

qalandarov:
It was just to simplify the code to avoid having to deal with optional enums

qalandarov:
Looks like we could use `case .none` to cover the `nil`, so probably not much value in this PR

PR #634, opened 2025-09-16
chore: debug icon

r1b2ns:
Add a debug icon in Bitchat to make it easier to identify debug builds while keeping the App Store version unchanged

IMG_B27FE0969633-1


bitchat/Assets.xcassets/AppIconDebug.appiconset/Contents.json
```diff
@@ -0,0 +1,116 @@
...
+      "idiom" : "ipad",
+      "scale" : "2x",
+      "size" : "76x76"
+    },
+    {
+      "filename" : "image-83.5@2x.png",
+      "idiom" : "ipad",
+      "scale" : "2x",
+      "size" : "83.5x83.5"
+    },
```
qalandarov: With the latest Xcode all you need is a single 1024x1024

bitchat/Assets.xcassets/AppIconDebug.appiconset/image-1024.png
qalandarov, line 1: Would this image be bundled with the production app too? is there any way we can prevent that from happening?

r1b2ns: Nice man, good question. We are already preventing this by setting up the AppIconDebug in the build settings

bitchat/Assets.xcassets/AppIconDebug.appiconset/Contents.json
r1b2ns: Awesome, you're right. I'll fix that

r1b2ns: Done

bitchat.xcodeproj/project.pbxproj
mmatoszko, line 1: doesn't it also change the app icon for the production app? 😄 how does it work?

r1b2ns: Hi man, no. This configuration is just for the app when using Debug. If you run the app with the Debug configuration, it will set up the AppIconDebug file.

You can see this configuration in
Project Settings > Build Settings > Primary App Icon Sets.

For release configuration nothing change

mmatoszko: thank you 👍

PR #635, opened 2025-09-16
Refactor 2/n: ChatViewModel's Geohash Subscription

qalandarov:
Continuing breaking down ChatViewModel from #613.

This PR breaks down / "flattens" 5 big functions into smaller ones, each as a separate commit, which could eventually be extracted into dedicated services/helpers.

### DIFF
--- a/bitchat/ViewModels/ChatViewModel.swift
+++ b/bitchat/ViewModels/ChatViewModel.swift
@@ -916,111 +916,18 @@ final class ChatViewModel: ObservableObject, BitchatDelegate {
             self?.subscribeNostrEvent(event)
         }
         // Resubscribe geohash DMs for this identity
-        if let dmSub = geoDmSubscriptionID { NostrRelayManager.shared.unsubscribe(id: dmSub); geoDmSubscriptionID = nil }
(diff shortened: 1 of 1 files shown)


bitchat/ViewModels/ChatViewModel.swift
```diff
@@ -913,179 +913,212 @@ final class ChatViewModel: ObservableObject, BitchatDelegate {
...
+            handleDelivered(noisePayload, senderPubkey: senderPubkey, convKey: convKey)
+        case .readReceipt:
+            handleReadReceipt(noisePayload, senderPubkey: senderPubkey, convKey: convKey)
+        case .verifyChallenge, .verifyResponse:
+            // QR verification payloads over Nostr are not supported; ignore in geohash DMs
+            break
+        }
+    }
+    
+    private func handlePrivateMessage(
```
qalandarov, line 1041: This and another `handlePrivateMessage` in line 1791 have **almost** identical implementation.
@jackjackbits, do you think we can combine these into one?

qalandarov:
The tests are passing fine locally

Issue #618, opened 2025-09-15
Move UserDefaults behind a protocol to be testable


qalandarov:
Currently `UserDefaults` used throughout the code directly and in some cases we're even [testing its implementation](https://github.com/permissionlesstech/bitchat/blob/main/bitchatTests/GeohashBookmarksStoreTests.swift#L43), rather we should test against a mock to validate our internal logic and not that of `UserDefaults`'

r1b2ns:
Hey @qalandarov, are you working on it? If not I can get this. Please let me know, I’d like to help.

qalandarov:
@r1b2ns please go ahead, and thank you!

r1b2ns:
Hi @qalandarov, #636 tries to solve this issue.

jackjackbits:
Closing as resolved by a different route than the protocol wrapper: stores now take injected `UserDefaults` instances (e.g. `GeohashBookmarksStore(storage:)`), and the test this issue linked — along with the rest of the suite — runs against throwaway `UserDefaults(suiteName: UUID().uuidString)` instances with `removePersistentDomain` cleanup, so nothing tests against `.standard`. The protocol approach (#636) was considered and closed; suite injection achieves the isolation with less indirection.

PR #636, opened 2025-09-17
feat: add StorageProtocol protocol to abstract UserDefaults

r1b2ns:
In order to address #618, this PR introduces KeyStorable, a protocol that abstracts implementations of key–value persistence systems. It will help avoid using UserDefaults directly in the app.

### DIFF
--- a/bitchat/Services/GeohashBookmarksStore.swift
+++ b/bitchat/Services/GeohashBookmarksStore.swift
@@ -20,13 +20,11 @@ final class GeohashBookmarksStore: ObservableObject {
     private let geocoder = CLGeocoder()
     private var resolving: Set<String> = []
     #endif
-    
-    private let storage: UserDefaults
 
-    private let store: StorageProtocol
+    private let storage: StorageProtocol
     
-    init(store: StorageProtocol = UserDefaultsStorage()) {
-        self.store = store
+    init(storage: StorageProtocol = UserDefaultsStorage()) {
+        self.storage = storage
         load()
     }
 
@@ -69,7 +67,7 @@ final class GeohashBookmarksStore: ObservableObject {
 
     // MARK: - Persistence
     private func load() {
-        guard let data: Data = store.get(storeKey) else { return }
+        guard let data: Data = storage.get(storeKey) else { return }
         if let arr = try? JSONDecoder().decode([String].self, from: data) {
             // Sanitize, normalize, dedupe while preserving order (first occurrence wins)
             var seen = Set<String>()
@@ -86,7 +84,7 @@ final class GeohashBookmarksStore: ObservableObject {
             membership = seen
         }
         // Load any saved names
-        if let namesData: Data = store.get(namesStoreKey),
+        if let namesData: Data = storage.get(namesStoreKey),
            let dict = try? JSONDecoder().decode([String: String].self, from: namesData) {
             bookmarkNames = dict
         }
@@ -94,13 +92,13 @@ final class GeohashBookmarksStore: ObservableObject {
 
     private func persist() {
         if let data = try? JSONEncoder().encode(bookmarks) {
-            store.set(data, key: storeKey)
+            storage.set(data, key: storeKey)
         }
     }
 
     private func persistNames() {
         if let data = try? JSONEncoder().encode(bookmarkNames) {
-            store.set(data, key: namesStoreKey)
+            storage.set(data, key: namesStoreKey)
         }
     }
 
--- a/bitchatTests/GeohashBookmarksStoreTests.swift
+++ b/bitchatTests/GeohashBookmarksStoreTests.swift
@@ -3,23 +3,21 @@ import XCTest
 
 final class GeohashBookmarksStoreTests: XCTestCase {
     let storeKey = "locationChannel.bookmarks"
-    var storage: UserDefaults!
-    var store: GeohashBookmarksStore!
 
-    private var store: UserDefaultsStorageMock!
+    private var storage: UserDefaultsStorageMock!
     private var sut: GeohashBookmarksStore!
     
     override func setUp() {
         super.setUp()
-        store = UserDefaultsStorageMock()
+        storage = UserDefaultsStorageMock()
         sut = GeohashBookmarksStore(
-            store: store
+            storage: storage
         )
     }
     
     override func tearDown() {
         super.tearDown()
-        store = nil
+        storage = nil
         sut = nil
     }
 
@@ -37,13 +35,13 @@ final class GeohashBookmarksStoreTests: XCTestCase {
         XCTAssertFalse(sut.isBookmarked("u4pruy"))
         XCTAssertTrue(sut.bookmarks.isEmpty)
     }
-
+    
     func testPersistenceWritten() throws {
         sut.toggle("ezs42")
         sut.toggle("u4pruy")
-
+        
         // Verify persisted JSON contains both (order not enforced here)
-        guard let data: Data = store.get(storeKey) else {
+        guard let data: Data = storage.get(storeKey) else {
             XCTFail("No persisted data found")
             return
         }
--- a/bitchat.xcodeproj/project.pbxproj
+++ b/bitchat.xcodeproj/project.pbxproj
@@ -70,6 +70,7 @@
 			isa = PBXFileSystemSynchronizedBuildFileExceptionSet;
 			membershipExceptions = (
 				Services/TransportConfig.swift,
+				Shared/StorageProtocol.swift,
 			);
 			target = 57CA17A36A2532A6CFF367BB /* bitchatShareExtension */;
 		};
--- a/bitchat/Views/Components/TextMessageView.swift
+++ b/bitchat/Views/Components/TextMessageView.swift
@@ -91,7 +91,8 @@ struct TextMessageView: View {
     .environmentObject(
         ChatViewModel(
             keychain: keychain,
-            identityManager: SecureIdentityStateManager(keychain)
+            identityManager: SecureIdentityStateManager(keychain),
+            storage: UserDefaultsStorage()
         )
     )
 }
--- a/bitchatTests/UserDefaultsStorageTests.swift
+++ b/bitchatTests/UserDefaultsStorageTests.swift
@@ -1,0 +1,48 @@
+//
+//  UserDefaultsStorageTests.swift
+//  bitchat
+//
+//  Created by Rubens Machion on 26/09/25.
+//
+
+import XCTest
+@testable import bitchat
+
+final class UserDefaultsStorageTests: XCTestCase {
+    var sut: UserDefaultsStorage!
+
+    override func setUp() {
+        super.setUp()
+        sut = UserDefaultsStorage()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        sut = nil
+    }
+
+    func testSetAndGetValue() {
+        sut.set("Hello", key: "greeting")
+        let value: String? = sut.get("greeting")
+        XCTAssertEqual(value, "Hello")
+    }
+
+    func testOverwriteValue() {
+        sut.set(1, key: "counter")
+        sut.set(2, key: "counter")
+        let value: Int? = sut.get("counter")
+        XCTAssertEqual(value, 2, "The last value should overwrite the previous one")
+    }
+
+    func testRemoveValue() {
+        sut.set(true, key: "flag")
+        sut.remove("flag")
+        let value: Bool? = sut.get("flag")
+        XCTAssertNil(value, "The removed value should be nil")
+    }
+
+    func testGetNonExistentKey() {
+        let value: String? = sut.get("not_exists")
+        XCTAssertNil(value, "Nonexistent keys should return nil")
+    }
+}


bitchat/Services/GeohashBookmarksStore.swift
```diff
@@ -21,7 +21,10 @@ final class GeohashBookmarksStore: ObservableObject {
     private var resolving: Set<String> = []
     #endif
 
-    private init() {
+    private let storage: KeyStorable
+    
+    private init(storage: KeyStorable = UserDefaultsKeyStorable()) {
```
qalandarov: keeping the init private won't helps us make `GeohashBookmarksStore` mockable

bitchat/Services/LocationChannelManager.swift
```diff
@@ -35,19 +35,21 @@ final class LocationChannelManager: NSObject, CLLocationManagerDelegate, Observa
 
     // Persisted set of geohashes that were selected via teleport
     private var teleportedSet: Set<String> = []
+    
+    private let storage: KeyStorable = UserDefaultsKeyStorable()
```
qalandarov: pass it in the init?

bitchat/Services/NetworkActivationService.swift
```diff
@@ -17,13 +17,15 @@ final class NetworkActivationService: ObservableObject {
     private let torPreferenceKey = "networkActivationService.userTorEnabled"
     private var torAutoStartDesired: Bool = false
 
+    private let storage: KeyStorable = UserDefaultsKeyStorable()
```
qalandarov: same here

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+
+///
+/// # KeyStorable
+///
+/// A protocol that abstracts key–value storage in the project.
+/// It makes it easier to save and retrieve values in memory, `UserDefaults`,
+/// or any other type of key–value persistence.
+///
+protocol KeyStorable {
```
qalandarov: I just find the name a bit ambiguous given there's "Keychain", calling it Storage or Persistence would cause confusion with a database, other names I can think of are LocalStorage, Preferences, Settings... what you guys think?

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,53 @@
...
+    func save(_ value: Any?, key: String)
```
qalandarov: Again naming :)

`save` sounds like "sync to the disk" - why not stick with a simple `set`?

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,53 @@
...
+    func data(_ key: String) -> Data?
+    func object(_ key: String) -> Any?
+    func string(_ key: String) -> String?
```
qalandarov: maybe we could combine these into a simple `get` with a generic? Something like `func get<T>(_ key: String) -> T?` (even the keys I'd prefer to create a local enum so we don't pass "magic strings" around

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,53 @@
...
+/// An implementation of `KeyStorable` that uses the standard `UserDefaults`
+/// for key–value storage.
+///
+final class UserDefaultsKeyStorable: KeyStorable {
+    
+    private let defaults: UserDefaults
+    
+    init(group: String) {
+        self.defaults = UserDefaults(suiteName: group) ?? .standard
+    }
```
qalandarov: I'd propose removing it and using a single init where you pass a custom UserDefaults object if you want or use the defaults...

bitchat/ViewModels/ChatViewModel.swift
```diff
@@ -463,9 +462,10 @@ final class ChatViewModel: ObservableObject, BitchatDelegate {
             // Only persist if there are changes
             guard oldValue != sentReadReceipts else { return }
             
+
```
qalandarov: 🤔

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,16 +4,18 @@ import XCTest
 final class GeohashBookmarksStoreTests: XCTestCase {
     let storeKey = "locationChannel.bookmarks"
 
+    private let storage: KeyStorable = UserDefaultsKeyStorable()
```
qalandarov: The whole point was not to use the real UserDefaults in tests :)

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: I think LocalStorage could be a good name, but something like LocalStorable to indicate that it’s a protocol (like Codable, Sendable, etc.) might be a better fit. I’d also like to add the word Key to the name to make it clear that this protocol is specifically for persisting key–value data. Maybe LocalKeyStorable?

What do you think?

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: Partially done. About the magic strings, I think the enum could become massive once the app starts getting bigger.

Maybe we can create some generic type of this key and let that the each classe implement you own keys? 

What do you think?

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: done

bitchat/Services/NetworkActivationService.swift
r1b2ns: done

bitchat/Services/LocationChannelManager.swift
r1b2ns: done

bitchat/Services/GeohashBookmarksStore.swift
r1b2ns: done

bitchat/ViewModels/ChatViewModel.swift
r1b2ns: done

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: done

bitchatTests/GeohashBookmarksStoreTests.swift
r1b2ns: done

bitchat/Shared/KeyStorableProtocol.swift
qalandarov: I think UserDefaults should be used sparingly, just for settings/preferences or small amount of data here and there... if things start growing, we can combine multiple fields under a single Codable and store that object as Data or we need to start using a proper DB

bitchat/Shared/KeyStorableProtocol.swift
qalandarov: I don't want to sound pedantic, but I have 2 points about naming :)

1) I'm not sure if protocols always have to be adjectives except maybe when describing a model, as Swift has other naming conventions too like Sequence, Collection, DataProtocol, and even View.
2) I'd either suggest adding both "Key" and "Value" in the name or none because otherwise it sounds like "storage for local keys" vs "local key-value storage". Also, UserDefaults also deals with keys and values, but doesn't have it in the name, so I'm not sure how much value are we getting by it.

bitchat/Services/GeohashBookmarksStore.swift
```diff
@@ -5,7 +5,7 @@ import CoreLocation
 #endif
 
 /// Stores a user-maintained list of bookmarked geohash channels.
-/// - Persistence: UserDefaults (JSON string array)
+/// - Persistence: KeyStorable (JSON string array)
```
qalandarov: Don't forget to update this once the naming is finalized :)

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,39 @@
...
+///
+/// # KeyStorable
+///
+/// A protocol that abstracts key–value storage in the project.
+/// It makes it easier to save and retrieve values in memory, `UserDefaults`,
+/// or any other type of key–value persistence.
+///
+protocol KeyStorable {
+    func set(_ value: Any?, key: String)
+    func get<T>(key: String) -> T?
```
qalandarov: I'd propose "hiding" the `key` so it's a simple `storage.get(<some_key>)`

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,39 @@
...
+    func remove(_ key: String)
+}
+
+///
+/// An implementation of `KeyStorable` that uses the standard `UserDefaults`
+/// for key–value storage.
+///
+final class UserDefaultsKeyStorable: KeyStorable {
```
qalandarov: I'd propose calling the protocol something like `StorageProtocol` and this object as `Storage` and use the `UserDefaults` as implementation detail

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,39 @@
...
+///
+protocol KeyStorable {
+    func set(_ value: Any?, key: String)
+    func get<T>(key: String) -> T?
+    func remove(_ key: String)
+}
+
+///
+/// An implementation of `KeyStorable` that uses the standard `UserDefaults`
+/// for key–value storage.
```
qalandarov: Since we can pass other `UserDefaults` instances (and not just the `standard`) how about something like this:

"An implementation of <protocol name> that uses `UserDefaults` as a backing storage"

bitchat/BitchatApp.swift
```diff
@@ -140,20 +141,21 @@ struct BitchatApp: App {
         guard let userDefaults = UserDefaults(suiteName: BitchatApp.groupID) else {
             return
         }
+        let storage = UserDefaultsKeyStorable(defaults: userDefaults)
```
qalandarov: I'd suggest having a failable initializer in the implementation that takes a suite name and internally fail if the passed `UserDefaults` instance is nil so the calling site never has to deal with `UserDefaults` and doesn't have to know the implementation details.
So this code could be something like:

```
guard let storage = <storage_type_name>(groupID: BitchatApp.groupID) else { return }
```

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,43 +4,45 @@ import XCTest
 final class GeohashBookmarksStoreTests: XCTestCase {
     let storeKey = "locationChannel.bookmarks"
 
+    private let storage = UserDefaultsKeyStorableStub()
```
qalandarov: I thought stubs provide like hard-coded responses where mocks simulate a real object

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,43 +4,45 @@ import XCTest
...
-    }
-
-    override func tearDown() {
-        // Clean after each test
-        UserDefaults.standard.removeObject(forKey: storeKey)
-        GeohashBookmarksStore.shared._resetForTesting()
-        super.tearDown()
+        sut = GeohashBookmarksStore(
+            storage: storage
+        )
```
qalandarov, line 15: Maybe to avoid confusion keep this property named `store` and create a new instance with the storage mock each time? Something like:

```
store = GeohashBookmarksStore(storage: <mock type>())
```

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,43 +4,45 @@ import XCTest
 final class GeohashBookmarksStoreTests: XCTestCase {
     let storeKey = "locationChannel.bookmarks"
 
+    private let storage = UserDefaultsKeyStorableStub()
+    private var sut: GeohashBookmarksStore?
```
qalandarov: why not keep it as implicitly-unwrapped to simplify the tests?

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,43 +4,45 @@ import XCTest
...
         super.setUp()
-        // Clear persisted state before each test
-        UserDefaults.standard.removeObject(forKey: storeKey)
-        GeohashBookmarksStore.shared._resetForTesting()
-    }
-
-    override func tearDown() {
-        // Clean after each test
-        UserDefaults.standard.removeObject(forKey: storeKey)
-        GeohashBookmarksStore.shared._resetForTesting()
```
qalandarov: If you're not using this function, might as well remove it?

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: Awesome man!!! 

I think `LocalStorage` is good enough for it. I’m really terrible at naming classes and structures 😅

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: Nice!!!
So do you think it’s better if we create an enum and centralize these keys in our new protocol?
For me, it’s definitely ok. As you said, UserDefaults needs to be used sparingly and I agree with you.

bitchatTests/GeohashBookmarksStoreTests.swift
r1b2ns: it was deleted

r1b2ns:
Thank you @qalandarov  for the review, it was amazing!

bitchat/Shared/KeyStorableProtocol.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+    
+    func get<T>(_ key: String) -> T? {
+        defaults.object(forKey: key) as? T
+    }
+    
+    func remove(_ key: String) {
+        defaults.removeObject(forKey: key)
+    }
+}
```
qalandarov: Is this file pending for deletion?

bitchat/Shared/StorageProtocol.swift
```diff
@@ -0,0 +1,46 @@
...
+///
+final class UserDefaultsStorage: StorageProtocol {
+    
+    private let defaults: UserDefaults
+    
+    init?(suiteName: String) {
+        guard let defaults = UserDefaults(suiteName: suiteName) else {
+            return nil
+        }
+        self.defaults = defaults
```
qalandarov, line 25: I'd make `init(defaults: UserDefaults = .standard)` as the default initializer and make the failable one as a convenience init and use `self.init(defaults: defaults)` so in case if you have custom logic you wouldn't have to repeat in both inits and have a single source of truth

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,43 +4,45 @@ import XCTest
...
-        super.tearDown()
+        sut = GeohashBookmarksStore(
+            store: store
+        )
```
qalandarov, line 15: `store = GeohashBookmarksStore(store: UserDefaultsStorageMock())!` so each test starts with a fresh mock and set `store = nil` in each `tearDown`

bitchatTests/GeohashBookmarksStoreTests.swift
```diff
@@ -4,43 +4,45 @@ import XCTest
...
     }
 
     func testToggleAndNormalize() {
-        let store = GeohashBookmarksStore.shared
+        guard let sut else {
+            XCTFail("Sut of GeohashBookmarksStore not initialized")
+            return
+        }
```
qalandarov: keeping implicitly unwrapped optional would clean this up (and using `store` variable wouldn't require below changes)... and once we move to Swift Testing we could use `try require(...)` for a safer replacement for `!`

bitchatTests/GeohashBookmarksStoreTests.swift
qalandarov: I was referring to the actual code of the `_resetForTesting()` in `GeohashBookmarksStore`:

```
    #if DEBUG
    /// Testing-only reset helper
    func _resetForTesting() {
        bookmarks.removeAll()
        membership.removeAll()
        bookmarkNames.removeAll()
        persist()
        persistNames()
    }
    #endif
```

bitchat/BitchatApp.swift
```diff
@@ -137,23 +138,22 @@ struct BitchatApp: App {
...
-              let sharedDate = userDefaults.object(forKey: "sharedContentDate") as? Date else {
+        guard let sharedContent: String = storage.get("sharedContent"),
+              let sharedDate: Date = storage.get("sharedContentDate") else {
             return
         }
         
         // Only process if shared within configured window
         if Date().timeIntervalSince(sharedDate) < TransportConfig.uiShareAcceptWindowSeconds {
-            let contentType = userDefaults.string(forKey: "sharedContentType") ?? "text"
+            let contentType: String = storage.get("sharedContentType") ?? "text"
```
qalandarov: nit: the type can be inferred from `?? "text"` here

bitchat/Nostr/GeoRelayDirectory.swift
```diff
@@ -44,7 +47,7 @@ final class GeoRelayDirectory {
     // MARK: - Remote Fetch
     func prefetchIfNeeded() {
         let now = Date()
-        let last = UserDefaults.standard.object(forKey: lastFetchKey) as? Date ?? .distantPast
+        let last: Date = storage.get(lastFetchKey) ?? .distantPast
```
qalandarov: nit: the type can be inferred from `?? .distantPast` here

bitchat/BitchatApp.swift
```diff
@@ -137,23 +138,22 @@ struct BitchatApp: App {
...
     private func checkForSharedContent() {
         // Check app group for shared content from extension
-        guard let userDefaults = UserDefaults(suiteName: BitchatApp.groupID) else {
+        guard let storage = UserDefaultsStorage(suiteName: BitchatApp.groupID) else {
             return
         }
-        
-        guard let sharedContent = userDefaults.string(forKey: "sharedContent"),
-              let sharedDate = userDefaults.object(forKey: "sharedContentDate") as? Date else {
+        guard let sharedContent: String = storage.get("sharedContent"),
```
qalandarov, line 145: Are we going to have statically-typed keys instead of raw strings?

bitchat/Shared/KeyStorableProtocol.swift
r1b2ns: was deleted

bitchatTests/GeohashBookmarksStoreTests.swift
r1b2ns: Fixed, but I have to keep the store as a variable class because in the testPersistenceWritten function we test if persistence works.

qalandarov: 2 things here:
1) `store` is referring to `GeohashBookmarksStore` and `UserDefaultsStorage` should be used as `storage` to be consistent with other places and to avoid confusion.
2) each test should work on isolation and as you can see in that test we're "toggling" the keys and we should check if that is resulting in the values being saved in the storage (and the results of other tests should never affect other tests)

bitchat/BitchatApp.swift
r1b2ns: I think it’s fine to keep strings for the keys for now, and in the next PR we can move them into an enum so this one doesn’t get too big or hard to test.

bitchatTests/GeohashBookmarksStoreTests.swift
r1b2ns: Yeah, you’re right, I got a little confused here with the words store and storage.

Thinking about what you said regarding isolation in tests, it doesn’t make sense to test storage persistence here, since the focus of GeohashBookmarksStoreTests is to test the functionalities of GeohashBookmarksStore. So the testPersistenceWritten doesn’t really make sense here, because the persistence of GeohashBookmarksStore was already covered in testToggleAndNormalize. 

I’ve implemented the StorageProtocolTests to provide test coverage for this protocol.

qalandarov: I think that test was testing persistence of multiple keys at the same time to make sure they're not affecting each other

bitchatTests/StorageProtocolTests.swift
```diff
@@ -0,0 +1,48 @@
...
+//  StorageProtocolTests.swift
+//  bitchat
+//
+//  Created by Rubens Machion on 26/09/25.
+//
+
+import XCTest
+@testable import bitchat
+
+final class StorageProtocolTests: XCTestCase {
```
qalandarov: I think here we might be testing the functionality of a mock, not sure how much value is in this? :)

r1b2ns: I tried testing StorageProtocol and reusing the mock I created. Do you think it’s okay if we also test just UserDefaultsStorage?

qalandarov: I was initially thinking to create a protocol that would cause UserDefaults conform to it directly so in the inits we'd accept that protocol and pass UserDefaults as a default parameter. This way all we'd have to do is mock that protocol without having to create another wrapper around UserDefaults:

Something like:
```
final class SomeClass {
    let storage: StorageProtocol
    init(storage: StorageProtocol = UserDefaults.standard) {
        self.storage = storage
    }
}
```

r1b2ns: Hi man, sorry for the delay. I think using this approach could create some problems. With the current implementation we can already mock StorageProtocol, and in the future we could add other types of storage decoupled from UserDefaults.

Making UserDefaults conform directly to a protocol may seem practical in the short term, but it leads to coupling, less flexibility, and more fragile tests. A thin wrapper is a small extra cost that buys a lot of architectural freedom.

qalandarov: Both have pros and cons, if you write a wrapper you just have to test it, if you wrap UserDefaults - you don't have to test it as it's Apple's job.

r1b2ns: What do you think? Should we implement this wrapper or leave it as it is?

qalandarov: It's your call, but if I were you I'd put UserDefaults behind a protocol as all we have is 2/3 functions.

Whichever way you go, just rebase and make it ready and see what @jackjackbits says

r1b2ns:
Thanks again @qalandarov for the review, it was really helpful.

@jackjackbits  I’ve just rebased this PR.

jackjackbits:
do we need this after #741? seems they solve the same problem.

qalandarov:
> do we need this after #741? seems they solve the same problem.

#741 was specifically targeting 1 file whereas this concept was to encapsulate the entire usage of UserDefaults so we could replace it with a different storage (like db) in the future if we wanted to, but it's your call @jackjackbits

r1b2ns:
Rebased to resolve conflicts 

cc: @qalandarov @jackjackbits

r1b2ns:
Hey guys, any chance we can move this forward? Just asking because if not, I’ll close this PR and won’t rebase it with main


@qalandarov @jackjackbits

qalandarov:
> Hey guys, any chance we can move this forward? Just asking because if not, I’ll close this PR and won’t rebase it with main
> 
> @qalandarov @jackjackbits

Hey, sorry for the late response. Given that UserDefaults is no longer a blocking issue we can either polish this and introduce as a clean-up or maybe work on centralizing all the storage-related code as we were thinking maybe to upgrade to iOS 17 so we can introduce Swift Data

PR #640, opened 2025-09-18
Refactor: Statically-typed `Peer` to unify the logic of `peerID`'s

qalandarov:
_Disclaimer: Given that `peerID` is **everywhere**, the PR is going to be large, but I've been trying to keep the individual commits as isolated and small as possible to simplify the review process_

---
## Problem

There're a lot of usages of `peerID` property which is mainly represented as String, but sometimes we have Data conversions, different length comparisons, prefixes of different sizes with hard-coded numbers, and utilities in different files taking care of different functionalities.

## Proposal

Create a struct to represent the Peer object to provide static-type safety, unify all the logic in a single place, and make it testable.

## Benefits
Statically-typed object, compile-time checks, centralized logic to avoid repetition, no more "magic string" comparisons, testable object...

---

There're a few other Peer objects (e.g. BitchatPeer, PeerInfo, TransportPeer...) which could all be consolidated somehow to simplify the concept of a "peer" to the contributors.

Once this conversion is done we can do a similar excersise to unify public key representation to avoid having String/Data passed around and manually calculated/converted throughout the code.


qalandarov:
These are the main changes:
- function arguments: `peerID: String` -> `peer: Peer`
- variable names: `peerID` -> `peer` or `peer.id` depending on the use case
- some variables that represent other "peer" objects like `PeerInfo` and `BitchatPeer` were renamed to avoid confusion
- `peerID == anotherPeerID` -> `peer == anotherPeer`
- `peerID.count == 16` -> `peer.isShort`
- `peerID.hasPrefix("nostr:")` -> `peer.isNostrColon` (need more descriptive naming, but at least we're getting compile-time checks
- `Peer` object conforms to `ExpressibleByStringLiteral` and `ExpressibleByStringLiteral` so it can be created from string literals or interpolations or even compared against
- it also conforms to `CustomStringConvertible` for backwards compatibility, so if it's used inside a string without accessing `.id` property it just works
- it conforms to `Equatable`, `Comparable`, `Hashable` so it can be compared and used as Dictionary Key or Set Element
- Removed a few utility classes and consolidated the logic of creation and validation into a single place
- Other minor clean ups removing some dead code and simplifying some logic (but I tried to avoid touching the logic as much as possible, except a few instances)

There're still areas for improvement, but the goal of this PR was to convert all the use cases of `peerID` from a primitive `String` to statically-typed `Peer`

---
I manually reviewed the code, tested and verified the app on an iPhone and a Mac for the basic functionalities like detecting available peers, seeing the names, ability to message in the group, mention, hug, DM.

jackjackbits:
did you also check relay behavior? multiple devices relaying over distance.

here's from gpt5:
Findings

  - Peer.isValid (bitchat/Models/Peer.swift:64) now rejects any
  ID containing : because the allowed character set no longer
  includes that symbol. We routinely build Peer values such as
  nostr:<pubkey-prefix> for geo timeline senders. Any code path
  that calls NoiseEncryptionService.initiateHandshake(with:)
  or processHandshakeMessage(from:) (bitchat/Services/
  NoiseEncryptionService.swift:397,420) with one of those peers will
  immediately throw invalidPeerID, cutting off the handshake; the old validator
  accepted the same IDs.
  - MessageRouter no longer routes Nostr fallbacks for the nostr_… conversation
  IDs produced by geo DMs. resolveRecipientNpub(for:) (bitchat/Services/
  NostrTransport.swift:173) only handles 64‑hex and 16‑hex peers, so when
  sendPrivate or sendReadReceipt is invoked with the nostr_… peer (see bitchat/
  ViewModels/ChatViewModel.swift:2166, 2203), it now returns nil and silently
  drops the message instead of falling back to Nostr. Previously these paths
  used the stored Nostr mapping.
  - This PR rewrites 50 files (~2.3k insertions / 2.5k deletions) across
  core transports, identity, storage, and views without companion regression
  coverage. There are no new XCTest cases validating the new Peer type or the
  routing refactors, so even if the above issues were addressed, we would still
  be flying blind on a massive behavior change.

  Recommendation
  Not safe to merge. Please fix the concrete regressions above, add focused
  tests that exercise the new peer-ID plumbing (especially DM routing /
  handshake flows), and then re-request review.

qalandarov:
> did you also check relay behavior? multiple devices relaying over distance.
I haven't, what would be the best way to test these scenarios? My main assumption was that if I were to wrap the strings into a struct then the behavior shouldn't change.

Regarding the first point:
- Peer.isValid - it's a clone of the old [InputValidator.validatePeerID](https://github.com/permissionlesstech/bitchat/blob/main/bitchat/Utils/InputValidator.swift#L20) implementation and double checked with [these tests](https://github.com/permissionlesstech/bitchat/blob/main/bitchatTests/Utils/InputValidatorTests.swift)
- Could you double check if these new tests are expected behavior or are they also going to cause the issues that this point is describing: #659

Regarding the second point:
- If I understood it correctly, I changed all the `peerID.count == 16` with `peer.isShort` (which internally checks for both count == 16 && valid hex) - I can separate the count check vs hex to make sure the existing logic is not changed, and eventually we can unify them separately

Regarding the third point:
- I'll add more tests.
- I'll do a gradual migration which I initially thought could bring the app to an inconsistent state

qalandarov:
Doing gradual migration starting with #688

PR #659, opened 2025-09-21
InputValidator to handle prefixes explicitly

qalandarov:
@jackjackbits, could you double check if these new tests are following expected behavior or are they also going to cause the issues that are described in the first point of #640 because I cloned this validator into Peer.swift (and can clone these new tests too)?

Or if I didn't understand the expectation correctly, could you give me example valid inputs and example broken inputs so we can add them to tests and make sure the new implementation also has them?

### DIFF
--- a/bitchat/Utils/InputValidator.swift
+++ b/bitchat/Utils/InputValidator.swift
@@ -20,6 +20,13 @@ struct InputValidator {
     
     /// Validates a peer ID from any source (short 16-hex, full 64-hex, or internal alnum/-/_ up to 64)
     static func validatePeerID(_ peerID: String) -> Bool {
+        // Handle different prefixes
+        if peerID.starts(with: "nostr_") || peerID.starts(with: "nostr:") || peerID.starts(with: "noise:") {
+            return validatePeerID(String(peerID.dropFirst(6)))
+        } else if peerID.starts(with: "mesh:") || peerID.starts(with: "name:") {
+            return validatePeerID(String(peerID.dropFirst(5)))
+        }
+        
         // Accept short routing IDs (exact 16-hex)
         if PeerIDResolver.isShortID(peerID) { return true }
         // If length equals short-hex length but isn't valid hex, reject
--- a/bitchatTests/Utils/InputValidatorTests.swift
+++ b/bitchatTests/Utils/InputValidatorTests.swift
@@ -33,5 +33,42 @@ final class InputValidatorTests: XCTestCase {
         let tooLong = String(repeating: "a", count: 65)
         XCTAssertFalse(InputValidator.validatePeerID(tooLong))
     }
+
+    func test_valid_prefixes() {
+        let hex64 = String(repeating: "a", count: 64)
+        XCTAssertTrue(InputValidator.validatePeerID("noise:\(hex64)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr:\(hex64)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr_\(hex64)"))
+        
+        let hex63 = String(repeating: "a", count: 63)
+        XCTAssertTrue(InputValidator.validatePeerID("noise:\(hex63)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr:\(hex63)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr_\(hex63)"))
+        
+        let hex16 = String(repeating: "a", count: 16)
+        XCTAssertTrue(InputValidator.validatePeerID("noise:\(hex16)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr:\(hex16)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr_\(hex16)"))
+        
+        let hex8 = String(repeating: "a", count: 8)
+        XCTAssertTrue(InputValidator.validatePeerID("noise:\(hex8)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr:\(hex8)"))
+        XCTAssertTrue(InputValidator.validatePeerID("nostr_\(hex8)"))
+        
+        let mesh = "mesh:abcdefg"
+        XCTAssertTrue(InputValidator.validatePeerID("name:\(mesh)"))
+        
+        let name = "name:some_name"
+        XCTAssertTrue(InputValidator.validatePeerID("name:\(name)"))
+        
+        let badName = "name:bad:name"
+        XCTAssertFalse(InputValidator.validatePeerID("name:\(badName)"))
+        
+        // Too long
+        let hex65 = String(repeating: "a", count: 65)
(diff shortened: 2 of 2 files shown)


qalandarov:
Given that we're adding the prefix onto a valid peerID, should we update the validation logic to validate the string after the prefix?
Also, when creating a `nostr:` key we're using 8-character keys (`nostrShortKeyDisplayLength`), should we include that in the validation because currently we're only checking for 16 and 64?

bitchat/Utils/InputValidator.swift
```diff
@@ -20,6 +20,13 @@ struct InputValidator {
     
     /// Validates a peer ID from any source (short 16-hex, full 64-hex, or internal alnum/-/_ up to 64)
     static func validatePeerID(_ peerID: String) -> Bool {
+        // Handle different prefixes
+        if peerID.starts(with: "nostr_") || peerID.starts(with: "nostr:") || peerID.starts(with: "noise:") {
+            return validatePeerID(String(peerID.dropFirst(6)))
```
qalandarov, line 25: Ideally we should have specific-size tests for these as they're always created with identical sizes, something like this:

```
        // Format: "nostr:<8-character valid hex>"
        if peerID.starts(with: "nostr:") {
            let id = peerID.dropFirst(6)
            return id.count == TransportConfig.nostrShortKeyDisplayLength && Data(hexString: id) != nil
        }
        
        // Format: "nostr_<16-character valid hex>"
        if peerID.starts(with: "nostr_") {
            let id = peerID.dropFirst(6)
            return id.count == TransportConfig.nostrConvKeyPrefixLength && Data(hexString: id) != nil
        }
        
        // Format: "noise:<64-character valid hex>"
        if peerID.starts(with: "noise:") {
            let id = peerID.dropFirst(6)
            return id.count == Limits.maxPeerIDLength && Data(hexString: id) != nil
        }
```

qalandarov: what do you think, @jackjackbits?

jackjackbits:
got this from sonnet-4.5. not sure if this pr is out of date or intention?

⚠️ NOT SAFE TO MERGE - Logic Bug

  This PR has a critical recursive logic bug that will cause infinite 
  recursion and stack overflow.

  The Bug:

  static func validatePeerID(_ peerID: String) -> Bool {
      // Handle different prefixes
      if peerID.starts(with: "nostr_") || peerID.starts(with: "nostr:") ||
  peerID.starts(with: "noise:") {
          return validatePeerID(String(peerID.dropFirst(6)))  // ⚠️ 
  RECURSIVE CALL
      } else if peerID.starts(with: "mesh:") || peerID.starts(with:
  "name:") {
          return validatePeerID(String(peerID.dropFirst(5)))  // ⚠️ 
  RECURSIVE CALL
      }

  The Problem:

  Example 1: Input "mesh:mesh:abcd"
  1. Strips first "mesh:" → recursively calls with "mesh:abcd"
  2. Strips second "mesh:" → recursively calls with "abcd"
  3. Finally validates "abcd" ✓

  Example 2: Input "name:name:name:test"
  1. Strips first "name:" → recursive call with "name:name:test"
  2. Strips second "name:" → recursive call with "name:test"
  3. Strips third "name:" → recursive call with "test"
  4. Validates "test" ✓

  While this technically works, it's inefficient and dangerous:
  - Unbounded recursion on deeply nested prefixes
  - No recursion limit
  - Stack overflow risk with malicious input like "name:".repeated(10000) +
   "test"

  Test Issues:

  The test on line 55-56 is wrong:
  let mesh = "mesh:abcdefg"
  XCTAssertTrue(InputValidator.validatePeerID("name:\(mesh)"))  // 
  "name:mesh:abcdefg"

  This tests "name:mesh:abcdefg" which:
  1. Strips "name:" → recursively validates "mesh:abcdefg"
  2. Strips "mesh:" → recursively validates "abcdefg"
  3. Passes ✓

  But this behavior is unintentional and exploitable.

  Recommendation:

  The recursive prefix stripping should be replaced with a loop or only 
  strip prefixes once. The current implementation allows infinite nesting
  which is a security vulnerability.

  Request changes from the author to fix the recursion issue before
  merging.

qalandarov:
> got this from sonnet-4.5. not sure if this pr is out of date or intention?
> 
> ⚠️ NOT SAFE TO MERGE - Logic Bug
> 
> This PR has a critical recursive logic bug that will cause infinite recursion and stack overflow.
> 
> The Bug:
> 
> static func validatePeerID(_ peerID: String) -> Bool { // Handle different prefixes if peerID.starts(with: "nostr_") || peerID.starts(with: "nostr:") || peerID.starts(with: "noise:") { return validatePeerID(String(peerID.dropFirst(6))) // ⚠️ RECURSIVE CALL } else if peerID.starts(with: "mesh:") || peerID.starts(with: "name:") { return validatePeerID(String(peerID.dropFirst(5))) // ⚠️ RECURSIVE CALL }
> 
> The Problem:
> 
> Example 1: Input "mesh:mesh:abcd"
> 
> 1. Strips first "mesh:" → recursively calls with "mesh:abcd"
> 2. Strips second "mesh:" → recursively calls with "abcd"
> 3. Finally validates "abcd" ✓
> 
> Example 2: Input "name:name:name:test"
> 
> 1. Strips first "name:" → recursive call with "name:name:test"
> 2. Strips second "name:" → recursive call with "name:test"
> 3. Strips third "name:" → recursive call with "test"
> 4. Validates "test" ✓
> 
> While this technically works, it's inefficient and dangerous:
> 
> * Unbounded recursion on deeply nested prefixes
> * No recursion limit
> * Stack overflow risk with malicious input like "name:".repeated(10000) +
>   "test"
> 
> Test Issues:
> 
> The test on line 55-56 is wrong: let mesh = "mesh:abcdefg" XCTAssertTrue(InputValidator.validatePeerID("name:(mesh)")) // "name:mesh:abcdefg"
> 
> This tests "name:mesh:abcdefg" which:
> 
> 1. Strips "name:" → recursively validates "mesh:abcdefg"
> 2. Strips "mesh:" → recursively validates "abcdefg"
> 3. Passes ✓
> 
> But this behavior is unintentional and exploitable.
> 
> Recommendation:
> 
> The recursive prefix stripping should be replaced with a loop or only strip prefixes once. The current implementation allows infinite nesting which is a security vulnerability.
> 
> Request changes from the author to fix the recursion issue before merging.

2 things:
1. I think the AI is missing the background context which is the peerID's can either be bare/raw or prefix:bare, and technically it is correct that I'm not "breaking" the recursion, but since the peerID is created by us and not passed by the user - it should be safe.
2. I already migrated this logic to the new [`PeerID.isValid`](https://github.com/permissionlesstech/bitchat/blob/main/bitchat/Models/PeerID.swift#L138) so this PR can be closed (and this file will be removed when I raise this PR [`PeerID 6/n`](https://github.com/permissionlesstech/bitchat/commit/ee2c5b7f5bf1ffb4938a1fc2f7d8092dcdc2d72b))

qalandarov:
#743 migrates it to the new `PeerID`

PR #669, opened 2025-09-24
interactive geohash map viewer

atridahmetaj:
Adds a WebKit-based map interface for visual geohash selection, bringing iOS/macOS parity with the Android app.


jackjackbits:
great start. let's make it a lot more snappy and intuitive first. should be able to tap on a geo rect and zoom in automatically and such.

atridahmetaj:
> great start. let's make it a lot more snappy and intuitive first. should be able to tap on a geo rect and zoom in automatically and such.

Thanks! @jackjackbits now its handling selection and zoom in by tap and double tap, also control buttons on the map.

atridahmetaj:
hello @jackjackbits , @qalandarov do you think this is okay to be merged, or if there is something to be done?

qalandarov:
Once you resolve the conflicts I think @jackjackbits would have to take another look as I'm not really sure what feature is this :)

PR #681, opened 2025-09-25
Feature/ble binary transfer

jackjackbits:
## Summary

  - Implement end-to-end BLE file transfer support:
      - Add BitchatFilePacket TLV codec and FileTransferLimits.
      - Extend BLEService for fragment tracking, transfer progress, incoming
  media persistence, and fragment reassembly.
      - Introduce TransferProgressManager and update ChatViewModel to manage
  voice/image/file sends, message state, and transfer cancellation.
      - Enrich ContentView with media attachment UI (voice recorder, image/
  file handling, delivery status) and new BlockRevealImageView, VoiceNoteView,
  FileAttachmentView.
  - Platform-specific refinements:
      - iOS/macOS share the same resizing via ImageUtils; mac encoding now
  flushes metadata, normalizes colorspace, and targets ~60 KB JPEGs.
      - macOS attachments use PhotosPicker, NSSavePanel for exports, user-
  selected read/write entitlement, and a consistent photo library prompt.
      - Recording HUD shows milliseconds; swiping a revealed image re-blurs it.
  - UX polish & stability:
      - Track transfers by queue to avoid collisions when sending duplicate
  files; clean temp files on failure.
      - Lowercase image preview buttons, improve message header formatting,
  ensure application support directories exist before use.
      - Add long-run tests (NotificationStreamAssemblerTests,
  BitchatFilePacketTests) covering fragment assembly and TLV round-trips.
  - Security/permissions:
      - Update mac entitlements (Bluetooth, Photos, microphone, user-selected
  read/write) and Info.plist usage descriptions for microphone & photo library
  access.

  ## Testing

  - Unit Tests: xcodebuild test passes with new
  NotificationStreamAssemblerTests and BitchatFilePacketTests.
  - Manual (iOS simulator + macOS app)
      - Send voice notes and images both directions; verify progress
  indicators, thumbnails, and swipe-to-reblur behavior.
      - Confirm duplicate file sends map to correct messages and can be
  cancelled individually.
      - Use mac Photos picker and drag File importer; check thumbnails render
  instantly.
      - Save an image from the mac preview via NSSavePanel—file lands at chosen
  destination, temp files cleaned.

  Let me know if you’d like additional screenshots or size comparison data
  before publishing the PR.

### DIFF
--- a/bitchat/Views/ContentView.swift
+++ b/bitchat/Views/ContentView.swift
@@ -73,7 +73,7 @@ struct ContentView: View {
     @State private var recordingTimer: Timer?
     @State private var recordingStartDate: Date?
     @State private var showFileImporter = false
-#if canImport(PhotosUI) && ((os(iOS) && compiler(>=5.7)) || (os(macOS) && compiler(>=5.7)))
+#if os(iOS)
     @State private var showPhotoPicker = false
     @State private var selectedPhotoPickerItem: PhotosPickerItem?
 #endif
@@ -210,7 +210,7 @@ struct ContentView: View {
                 FingerprintView(viewModel: viewModel, peerID: peerID)
             }
         }
-#if canImport(PhotosUI) && ((os(iOS) && compiler(>=5.7)) || (os(macOS) && compiler(>=5.7)))
+#if os(iOS)
         .photosPicker(isPresented: $showPhotoPicker, selection: $selectedPhotoPickerItem, matching: .images)
         .onChange(of: selectedPhotoPickerItem) { newItem in
             guard let item = newItem else { return }
@@ -229,7 +229,7 @@ struct ContentView: View {
             }
         }
         .confirmationDialog("Attach", isPresented: $showAttachmentActions, titleVisibility: .visible) {
-#if canImport(PhotosUI) && ((os(iOS) && compiler(>=5.7)) || (os(macOS) && compiler(>=5.7)))
+#if os(iOS)
             Button("Image") {
                 showAttachmentActions = false
                 DispatchQueue.main.async { showPhotoPicker = true }
@@ -1982,7 +1982,7 @@ private extension ContentView {
         }
     }
 
-#if canImport(PhotosUI) && ((os(iOS) && compiler(>=5.7)) || (os(macOS) && compiler(>=5.7)))
+#if os(iOS)
     func handlePhotoSelection(_ item: PhotosPickerItem) async {
         defer { Task { @MainActor in selectedPhotoPickerItem = nil } }
         do {


bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+        let scaled = scaledImage(image, maxDimension: maxDimension)
+        var quality = compressionQuality
+        guard var jpegData = scaled.jpegData(compressionQuality: quality) else {
+            throw ImageUtilsError.encodingFailed
+        }
+        while jpegData.count > targetImageBytes && quality > 0.3 {
+            quality -= 0.1
+            guard let next = scaled.jpegData(compressionQuality: quality) else { break }
+            jpegData = next
+        }
```
qalandarov: we could probably have some gain if we were to use binary-search kind of approach. In case if the image is huge we'd have to run this processing more than if we were to half the quality each time

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+
+    static func processImage(at url: URL, maxDimension: CGFloat = 512) throws -> URL {
+        let data = try Data(contentsOf: url)
+        #if os(iOS)
+        guard let image = UIImage(data: data) else { throw ImageUtilsError.invalidImage }
+        return try processImage(image, maxDimension: maxDimension)
+        #else
+        guard let image = NSImage(data: data) else { throw ImageUtilsError.invalidImage }
+        return try processImage(image, maxDimension: maxDimension)
+        #endif
```
qalandarov, line 37: not sure if this would work:

```
        #if os(iOS)
        guard let image = UIImage(data: data) else { throw ImageUtilsError.invalidImage }
        #else
        guard let image = NSImage(data: data) else { throw ImageUtilsError.invalidImage }
        #endif
        return try processImage(image, maxDimension: maxDimension)
```

also, we could have 2 other options about situations like this:
1) create extensions on the actual image types (UIImage/NSImage)
2) create a custom typealias and assign it to platform-specific image type and make functions accept that type
```
enum ImageUtils {
    #if os(iOS)
    typealias Image = UIImage
    #else
    typelias Image = NSImage
    #endif

    func process(_ image: Image) { ... }
```

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+        return outputURL
+    }
+
+    private static func scaledImage(_ image: UIImage, maxDimension: CGFloat) -> UIImage {
+        let size = image.size
+        let maxSide = max(size.width, size.height)
+        guard maxSide > maxDimension else { return image }
+        let scale = maxDimension / maxSide
+        let newSize = CGSize(width: size.width * scale, height: size.height * scale)
+        UIGraphicsBeginImageContextWithOptions(newSize, true, 1.0)
```
qalandarov, line 68: `1.0` could affect the "retina resolution"

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+        try jpegData.write(to: outputURL, options: .atomic)
+        return outputURL
+    }
+
+    private static func scaledImage(_ image: UIImage, maxDimension: CGFloat) -> UIImage {
+        let size = image.size
+        let maxSide = max(size.width, size.height)
+        guard maxSide > maxDimension else { return image }
+        let scale = maxDimension / maxSide
+        let newSize = CGSize(width: size.width * scale, height: size.height * scale)
```
qalandarov, line 67: We might want to get the integer version of the sides to avoid sub-pixel sizes

        let newSize = CGSize(width: Int(size.width * scale), height: Int(size.height * scale))

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+        guard let image = NSImage(data: data) else { throw ImageUtilsError.invalidImage }
+        return try processImage(image, maxDimension: maxDimension)
+        #endif
+    }
+
+    #if os(iOS)
+    static func processImage(_ image: UIImage, maxDimension: CGFloat = 512) throws -> URL {
+        let scaled = scaledImage(image, maxDimension: maxDimension)
+        var quality = compressionQuality
+        guard var jpegData = scaled.jpegData(compressionQuality: quality) else {
```
qalandarov: we might want to check if the image is already small enough to avoid unnecessary processing

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+        let height = inputCG.height
+        let colorSpace = CGColorSpace(name: CGColorSpace.sRGB) ?? CGColorSpaceCreateDeviceRGB()
+        guard let context = CGContext(
+            data: nil,
+            width: width,
+            height: height,
+            bitsPerComponent: 8,
+            bytesPerRow: 0,
+            space: colorSpace,
+            bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
```
qalandarov: I think this approach works for both platforms and could simplify the code by unifying the code

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+    private static func scaledImage(_ image: NSImage, maxDimension: CGFloat) -> NSImage {
+        let size = image.size
+        let maxSide = max(size.width, size.height)
+        guard maxSide > maxDimension else { return image }
+        let scale = maxDimension / maxSide
+        let newSize = NSSize(width: size.width * scale, height: size.height * scale)
+        let scaledImage = NSImage(size: newSize)
+        scaledImage.lockFocus()
+        image.draw(in: NSRect(origin: .zero, size: newSize),
+                   from: NSRect(origin: .zero, size: size),
```
qalandarov, line 126: We could probably unify the scaling math to avoid duplicating the logic and avoiding any room for bugs

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+    #endif
+
+    private static func makeOutputURL() throws -> URL {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyyMMdd_HHmmss"
+        let fileName = "img_\(formatter.string(from: Date())).jpg"
+
+        let directory = try applicationFilesDirectory().appendingPathComponent("images/outgoing", isDirectory: true)
+        try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true, attributes: nil)
+        return directory.appendingPathComponent(fileName)
```
qalandarov, line 160: given that we're dealing with file manager in multiple places of the app it makes sense to have a separate utility helper for it

bitchat/Features/voice/VoiceNotePlaybackController.swift
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    // MARK: - Private Helpers
+
+    private func preparePlayer(for url: URL) {
+        do {
+            let player = try AVAudioPlayer(contentsOf: url)
+            player.delegate = self
+            player.prepareToPlay()
+            self.player = player
```
qalandarov, line 115: we could probably have a non-optional AVPlayer object and replace its item each time url changes?

bitchat/Features/voice/VoiceNotePlaybackController.swift
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    private func startTimer() {
+        if timer != nil { return }
+        timer = Timer.scheduledTimer(withTimeInterval: 0.05, repeats: true) { [weak self] _ in
+            self?.updateProgress()
+        }
+        if let timer = timer {
+            RunLoop.main.add(timer, forMode: .common)
+        }
```
qalandarov, line 151: Using [`addPeriodicTimeObserver(forInterval:queue:using:)`](https://developer.apple.com/documentation/avfoundation/avplayer/addperiodictimeobserver(forinterval:queue:using:)) is probably better

bitchat/Features/voice/VoiceRecorder.swift
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+        #elseif os(macOS)
+        return await withCheckedContinuation { continuation in
+            AVCaptureDevice.requestAccess(for: .audio) { granted in
+                continuation.resume(returning: granted)
+            }
+        }
+        #else
+        return true
```
qalandarov, line 43: Shouldn't this be false to make sure we don't support platforms that we're not explicitly handling?

bitchat/Features/voice/VoiceRecorder.swift
```diff
@@ -0,0 +1,169 @@
...
+                recorder.stop()
+                self.cleanupSession()
+                let url = self.currentURL
+                self.recorder = nil
+                self.currentURL = url
+                completion(url)
+            }
+            self.stopWorkItem = item
+            self.queue.asyncAfter(deadline: .now() + self.paddingInterval, execute: item)
+        }
```
qalandarov, line 113: since we're already inside a custom queue and it's async - should we just go ahead and do this right here?

```
queue.asyncAfter(deadline: .now() + paddingInterval) { [weak self in
    self?.recorder?.stop()
    self?.cleanupSession()
    let url = self?.currentURL
    self?.recorder = nil
    self?.currentURL = nil
    completion(url)
}
```

bitchat/Features/voice/VoiceRecorder.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    private func applicationFilesDirectory() throws -> URL {
+        #if os(iOS)
+        return try FileManager.default.url(for: .applicationSupportDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
+            .appendingPathComponent("files", isDirectory: true)
+        #else
+        let base = try FileManager.default.url(for: .applicationSupportDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
+        return base.appendingPathComponent("files", isDirectory: true)
+        #endif
+    }
```
qalandarov, line 162: this seems very similar to the ImageUtils

bitchat/Features/voice/VoiceRecorder.swift
```diff
@@ -0,0 +1,169 @@
...
+            let session = AVAudioSession.sharedInstance()
+            guard session.recordPermission == .granted else {
+                throw RecorderError.microphoneAccessDenied
+            }
+            try session.setCategory(
+                .playAndRecord,
+                mode: .default,
+                options: [.defaultToSpeaker, .allowBluetoothA2DP, .allowBluetoothHFP]
+            )
+            try session.setActive(true, options: .notifyOthersOnDeactivation)
```
qalandarov, line 65: isn't `.notifyOthersOnDeactivation` only when deactivating the session?

bitchat/Features/voice/Waveform.swift
```diff
@@ -0,0 +1,95 @@
+import AVFoundation
+import Foundation
+import BitLogger
+
+/// Generates and caches downsampled waveforms for audio files so UI rendering is cheap.
+final class WaveformCache {
```
qalandarov, line 6: I think a lot of these files dealing with custom queues could benefit from Swift Actors

bitchat/Features/voice/Waveform.swift
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            guard let computed = self.computeWaveform(url: url, bins: bins) else {
+                DispatchQueue.main.async { completion([]) }
+                return
+            }
+
+            self.queue.async(flags: .barrier) {
+                self.cache[url] = computed
+            }
```
qalandarov: why not just call `self.cache[url] = computed` directly as we're already inside the queue + we have the `computed` right above?

bitchat/Info.plist
```diff
@@ -37,6 +37,10 @@
 	<string>bitchat uses Bluetooth to discover and connect with other bitchat users nearby.</string>
 	<key>NSCameraUsageDescription</key>
 	<string>bitchat uses the camera to scan QR codes to verify peers.</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>bitchat lets you pick images from your photo library to share with nearby peers.</string>
```
qalandarov, line 41: I think there's a way to ask users to share photos without asking for permission

bitchat/Protocols/BinaryProtocol.swift
```diff
@@ -105,10 +105,23 @@ extension Data {
...
+    static let v2HeaderSize = 16
     static let senderIDSize = 8
     static let recipientIDSize = 8
     static let signatureSize = 64
+
+    static func headerSize(for version: UInt8) -> Int {
+        switch version {
+        case 1: return v1HeaderSize
+        case 2: return v2HeaderSize
+        default: return 0
```
SkinnnyJay: Binary Protocol Changes (Critical)
```
  // bitchat/Protocols/BinaryProtocol.swift:108-122
  static func headerSize(for version: UInt8) -> Int {
      switch version {
      case 1: return v1HeaderSize
      case 2: return v2HeaderSize
      default: return 0  // ⚠️ ISSUE: Returns 0 for unknown versions
      }
  }
```
  **Suggestion:** Return an error or throw for unknown versions instead of silently returning 0,
  which could cause buffer underflows.

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+    static func processImage(_ image: UIImage, maxDimension: CGFloat = 512) throws -> URL {
+        let scaled = scaledImage(image, maxDimension: maxDimension)
+        var quality = compressionQuality
+        guard var jpegData = scaled.jpegData(compressionQuality: quality) else {
+            throw ImageUtilsError.encodingFailed
+        }
+        while jpegData.count > targetImageBytes && quality > 0.3 {
+            quality -= 0.1
+            guard let next = scaled.jpegData(compressionQuality: quality) else { break }
+            jpegData = next
```
SkinnnyJay: Memory Management in Image Processing
```
  // bitchat/Features/media/ImageUtils.swift:34-41
  while jpegData.count > targetImageBytes && quality > 0.3 {
      quality -= 0.1
      guard let next = scaled.jpegData(compressionQuality: quality) else { break }
      jpegData = next  // ⚠️ ISSUE: Potential memory spikes
  }
```
  **Suggestion:** Consider using autoreleasepool for memory-intensive operations and implement
  progressive quality reduction to avoid memory spikes.

bitchat/Features/voice/VoiceNotePlaybackController.swift
```diff
@@ -0,0 +1,168 @@
...
+            currentTime = 0
+            progress = 0
+        }
+    }
+
+    private func ensurePlayerReady() -> Bool {
+        if player == nil {
+            preparePlayer(for: url)
+        }
+        #if os(iOS)
```
SkinnnyJay, line 132:
```
 // liines: 108-115
  #if os(iOS)
  try session.setCategory(
      .playAndRecord,
      mode: .default,
      options: [.defaultToSpeaker, .allowBluetoothA2DP, .allowBluetoothHFP]
  )
  #endif
```
  **Suggestion:** Consider using `.mixWithOthers` option to allow background audio to continue
  playing during recording.

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,149 @@
...
+        var quality = compressionQuality
+        guard var jpegData = encodeJPEG(from: cgImage, quality: quality) else {
+            throw ImageUtilsError.encodingFailed
+        }
+        while jpegData.count > targetImageBytes && quality > 0.3 {
+            quality -= 0.1
+            guard let next = encodeJPEG(from: cgImage, quality: quality) else { break }
+            jpegData = next
+        }
+        let outputURL = try makeOutputURL()
```
SkinnnyJay: File Cleanup Strategy
 ```
 // Multiple locations with temp file creation
  let outputURL = try makeOutputURL()
```
**Suggestion:** Implement a consistent cleanup strategy for temp files, especially on app
  backgrounding or crashes.

File: bitchat/Features/media/ImageUtils.swift [hunk 22]
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    private static func encodeJPEG(from cgImage: CGImage, quality: CGFloat) -> Data? {
+        guard let data = CFDataCreateMutable(nil, 0) else {
+            return nil
+        }
+        guard let destination = CGImageDestinationCreateWithData(data, UTType.jpeg.identifier as CFString, 1, nil) else {
+            return nil
+        }
+        let options: [CFString: Any] = [
```
SkinnnyJay, line 142: Image Metadata Handling
```
  let options: [CFString: Any] = [
      kCGImageDestinationLossyCompressionQuality: quality,
      kCGImagePropertyExifDictionary: [:],  // ✅ Good: Strips EXIF
      kCGImagePropertyTIFFDictionary: [:],
      kCGImagePropertyIPTCDictionary: [:]
  ]
``` 
**Good:** Properly strips metadata from images.

bitchat/Features/voice/Waveform.swift
```diff
@@ -0,0 +1,95 @@
...
+                DispatchQueue.main.async { completion(cached) }
+                return
+            }
+
+            guard let computed = self.computeWaveform(url: url, bins: bins) else {
+                DispatchQueue.main.async { completion([]) }
+                return
+            }
+
+            self.queue.async(flags: .barrier) {
```
SkinnnyJay:
```
self.queue.async(flags: .barrier) {
    self.cache[url] = computed. // ⚠️ ISSUE: dd cache size limits and LRU eviction 
}
```
**Suggestion:** Add cache size limits and LRU eviction to prevent unbounded memory growth.

bitchatTests/NotificationStreamAssemblerTests.swift
```diff
@@ -108,4 +108,67 @@ final class NotificationStreamAssemblerTests: XCTestCase {
         }
         XCTAssertEqual(decoded.timestamp, packet.timestamp)
     }
+
+    func testAssemblesCompressedLargeFrame() throws {
+        var assembler = NotificationStreamAssembler()
+
+        let largeContent = Data(repeating: 0x41, count: 2_500_000)
```
SkinnnyJay:
```
// bitchatTests/NotificationStreamAssemblerTests.swift:testAssemblesCompressedLargeFrame
let largeContent = Data(repeating: 0x41, count: 2_500_000) // ❌ Exceeds 1 MiB limit
```
**Suggestion:** Reduce to 700_000 bytes to stay under the limit while still triggering compression.

bitchat/Views/Media/FileAttachmentView.swift
```diff
@@ -0,0 +1,118 @@
...
+            Image(systemName: "doc.fill")
+                .foregroundColor(Color.blue)
+                .font(.bitchatSystem(size: 24))
+
+            VStack(alignment: .leading, spacing: 4) {
+                Text(fileName)
+                    .font(.bitchatSystem(size: 14, weight: .medium))
+                    .foregroundColor(.primary)
+                    .lineLimit(2)
+                Text(url.path)
```
SkinnnyJay: `Text(url.path) // ❌ Exposes full system paths`

**Suggestion:** Use `url.lastPathComponent` to show the filename only.

bitchatTests/NotificationStreamAssemblerTests.swift
```diff
@@ -108,4 +108,67 @@ final class NotificationStreamAssemblerTests: XCTestCase {
...
+            signature: nil,
+            ttl: 3,
+            version: 2
+        )
+
+        guard let frame = packet.toBinaryData(padding: false) else {
+            return XCTFail("Failed to encode packet frame")
+        }
+
+        let flagsOffset = 1 + 1 + 1 + 8
```
SkinnnyJay:
```
let flagsOffset = 1 + 1 + 1 + 8 // ❌ Magic numbers
let lengthOffset = 12
```
**Fix:** Centralize in BinaryProtocol.Offsets struct.

bitchat/Features/voice/Waveform.swift
SkinnnyJay: Should the URL be hashed or something?

bitchat/Info.plist
SkinnnyJay: Yes! You're thinking of `PHPickerViewController` (iOS 14+) and `UIImagePickerController`. These allow users to select photos without requiring permission to access the photo library.

### PHPickerViewController (Recommended)

```
import PhotosUI

var configuration = PHPickerConfiguration()
configuration.selectionLimit = 1
configuration.filter = .images

let picker = PHPickerViewController(configuration: configuration)
picker.delegate = self
present(picker, animated: true)
```

#### Benefits:
- No permissions required
- User controls what they share
- More privacy-focused
- Modern iOS-style interface

#### UIImagePickerController (Legacy)

```
let picker = UIImagePickerController()
picker.sourceType = .photoLibrary
picker.delegate = self
present(picker, animated: true)
```

**Key Point:** Both present a system picker where users explicitly choose what to share - no broad photo library access needed. The app only gets the specific photos the user selects.

This approach offers a better user experience than requesting full photo library permission, which many users deny due to privacy concerns.

bitchat/Features/media/ImageUtils.swift
SkinnnyJay: Good point! Binary search would be more efficient. The current approach could do 5-6 iterations for large images. We could implement bisection to find optimal quality faster.

bitchat/Features/media/ImageUtils.swift
SkinnnyJay: Great suggestions! The typealias approach would clean this up nicely:
```
#if os(iOS)
typealias PlatformImage = UIImage
#else
typealias PlatformImage = NSImage
#endif
```
Should refactor to reduce duplication.

bitchat/Features/media/ImageUtils.swift
SkinnnyJay: You're right - using 1.0 forces non-retina rendering. Should use 0.0 to maintain device scale or calculate an appropriate scale for the target size. Should get fixed.

bitchat/Features/media/ImageUtils.swift
SkinnnyJay: Good catch! Sub-pixel sizes can cause rendering issues. Converting to Int will ensure clean pixel boundaries.

bitchat/Features/media/ImageUtils.swift
SkinnnyJay: Absolutely - should check original data size first before any processing—no point compressing a 30KB image when the target is 60KB.

bitchat/Features/voice/Waveform.swift
SkinnnyJay: Good point! Swift Actors would be cleaner than manual queue management. The current `DispatchQueue(label:, attributes: .concurrent)` approach works but Actors would provide better type safety and eliminate the need for explicit `queue.sync/queue.async` calls. Could refactor in a follow-up.

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    static func processImage(at url: URL, maxDimension: CGFloat = 512) throws -> URL {
+        // Security H1: Check file size BEFORE reading into memory
+        let attrs = try FileManager.default.attributesOfItem(atPath: url.path)
+        guard let fileSize = attrs[.size] as? Int else {
+            throw ImageUtilsError.invalidImage
+        }
+        // Allow up to 10MB source images (will be scaled down)
+        guard fileSize <= 10 * 1024 * 1024 else {
+            throw ImageUtilsError.invalidImage
```
qalandarov, line 27: I'd suggest adding "image is too large" to make the error message more meaningful to the user

bitchat/Features/media/ImageUtils.swift
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        let data = try Data(contentsOf: url)
```
qalandarov, line 30: Reading the entire data into the memory and then decoding with UIImage(data:) would take a lot of memory. Also, how about WebP as it's smaller and supports transparency + animations?

GPT is suggesting to encode right on disk with something like this:

```
import UIKit
import ImageIO
import MobileCoreServices

func encodeWebP(from image: UIImage, quality: CGFloat = 0.8) -> Data? {
    let mutableData = NSMutableData()
    guard let destination = CGImageDestinationCreateWithData(mutableData, AVFileType.webp as CFString, 1, nil) else {
        return nil
    }

    let options: [CFString: Any] = [
        kCGImageDestinationLossyCompressionQuality: quality,
        kCGImagePropertyHasAlpha: true // allow transparency
    ]

    if let cgImage = image.cgImage {
        CGImageDestinationAddImage(destination, cgImage, options as CFDictionary)
    } else if let ciImage = image.ciImage {
        let context = CIContext()
        if let cgImage = context.createCGImage(ciImage, from: ciImage.extent) {
            CGImageDestinationAddImage(destination, cgImage, options as CFDictionary)
        }
    }

    guard CGImageDestinationFinalize(destination) else { return nil }
    return mutableData as Data
}
```

Usage:
```
func processImageToWebP(at url: URL, maxSize: CGSize) -> Data? {
    // 1. Downsample at decode time
    guard let image = downsampleImage(at: url, to: maxSize) else { return nil }

    // 2. Encode to WebP
    return encodeWebP(from: image, quality: 0.8)
}
```

qalandarov:
Thinking about this PR more I think you're working on 3 chained features in a single PR:
1. Ability to send/receive raw binary data through BLE
2. Ability to record and send audios (+ UI changes)
3. Ability to select and send [downsampled] images (+ UI changes)

---
Speaking of chaining: GitHub has a [feature](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches?utm_source=chatgpt.com#working-with-branches) where it automatically updates the chained PR's base branch if the first PR is merged *and* the branch is deleted (which can and should already be automated from the [Settings](https://github.com/permissionlesstech/bitchat/settings#merge_types_delete_branch)).

Here's the diagram:


---
If you combine the above automation with [protecting the main branch](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule#creating-a-branch-protection-rule) you could allow more people to have right access to the repo (or a specific branch pattern) to speed up the collaboration.

jackjackbits:
@qalandarov seems peerID work is failing CI?

qalandarov:
Other prs are fine. Let me look into what’s going on in this pr

qalandarov:
> @qalandarov seems peerID work is failing CI?

It looks like this PR hasn't been rebased with the recent changes like extracting NotificationStreamAssembler and rearranging BLEService's functions

Maybe it could be easier to cherry pick the new changes onto the main in a different branch

qalandarov:
ok, looks like if we rebase this branch onto main without #801 and then apply #801's changes onto this branch it generates way smaller conflicts than the other way around... let me take a stab at it and I'll create another PR to this branch that should solve all the conflicts

qalandarov:
Created 2 PRs by manually rebasing: #808 and using Codex: #809

PR #684, opened 2025-09-25
Fix Localizations

qalandarov:
- Uses a centralized `L10n.string` instead of `NSLocalizedString` (including the Share Extension)
- Unifies `L10n.string` and `L10n.format`
- Works with both `bitchat.xcodeproj` and `Package.swift` so we fix issues like [this](https://github.com/permissionlesstech/bitchat/actions/runs/18020736536/job/51277229770?pr=683#step:5:171)

---

I also propose to rename `L10n` to `Localization` as it's easier to type a single word then mixing an uppercase letter, digits and a lowercase letter


jackjackbits:
thanks!

Issue #515, opened 2025-08-25
grey bar over text input on macos


derpsteb:
<img width="293" height="548" alt="Image" src="https://github.com/user-attachments/assets/1c004a43-5d94-47f6-8cf5-d34d5ee6b454" />

Version: 1.3.1 (9)

When opening the app on macos I get a grey bar over the text input field as soon as I click on the input field. I can still type, but don't see what is written.

dim13:
Same with 1.3.3 (3)

MichelDiz:
Same here!

SkinnnyJay:
Thanks for the report! I can take a look. To help debug, could you share a bit more detail:

- **macOS version/build**: (`sw_vers`)
- **Chip type**: Apple Silicon or Intel?
- **Display setup**: Are you on the built-in display or external monitor(s)? If external, what resolution/scaling setting?
- **Accessibility settings**: Is *Reduce transparency* or *Increase contrast* enabled under **System Settings → Accessibility → Display**?
- **Other contexts**: Does this happen only in bitchat's input field, or also in other text fields/apps?

This info will help narrow down whether it’s a macOS display/compositor issue, an accessibility interaction, or something specific to our text field implementation.

SkinnnyJay:
I was unable to reproduce this issue. Can you see if upgrading to 1.4.2 resolves your problem?

### **My test stats:**

- **macOS version/build:** 15.6.1 (24G90)
- **Chip type:** Apple Silicon
- **Display:** built-in display
- **Accessibility settings:** Tested both, including Light/Dark mode.
 - **Tested:** 1.4.2 - 1.3.1, could not reproduce this.

dim13:
macOS version/build: 15.6.1 & tahoe 26
Chip type: Apple Silicon
Display setup: built-in
Accessibility settings: default
Other contexts: only bitchat

Still reproducible with latest AppStore version (1.4.1)

RaptorHacker11000:
Same problem here on a MacBook Air M2 running Sequoia 15.6.1.

SkinnnyJay:
# Bug #515: Gray Overlay on Text Fields in iOS App on macOS

## How to Reproduce

1. Download the iOS version of BitChat from the App Store and run it on a Mac
2. Go to System Settings > Accessibility > Display
3. Make sure "Increase Contrast" is turned OFF
4. Open the BitChat app
5. Click on the message input field
6. Notice that the text field has a gray overlay that makes typed text invisible
[...]

PR #690, opened 2025-09-27
Fix: iOS TextField gray overlay when running on macOS with Increase Contrast disabled

SkinnnyJay:
# Fix: iOS TextField gray overlay when running on macOS with Increase Contrast disabled

This PR fixes issue #515, where text becomes invisible in the message input field when the iOS app is running on macOS with the "Increase Contrast" accessibility setting disabled. The fix adds iOS-specific modifiers to prevent the gray overlay from appearing on text fields.

## Changes
- Added `.autocorrectionDisabled(true)` modifier to prevent predictive text overlay
- Added `.focusEffectDisabled()` modifier to disable the system focus effect
- Wrapped these modifiers in `#if os(iOS)` to ensure they only apply to the iOS build

## Testing
- Verified the fix works on iOS apps running on macOS with "Increase Contrast" disabled
- Confirmed no regressions in native macOS app functionality
- All tests pass successfully


SkinnnyJay:
@jackjackbits just a heads up: we’ve seen 5+ reports of this issue tied to how the app is bundled and delivered via the App Store, which ends up serving it as an iOS build.

jackjackbits:
but why are we turning off autocorrection for people? that would be unexpected behavior.

SkinnnyJay:
@codex review

SkinnnyJay:
@codex review

SkinnnyJay:
@codex review round 3

PR #702, opened 2025-09-30
Fix broken build + uncover localization tests

qalandarov:
Trying run tests when the project is opened with bitchat.xcodeproj immediately fails because `LocalizationCatalogTests.swift` has an error, but CI didn't catch it because it was under `Localization` folder and SPM was treating it like a resource as per Package.swift line 64.

I also added an explicit checks to `LocationNotesManagerTests.swift` to verify that the expected output is not a localization key to avoid false positives.

---
Note: this PR is going to cause CI to fail because now all the localization tests are uncovered and some of them are broken


SkinnnyJay:
LGTM!

PR #738, opened 2025-10-01
Shared schemes

qalandarov:
Shared schemes are used to sync some of the settings between different team members so we can enable / disable certain features without everyone repeating them on their machines.

This PR also enables parallelized and randomized test execution.

It also gather code coverage to help us understand the state of the code base + hopefully encourage more contributors to start adding tests as there would be a clear action items they could take.

| Before | After |
| - | - |
| Screenshot 2025-09-30 at 4 30 50 PM | Screenshot 2025-09-30 at 4 31 05 PM |

Result:
Screenshot 2025-09-30 at 4 40 06 PM


jackjackbits:
excellent. thank you.

PR #739, opened 2025-10-01
PeerID 4/n: `BitchatMessage.senderPeerID` + String equality

qalandarov:


qalandarov:
The broken test is related to `UserDefaults` and #636 or #741 should make it more deterministic

PR #740, opened 2025-10-01
Temporarily disable broken tests to catch actual issues

qalandarov:
Following #702 - I think it's better to temporarily comment them out so we can catch actual issues with CI


qalandarov:
The only broken test now is related to UserDefaults and #636 should make it more deterministic

PR #741, opened 2025-10-01
Injectable UserDefaults to fix race condition in tests

qalandarov:
We're having flaky tests because the tests can run in parallel and we're using a `UserDefaults` singleton which could cause race condition. Passing a unique instance for each test should guarantee isolation and less flakiness.


SkinnnyJay:
MVP!

PR #748, opened 2025-10-03
Swift Testing

qalandarov:
Not updating the GitHub action yet because they fail when they run in parallel due to the mesh-mocking architecture

### DIFF
--- a/bitchatTests/EndToEnd/PrivateChatE2ETests.swift
+++ b/bitchatTests/EndToEnd/PrivateChatE2ETests.swift
@@ -6,115 +6,81 @@
 // For more information, see <https://unlicense.org>
 //
 
-import XCTest
+import Testing
 import CryptoKit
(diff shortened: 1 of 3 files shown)


qalandarov:
@jackjackbits while I appreciate the enthusiasm, but this PR wasn't ready as it hasn't enabled the "swift testing" from the GitHub action since the tests can't run in parallel because of the MockBLEService's global registry.

I guess I now have to quickly refactor that and fix and enable the tests 😂

qalandarov:
> @jackjackbits while I appreciate the enthusiasm, but this PR wasn't ready as it hasn't enabled the "swift testing" from the GitHub action since the tests can't run in parallel because of the MockBLEService's global registry.
> 
> I guess I now have to quickly refactor that and fix and enable the tests 😂

just saw #764

PR #752, opened 2025-10-04
PeerID 13/n: Tidy up `Transport`-related code and files

qalandarov:
Chained after #751

---

Eventually, we can create a dedicated module for BLE and make all those `private` functions as `internal` utilities so we can test them


jackjackbits:
conflicts. still don't know what you mean by chaining PRs or how to do them

qalandarov:
> conflicts

Looks like you closed it with a comment :) - I created another PR after resolving the conflicts: #761 

> still don't know what you mean by chaining PRs or how to do them

Chaining PRs doesn't work for those of us who can only use forks and don't have write access to the repo.

For you and those with write access this is how it works:
1. you create a branch `A` and create a PR: `A -> main`
2. then create a branch `B` from `A` and create a PR to A: `B -> A` (not `B -> main`)
3. when `A` is merged into `main` + `A` is deleted, then GitHub automatically updates `B -> A` to be `B -> main`

Btw, the branches are still not deleted after merge: you can tick [this checkbox](https://github.com/permissionlesstech/bitchat/settings#merge_types_delete_branch)

---

My current chaining:
1. same as above
2. here if I'm creating `B -> main` because if create `B -> A` you wouldn't see it as it would only live in my fork
3. when `A` is merged, `main` is updated and `B -> main` is seeing conflicts

jackjackbits:
hm. that box is checked.

qalandarov:
Interesting… my branches are not deleted after merge (you can check the
ones you just merged)

Maybe because they’re in my fork GitHub is not removing them - as it’s
technically a different repo 🤔

PR #757, opened 2025-10-04
Extract `DeliveryStatusView` into a separate file

qalandarov:
#Preview result:

Screenshot 2025-10-04 at 9 36 38 PM


bitchat/Protocols/BitchatProtocol.swift
```diff
@@ -131,7 +131,7 @@ enum LazyHandshakeState {
 // MARK: - Delivery Status
 
 // Delivery status for messages
-enum DeliveryStatus: Codable, Equatable {
+enum DeliveryStatus: Codable, Equatable, Hashable {
```
qalandarov, line 134: This was required to be able to support ForEach (used in a #Preview)

Issue #703, opened 2025-09-30
Localization Test Issues


qalandarov:
@SkinnnyJay - could you please check #702 and fix the localization issues?

Also, could you please keep the indentation to standard Xcode (4 spaces) for the sake of consistency? :)

Lastly, it looks like the tests were created by an AI and seems over-engineered, can you check if you can simplify it so others can also improve/fix/add more tests in the future?

SkinnnyJay:
Yeah, I will revisit the test and clean it up. Added a few ways to help find issues with untranslated strings and ensure that we respect placeholders, reserved words etc.

@qalandarov, are there any functional issues here or tests you don't think we should worry about in the revised version you are requesting?

qalandarov:
> [@qalandarov](https://github.com/qalandarov), are there any functional issues here or tests you don't think we should worry about in the revised version you are requesting?

I actually never unit tested localizations and not sure if that's even a practice in the industry, so can't comment on the content of the tests, but one thing I realized is that I wasn't able to understand much of the code with very complex functions and return types. Keeping functions as simple/flat as possible and maybe introducing custom types with meaningful property names would be better than returning nested dicts and the like.

SkinnnyJay:
Yes, I'd like to break the tests out into smaller, isolated test files, which would make grepping them easier. I was following existing patterns established. I can break up the test into 1-3 logically grouped types of tests for localization.

Some of the things the test does:

0. Check if only the English key were added.
1. Do we have all the keys needed?
2. Does the string respect the placeholder and plural rule
3. Does the package load?
4. Can we iterate through each language, pull strings, and assert that they are translated?
5. Checks for untranslated strings

qalandarov:
Yeah, maybe creating separate files to test different functionality would simplify it too.

I wanted to break up the functions with 3-level nested for-loops and dicts like `[String: [String: [String: [String]]]]` :)

qalandarov:
Broken tests are commented out in #740

jackjackbits:
Closing as resolved/obsolete against the current `main` and v1.7.1-era architecture. The implementation has changed materially since this report and the reported gap is now covered. If this still reproduces on a current release, please open a focused report with the app version, OS/device, exact steps, and relevant logs.

PR #763, opened 2025-10-06
Addresses #703 by simplifying localization tests and fixing issues

SkinnnyJay:
### Overview
**Closes:** #703 by simplifying localization tests and fixing critical issues.

### Changes

- Focus on value: 26 tests → 8 high-value tests that catch real localization issues
- Remove complexity: Eliminate dead code, duplicate helpers, and meaningless validations
- Improve maintainability: Clear sections with documented purpose

### Issues Resolved

- ✅ Tests now discoverable by Swift Package Manager
- ✅ Schema mismatch between `JSON` config and test expectations fixed
- ✅ Removed over-engineered runtime tests that only validated Swift's built-in APIs
- ✅ Eliminated duplicate validation logic across multiple files
- ✅ Replaced unsafe `try!` with proper error handling

### Test Categories (8 focused tests)

- Configuration: Key existence, expected values validation
- Structure: Locale parity, translation completeness
- Content: Non-empty translations, expected value matching
- Format: Placeholder consistency across locales

### Notes
- While testing, the test uncovered that after enabling this test, several localized strings are missing in the `bitchatShareExtension\Localization\Localizable.xcstrings`. These seemed to be introduced by https://github.com/permissionlesstech/bitchat/commit/b873b1910452d8c624c2a3e80a861a0383612321

### Next Steps
- Will follow up with a PR to address the observed missing strings.

### DIFF
--- a/bitchatTests/LocalizationTests.swift
+++ b/bitchatTests/LocalizationTests.swift
@@ -13,7 +13,6 @@ import Foundation
 /// Comprehensive localization tests for bitchat.
 /// Validates configuration integrity, structural consistency, content quality, and format consistency.
 /// 
-@Suite
 struct LocalizationTests {
     
     // MARK: - Static Configuration Loading
@@ -21,44 +20,28 @@ struct LocalizationTests {
     private static let testsDirectoryURL = URL(fileURLWithPath: #filePath).deletingLastPathComponent()
     private static let repoRootURL = URL(fileURLWithPath: #filePath).deletingLastPathComponent().deletingLastPathComponent()
     
+    /// Cached configuration for reuse across tests (non-crashing on failure)
+    private static let testConfig: LocalizationTestConfig? = {
+        do {
+            return try loadConfig(testsDirectoryURL: testsDirectoryURL)
+        } catch {
+            return nil
+        }
+    }()
+
     /// Dynamically loaded enabled locales from configuration
     private static let enabledLocales: [String] = {
-        do {
-            let config = try loadConfig(testsDirectoryURL: testsDirectoryURL)
+        if let config = testConfig {
             return getEnabledTestLocales(from: config)
-        } catch {
-            // Log the error for debugging while providing graceful fallback
-            print("⚠️ Failed to load localization config: \(error). Falling back to [\"en\"]")
+        } else {
+            // Provide a minimal fallback for parameterized tests without crashing the suite
+            print("⚠️ Failed to load localization config. Falling back to [\"en\"] for parameterized tests")
             return ["en"]
         }
     }()
     
-    /// Cached configuration for reuse across tests
-    private static let testConfig: LocalizationTestConfig = {
-        do {
-            return try loadConfig(testsDirectoryURL: testsDirectoryURL)
-        } catch {
-            // Fail fast with clear context - localization config is critical for proper test coverage
-            fatalError("""
-                ❌ LOCALIZATION CONFIG MISSING ❌
-                
-                Error: \(error)
-                
-                Expected: \(testsDirectoryURL.appendingPathComponent("LocalizationTestsConfig.json"))
-                
-                This file is required for comprehensive localization testing across all supported locales.
-                Without it, we cannot validate translations, placeholder consistency, or locale completeness.
-                
-                To fix: Ensure LocalizationTestsConfig.json exists in bitchatTests/ directory.
-                """)
-        }
-    }()
-    
     // MARK: - Instance Variables
-    
-    private let testsDirectoryURL = Self.testsDirectoryURL
-    private let repoRootURL = Self.repoRootURL
-    
+
     /// Path to the main app's localization catalog
     private let appCatalogPath = "bitchat/Localizable.xcstrings"
     
@@ -69,9 +52,12 @@ struct LocalizationTests {
     
     /// Tests that configured keys actually exist in their respective catalogs.
     @Test func configuredKeysExistInCatalogs() throws {
-        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
-        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
-        let config = Self.testConfig
+        guard let config = Self.testConfig else {
+            Issue.record("Missing localization config at bitchatTests/LocalizationTestsConfig.json")
+            return
+        }
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         
         // Verify all required app keys exist in app catalog
         let allAppKeys = getAllAppKeys(from: config)
@@ -90,9 +76,12 @@ struct LocalizationTests {
     
     /// Tests that expected values reference valid keys and locales.
     @Test func expectedValuesConfiguration() throws {
-        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
-        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
-        let config = Self.testConfig
+        guard let config = Self.testConfig else {
+            Issue.record("Missing localization config at bitchatTests/LocalizationTestsConfig.json")
+            return
+        }
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         
         for (locale, localeConfig) in config.testLocales {
             guard localeConfig.enabled else { continue }
@@ -117,19 +106,19 @@ struct LocalizationTests {
     /// Ensures every locale includes exactly the same keys as base locale.
     @Test func catalogLocaleParity() throws {
         // Test app catalog
-        let appContext = try loadContextWithKeys(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let appContext = try loadContextWithKeys(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
         assertLocaleParity(context: appContext, catalogName: "App")
         
         // Test share extension catalog
-        let shareExtContext = try loadContextWithKeys(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let shareExtContext = try loadContextWithKeys(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         assertLocaleParity(context: shareExtContext, catalogName: "ShareExtension")
     }
     
     /// Validates that a specific enabled locale has complete translations.
     @Test("Locale completeness", arguments: enabledLocales)
     func localeCompleteness(locale: String) throws {
-        let appContext = try loadContextWithKeys(relativePath: appCatalogPath, repoRootURL: repoRootURL)
-        let shareExtContext = try loadContextWithKeys(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let appContext = try loadContextWithKeys(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
+        let shareExtContext = try loadContextWithKeys(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         
         let baseLocale = appContext.baseLocale
         let baseAppKeys = appContext.keysByLocale?[baseLocale] ?? Set()
@@ -139,27 +128,30 @@ struct LocalizationTests {
         if locale == baseLocale { return }
         
         // Verify locale is present in both catalogs
-        #expect(appContext.locales.contains(locale), 
+        expectOrRecord(appContext.locales.contains(locale), 
                "Locale '\(locale)' missing from app catalog")
-        #expect(shareExtContext.locales.contains(locale), 
+        expectOrRecord(shareExtContext.locales.contains(locale), 
                "Locale '\(locale)' missing from share extension catalog")
         
         // Verify locale has same number of keys as base locale
         let appLocaleKeys = appContext.keysByLocale?[locale] ?? Set()
-        #expect(appLocaleKeys.count == baseAppKeys.count, 
+        expectOrRecord(appLocaleKeys.count == baseAppKeys.count, 
                "Locale '\(locale)' app catalog missing keys compared to \(baseLocale)")
         
         let shareLocaleKeys = shareExtContext.keysByLocale?[locale] ?? Set()
-        #expect(shareLocaleKeys.count == baseShareKeys.count, 
+        expectOrRecord(shareLocaleKeys.count == baseShareKeys.count, 
                "Locale '\(locale)' share extension catalog missing keys compared to \(baseLocale)")
     }
     
     /// Validates that a specific enabled locale has translated state for all required keys.
     @Test("Locale translation state", arguments: enabledLocales)
     func localeTranslationState(locale: String) throws {
-        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
-        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
-        let config = Self.testConfig
+        guard let config = Self.testConfig else {
+            Issue.record("Missing localization config at bitchatTests/LocalizationTestsConfig.json")
+            return
+        }
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         
         // Skip base locale as it's the source
         if locale == appCatalog.sourceLanguage { return }
@@ -174,7 +166,7 @@ struct LocalizationTests {
                 continue
             }
             
-            #expect(unit.state == "translated", 
+            expectOrRecord(unit.state == "translated", 
                    "Required app key '\(key)' not marked as translated in enabled locale '\(locale)' (state: '\(unit.state)')")
         }
         
@@ -188,7 +180,7 @@ struct LocalizationTests {
                 continue
             }
             
-            #expect(unit.state == "translated", 
+            expectOrRecord(unit.state == "translated", 
                    "Required share extension key '\(key)' not marked as translated in enabled locale '\(locale)' (state: '\(unit.state)')")
         }
     }
@@ -198,17 +190,20 @@ struct LocalizationTests {
     
     /// Guards required strings from going empty per enabled locale.
     @Test func requiredKeysNonEmpty() throws {
-        let config = Self.testConfig
+        guard let config = Self.testConfig else {
+            Issue.record("Missing localization config at bitchatTests/LocalizationTestsConfig.json")
+            return
+        }
         let enabledLocales = Self.enabledLocales
         
         // Test app required keys
-        let appContext = try loadContext(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let appContext = try loadContext(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
         let allAppKeys = getAllAppKeys(from: config)
         assertRequiredKeysPresent(context: appContext, keys: allAppKeys, 
                                 enabledLocales: enabledLocales, catalogName: "App")
         
         // Test share extension required keys
-        let shareExtContext = try loadContext(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let shareExtContext = try loadContext(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         let allShareKeys = getAllShareExtensionKeys(from: config)
         assertRequiredKeysPresent(context: shareExtContext, keys: allShareKeys, 
                                 enabledLocales: enabledLocales, catalogName: "ShareExtension")
@@ -217,9 +212,12 @@ struct LocalizationTests {
     /// Validates that a specific locale contains expected string values.
     @Test("Locale expected values", arguments: enabledLocales)
     func localeExpectedValues(locale: String) throws {
-        let appContext = try loadContext(relativePath: appCatalogPath, repoRootURL: repoRootURL)
-        let shareExtContext = try loadContext(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
-        let config = Self.testConfig
+        guard let config = Self.testConfig else {
+            Issue.record("Missing localization config at bitchatTests/LocalizationTestsConfig.json")
+            return
+        }
+        let appContext = try loadContext(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
+        let shareExtContext = try loadContext(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         
         guard let localeConfig = config.testLocales[locale],
               !localeConfig.assertValues.isEmpty else {
@@ -248,17 +246,20 @@ struct LocalizationTests {
     
     /// Verifies format placeholders stay consistent across locales for required keys.
     @Test func placeholderConsistency() throws {
-        let config = Self.testConfig
+        guard let config = Self.testConfig else {
+            Issue.record("Missing localization config at bitchatTests/LocalizationTestsConfig.json")
+            return
+        }
         let enabledLocales = Self.enabledLocales
         
         // Test app catalog placeholder consistency
-        let appContext = try loadContextWithPlaceholders(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let appContext = try loadContextWithPlaceholders(relativePath: appCatalogPath, repoRootURL: Self.repoRootURL)
         let allAppKeys = getAllAppKeys(from: config)
         assertPlaceholderConsistencyForKeys(context: appContext, keys: allAppKeys, 
                                           enabledLocales: enabledLocales, catalogName: "App")
         
         // Test share extension catalog placeholder consistency
-        let shareExtContext = try loadContextWithPlaceholders(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let shareExtContext = try loadContextWithPlaceholders(relativePath: shareExtCatalogPath, repoRootURL: Self.repoRootURL)
         let allShareKeys = getAllShareExtensionKeys(from: config)
         assertPlaceholderConsistencyForKeys(context: shareExtContext, keys: allShareKeys, 
                                           enabledLocales: enabledLocales, catalogName: "ShareExtension")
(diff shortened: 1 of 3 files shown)


SkinnnyJay:
@codex, review this.

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,582 @@
...
+  
+  private let testsDirectoryURL = URL(fileURLWithPath: #filePath).deletingLastPathComponent()
+  private let repoRootURL = URL(fileURLWithPath: #filePath).deletingLastPathComponent().deletingLastPathComponent()
+  
+  // MARK: - Configuration Validation
+  
+  /// Tests that configured keys actually exist in their respective catalogs.
+  func testConfiguredKeysExistInCatalogs() throws {
+    let appCatalog = try loadCatalog(relativePath: "bitchat/Localizable.xcstrings")
+    let shareCatalog = try loadCatalog(relativePath: "bitchatShareExtension/Localization/Localizable.xcstrings")
```
qalandarov: - maybe you could have these as instance variables?
- maybe call `shareCatalog` either `shareExtCatalog` or `extensionCatalog` to avoid it sound like a "shared" catalog?
- is it not possible to configure the localization strings from the packages instead of hard-coding?
- also, would be nice to keep the indentation to 4 spaces for the sake of consistency :)

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,582 @@
...
+                        "Placeholder mismatch for key \(key) at \(path) in locale \(locale) (\(catalogName))", 
+                        file: file, line: line)
+        }
+      }
+    }
+  }
+  
+  // MARK: - Context Loading Helpers
+  
+  private func loadContext(relativePath: String) throws -> CatalogContext {
```
qalandarov: would be nice to extract the utility functions into a separate file to keep this file focused on tests

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,582 @@
...
+      }
+    }
+    return tokens
+  }
+}
+
+// MARK: - Supporting Types
+
+/// Represents a string catalog file structure (.xcstrings format)
+private struct StringCatalog: Decodable {
```
qalandarov: types would definitely benefit from separate files

qalandarov:
btw, #740 commented out localization-related tests from `LocationNotesManagerTests.swift` as well

bitchatTests/LocalizationTests.swift
SkinnnyJay: @qalandarov How do you feel about we add a `.editorconfig`  at the repo root to make CLI, Cursor, Xcode, and any other editor consistent?

**Contents:**
```
root = true

[*.swift]
indent_style = space
indent_size = 4
```
This way we enforce consistency.

_Can be added as a small PR separate from this_

SkinnnyJay: FWIW from GPT:

> ```
> The project currently **doesn't have automated formatting enforcement**. Here are the options:
> 
> ### **Option 1: SwiftFormat (Recommended)**
> ```bash
> # Install SwiftFormat
> brew install swiftformat
> 
> # Create .swiftformat config file
> echo "
> --indent 4
> --indentcase false
> --trimwhitespace always
> --insertlines true
> --removelines true
> --allman false
> --wraparguments preserve
> --wrapcollections preserve
> --closingparen same-line
> --commas inline
> --semicolons never
> --ranges no-space
> --empty void
> --hexliteralcase uppercase
> --exponentcase lowercase
> --fractiongrouping disabled
> --binarygrouping 4,8
> --octalgrouping 4,8
> --hexgrouping 4,8
> " > .swiftformat
> ```
> 
> ### **Option 2: SwiftLint (Style Checking)**
> ```bash
> # Install SwiftLint
> brew install swiftlint
> 
> # Create .swiftlint.yml
> echo "
> rules:
>   - indentation_width
> indentation_width:
>   indentation_width: 4
> " > .swiftlint.yml
> ```
> 
> ### **Option 3: EditorConfig (Editor-level)**
> ```bash
> # Create .editorconfig
> echo "
> root = true
> 
> [*.swift]
> indent_style = space
> indent_size = 4
> end_of_line = lf
> charset = utf-8
> trim_trailing_whitespace = true
> insert_final_newline = true
> " > .editorconfig
> ```
> 
> ### **Option 4: Xcode Settings**
> - In Xcode: Preferences → Text Editing → Indentation
> - Set "Prefer indent using" to "Spaces" 
> - Set "Tab width" and "Indent width" to 4
> 
> ### **Recommendation:**
> I'd suggest **SwiftFormat** as it can both check and automatically fix formatting issues. It integrates well with Git hooks and CI/CD pipelines.
> 
>

qalandarov: We can introduce swiftlint, but integrating it now would give us a lot of errors in the current state of the repo (although we could lax all the rules now and gradually tighten them up)

qalandarov: Xcode's default indentation is already 4 and you can update Cursor/VSCode to be 4 as well.

I was already running swiftlint locally but wanted to introduce a bit later due to the issues mentioned above.

SkinnnyJay:
@qalandarov feedback addressed 🤞:

- ✅ Extract supporting types to separate file
- ✅ Extract utility functions to separate file
- ✅ Convert hardcoded paths to instance variables
- ✅ Rename confusing variable names
- ✅ Update indentation to 4 spaces
- ✅ Follow project conventions with TestUtilities/ placement
- ✅ Maintain all test functionality

Feel free to resolve any threads that have been addressed by these changes.

qalandarov:
after rebasing, do you think you could convert the tests to SwiftTesting from the get go so we don't have to convert it later?

SkinnnyJay:
@qalandarov anything left or we good to go?

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+//
+
+import Testing
+import Foundation
+@testable import bitchat
+
+/// Comprehensive localization tests for bitchat.
+/// Validates configuration integrity, structural consistency, content quality, and format consistency.
+/// 
+@Suite
```
qalandarov: is there a benefit from adding this if we're not customizing it? I thought the struct was already grouping it

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+    
+    // MARK: - Static Configuration Loading
+    
+    private static let testsDirectoryURL = URL(fileURLWithPath: #filePath).deletingLastPathComponent()
+    private static let repoRootURL = URL(fileURLWithPath: #filePath).deletingLastPathComponent().deletingLastPathComponent()
+    
+    /// Dynamically loaded enabled locales from configuration
+    private static let enabledLocales: [String] = {
+        do {
+            let config = try loadConfig(testsDirectoryURL: testsDirectoryURL)
```
qalandarov: maybe you could reuse `testConfig` from below to avoid unnecessary repetition?

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+        }
+    }()
+    
+    /// Cached configuration for reuse across tests
+    private static let testConfig: LocalizationTestConfig = {
+        do {
+            return try loadConfig(testsDirectoryURL: testsDirectoryURL)
+        } catch {
+            // Fail fast with clear context - localization config is critical for proper test coverage
+            fatalError("""
```
qalandarov: `Issue.record()` here? It's better to fail the test than crashing it because it would stop processing completely (not just this test)

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+                
+                To fix: Ensure LocalizationTestsConfig.json exists in bitchatTests/ directory.
+                """)
+        }
+    }()
+    
+    // MARK: - Instance Variables
+    
+    private let testsDirectoryURL = Self.testsDirectoryURL
+    private let repoRootURL = Self.repoRootURL
```
qalandarov: why do you need these?

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
qalandarov, line 1: I think you could make it a struct/final class and encapsulate all the configs in a single place instead of having some of the configs (like urls/paths) in the test files and global functions in this one

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+    /// Path to the main app's localization catalog
+    private let appCatalogPath = "bitchat/Localizable.xcstrings"
+    
+    /// Path to the share extension's localization catalog
+    private let shareExtCatalogPath = "bitchatShareExtension/Localization/Localizable.xcstrings"
+    
+    // MARK: - Configuration Validation
+    
+    /// Tests that configured keys actually exist in their respective catalogs.
+    @Test func configuredKeysExistInCatalogs() throws {
```
qalandarov, line 54: Ideally, the code should be descriptive enough that it should not require comments.
In this particular case we have another option of adding that description right in the code:

```
@Test("Tests that configured keys actually exist in their respective catalogs")
func configuredKeysExistInCatalogs() throws {
```

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let config = Self.testConfig
+        
+        // Verify all required app keys exist in app catalog
+        let allAppKeys = getAllAppKeys(from: config)
+        for key in allAppKeys {
+            #expect(appCatalog.strings.keys.contains(key), 
+                   "Configured app key '\(key)' not found in app catalog")
+        }
```
qalandarov, line 67: isn't this getting the keys from the same configs? is there any value is this test? 🤔

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+    }
+    
+    /// Tests that expected values reference valid keys and locales.
+    @Test func expectedValuesConfiguration() throws {
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let config = Self.testConfig
+        
+        for (locale, localeConfig) in config.testLocales {
+            guard localeConfig.enabled else { continue }
```
qalandarov, line 87: `for (locale, localeConfig) in config.testLocales where localeConfig.enabled {`

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+                // Key should exist in either app or share extension
+                let keyExistsInApp = appCatalog.strings.keys.contains(key)
+                let keyExistsInShare = shareExtCatalog.strings.keys.contains(key)
+                
+                #expect(keyExistsInApp || keyExistsInShare, 
+                       "Expected value key '\(key)' for locale '\(locale)' not found in any catalog")
+                
+                // Expected value should not be empty
+                #expect(!expectedValue.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty, 
+                       "Expected value for key '\(key)' in locale '\(locale)' should not be empty")
```
qalandarov, line 99: I think `#expect` already provides comments like these when tests fail

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+    func localeCompleteness(locale: String) throws {
+        let appContext = try loadContextWithKeys(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let shareExtContext = try loadContextWithKeys(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        
+        let baseLocale = appContext.baseLocale
+        let baseAppKeys = appContext.keysByLocale?[baseLocale] ?? Set()
+        let baseShareKeys = shareExtContext.keysByLocale?[baseLocale] ?? Set()
+        
+        // Skip base locale comparison with itself
+        if locale == baseLocale { return }
```
qalandarov, line 128: maybe do this check as early as possible to avoid unnecessary processing?

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+    /// Tests that expected values reference valid keys and locales.
+    @Test func expectedValuesConfiguration() throws {
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let config = Self.testConfig
+        
+        for (locale, localeConfig) in config.testLocales {
+            guard localeConfig.enabled else { continue }
+            
+            for (key, expectedValue) in localeConfig.assertValues {
```
qalandarov, line 89: maybe we could have a dedicated test that accepts an array of values instead of manually creating nested loops?

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+    
+    /// Validates that a specific enabled locale has translated state for all required keys.
+    @Test("Locale translation state", arguments: enabledLocales)
+    func localeTranslationState(locale: String) throws {
+        let appCatalog = try loadCatalog(relativePath: appCatalogPath, repoRootURL: repoRootURL)
+        let shareExtCatalog = try loadCatalog(relativePath: shareExtCatalogPath, repoRootURL: repoRootURL)
+        let config = Self.testConfig
+        
+        // Skip base locale as it's the source
+        if locale == appCatalog.sourceLanguage { return }
```
qalandarov, line 157: in the above test you compared against `appCatalog.baseLocale` here it's different - why are there different options? 🤔

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+        
+        // Check app required keys have translated state
+        let allAppKeys = getAllAppKeys(from: config)
+        for key in allAppKeys {
+            guard let entry = appCatalog.strings[key],
+                  let localization = entry.localizations[locale],
+                  let unit = localization.stringUnit else {
+                Issue.record("Required app key '\(key)' missing localization for enabled locale '\(locale)'")
+                continue
+            }
```
qalandarov, line 167: you can use `try #require` to unwrap or fail

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+            
+            #expect(unit.state == "translated",
```
qalandarov: Would this cause tests to fail if we add new untranslated strings to the app?
also, should we do case-insensitive compare to be safer? (or even better if we could have an enum for things like a "state")

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+            guard let entry = shareExtCatalog.strings[key],
+                  let localization = entry.localizations[locale],
+                  let unit = localization.stringUnit else {
+                Issue.record("Required share extension key '\(key)' missing localization for enabled locale '\(locale)'")
+                continue
+            }
+            
+            #expect(unit.state == "translated", 
+                   "Required share extension key '\(key)' not marked as translated in enabled locale '\(locale)' (state: '\(unit.state)')")
+        }
```
qalandarov, line 185: anywhere you see a for-loop, we could probably have a test that would accept the values as an argument to have cleaner and understandable tests

bitchatTests/LocalizationTests.swift
```diff
@@ -0,0 +1,403 @@
...
+                       "No content for key \(key) in enabled locale \(locale) (\(catalogName))")
+                
+                for segment in segments {
+                    let trimmed = segment.value.trimmingCharacters(in: .whitespacesAndNewlines)
+                    #expect(!trimmed.isEmpty, 
+                           "Empty translation for key \(key) at \(segment.path) in enabled locale \(locale) (\(catalogName))")
+                }
+            }
+        }
+    }
```
qalandarov, line 356: isn't this doing what you already have in tests?

bitchatTests/LocalizationTestsConfig.json
qalandarov, line 1: would we need to manually keep this in sync if a translation changes?

bitchatTests/TestUtilities/LocalizationTestTypes.swift
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/// Represents a localization for a specific locale
+struct CatalogLocalization: Decodable {
+    let stringUnit: CatalogStringUnit?
+}
+
+/// Represents a string unit with state and value
+struct CatalogStringUnit: Decodable {
+    let state: String
```
qalandarov, line 39: enum with a raw-type of String?

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+
+// MARK: - Configuration Helpers
+
+/// Extracts all app keys from the hierarchical configuration structure
+func getAllAppKeys(from config: LocalizationTestConfig) -> [String] {
+    var allKeys: [String] = []
+    for (_, keys) in config.requiredKeys.app {
+        allKeys.append(contentsOf: keys)
+    }
+    return allKeys
```
qalandarov, line 58: can you not do a simple `Array(config.requiredKeys.app.keys)`?

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/// Extracts all share extension keys from the hierarchical configuration structure
+func getAllShareExtensionKeys(from config: LocalizationTestConfig) -> [String] {
+    var allKeys: [String] = []
+    for (_, keys) in config.requiredKeys.shareExtension {
+        allKeys.append(contentsOf: keys)
+    }
+    return allKeys
+}
```
qalandarov, line 68: maybe instead of these separate and almost identical functions you could have an extension on `LocalizationTestConfig` that would centralize the code to avoid duplications?

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+func gatherSegments(from unit: CatalogStringUnit, prefix: [String] = []) -> [Segment] {
+    var segments: [Segment] = []
+    if let value = unit.value {
+        segments.append(Segment(components: prefix, value: value))
+    } else if prefix.isEmpty {
+        segments.append(Segment(components: [], value: ""))
+    }
+    if let plural = unit.variations?.plural {
+        for (variable, categories) in plural.sorted(by: { $0.key < $1.key }) {
+            for (category, variation) in categories.sorted(by: { $0.key < $1.key }) {
```
qalandarov: if we're mainly checking if keys exists in the tests, do we gain anything from these sortings?

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+    return segments
+}
+
+// MARK: - Placeholder Processing
+
+/// Safe regex creation that returns nil on failure instead of crashing
+func createPlaceholderRegex() -> NSRegularExpression? {
+    let pattern = "%(?:\\d+\\$)?#@[A-Za-z0-9_]+@|%(?:\\d+\\$)?[#0\\- +'\"]*(?:\\d+|\\*)?(?:\\.\\d+)?(?:hh|h|ll|l|z|t|L)?[a-zA-Z@]"
```
qalandarov, line 113: shouldn't the comment describe what this regex actually does with an example input/output? :)

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+            tokens.append(token)
+        }
+    }
+    return tokens
+}
+
+/// Normalizes placeholder tokens by sorting them
+func normalizedPlaceholders(_ tokens: [String]) -> [String] {
+    tokens.sorted()
+}
```
qalandarov: really? 😂
isn't calling the `tokens.sorted()` way better than this extra global function with a comment? :)

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+    let data = try Data(contentsOf: url)
+    return try JSONDecoder().decode(StringCatalog.self, from: data)
+}
+
+/// Loads the localization test configuration
+func loadConfig(testsDirectoryURL: URL) throws -> LocalizationTestConfig {
+    let url = testsDirectoryURL.appendingPathComponent("LocalizationTestsConfig.json")
+    let data = try Data(contentsOf: url)
+    return try JSONDecoder().decode(LocalizationTestConfig.self, from: data)
+}
```
qalandarov, line 147: could probably combine the logic using generics

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
```diff
@@ -0,0 +1,176 @@
...
+        var localeKeys: Set<String> = []
+        for (key, entry) in catalog.strings {
+            guard let localization = entry.localizations[locale], 
+                  localization.stringUnit != nil else {
+                continue
+            }
+            localeKeys.insert(key)
+        }
+        keysByLocale[locale] = localeKeys
+    }
```
qalandarov, line 171: introducing a computed property like `entry.isValid(for: locale)` could simplify this logic a lot, to something like this:
```
for locale in locales {
    keysByLocale[locale] = catalog.strings.keys.filter { $0.isValid(for: locale) }
}
```

qalandarov:
> @qalandarov anything left or we good to go?

sorry for the delay, having a busy week

SkinnnyJay:
Thanks @qalandarov for the quick response , sorry its been a busy week hopefully everything is good on your side.

> As I've mentioned before, I haven't seen unit testing localizations

Understood this was more inline with your above comments on the PR and recommendations and our testing suite strategy going forward.

> "forgive me if my comments sound silly..."

"No comment is silly" IMO, we are both learning and holding each other accountable for quality, 

> It would be nice to describe the "why" of the localization tests and maybe ensure that the tests won't start failing if/when a translation changes, or if a string is added/removed.

It sounds like you are suggesting this test to warn not fail around localization concerns it finds during a build? 

1. I can add a const `FAIL_ON_ERROR=false` in the test you can override to not fail the test if a string isn't present.
2. I would argue thats part of the purpose of the test. 
3. These tests are used to ensure when we are including localization of strings when shipping PRs. 
4. Address your comments in the open feedback above.


**Example where the issue could have caught something:**

We are seeing some gaps come in because shared localized strings were skipped during https://github.com/permissionlesstech/bitchat/commit/b873b1910452d8c624c2a3e80a861a0383612321 this would have been caught by this test and told the dev who did 95% of the work could have been notified via the failed test to submit the remaining 5%. 

I will add the optional flag for warning vs. failure. otherwise I think the test are ready to be merged.

👎  if you disagree.

Again @qalandarov I appreciate your feedback and collaboration around these things. There is not much direction and I want to ensure I am focusing time where its needed. You have always replied🙏 , I believe it those interactions in PR in a open-source project that  builds community around the codebase and a common vision to quality and direction jack has.

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
SkinnnyJay: I mean you said extract the functions bruh... lol you are right of course.

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
SkinnnyJay: Sure I thought line 66 did that but I can insert one and provide an example of good vs. bad.

bitchatTests/LocalizationTests.swift
SkinnnyJay: My understanding is, if in the future we wanted to catalog as we build out the suite and run parts of the suite. its similar to tagging. I will remove it for now.

bitchatTests/TestUtilities/LocalizationTestUtilities.swift
SkinnnyJay: Using to make failures deterministic and readable. Dictionaries iterate in undefined order; sorting locales ensures the first reported mismatch is consistent across runs, making CI logs stable.

qalandarov: > you said extract the functions

I guess there was a slight misunderstanding, I meant to extract your own "functions" from the localization tests and not create a new function to replace one-liner "function call" with another :)

qalandarov:
> It sounds like you are suggesting this test to warn not fail around localization concerns it finds during a build?

Not really, I wanted to understand a general idea of what you're trying to achieve with testing translations?
I'm worried it could become too noisy and fail on unrelated PRs if the translations are updated, especially if the translations are updated by other tools/automations in the future (to separate "content" from "code").

If my understanding of the intention is correct: you want to "catch" PRs that introduce a new/untranslated/bad strings - if that's right, then it sounds like a tool/script and not a unit test. Unit tests should validate implementations against expectations and they should help detect if the new code broke existing functionality/expectations

SkinnnyJay:
> > It sounds like you are suggesting this test to warn not fail around localization concerns it finds during a build?
> 
> 
> 
> Not really, I wanted to understand a general idea of what you're trying to achieve with testing translations?
> 
> I'm worried it could become too noisy and fail on unrelated PRs if the translations are updated, especially if the translations are updated by other tools/automations in the future (to separate "content" from "code").
> 
> 
> 
> If my understanding of the intention is correct: you want to "catch" PRs that introduce a new/untranslated/bad strings - if that's right, then it sounds like a tool/script and not a unit test. Unit tests should validate implementations against expectations and they should help detect if the new code broke existing functionality/expectations

The intent was to ensure PR are not causing localization issues. Are the keys defined for each locale. I don't see it causing an issue in future automation. Happy to close it.

qalandarov:
> The intent was to ensure PR are not causing localization issues. Are the keys defined for each locale. I don't see it causing an issue in future automation. Happy to close it.

I didn't mean for you to close the PR or be discouraged, I think the underlying feature of being able to detect bad/new Strings is very useful. I was just brainstorming about "how" we should go about it - unit tests vs script/tool/github-action. My only concern with the unit testing was that if it were to start failing for translations that it would become a "noise".

SkinnnyJay:
Hey @qalandarov  - sorry was dealing with other fires last week. Apologies for the "word vomit" below, hopefully it helps to clarify my approach and thinking to the test. I enjoy the feedback, I like that you are asking questions to keep the codebase clean and evolving.

> I didn't mean for you to close the PR or be discouraged, I think the underlying feature of being able to detect bad/new Strings is very useful.

no worries, don't want to clog the queue.

> unit tests vs script/tool/github-action

- I think this is a good question that was brought up in the original PR and I attempted to provide both flavors. 
- The reason behind adding this as a unit test was, to ensure we don't break our localization or cause locale drift which we then don't find out about for a while...
- We don't want an aggressive localization unit test barking while you are in dev and failing.
- If we do want a check to ensure localization is complete across all locales and translated 100% I'd say that should be a script (see below) that could be leveraged in the `github-action` or a pre-commit check. 
- I think it feels late in the SDLC to give a warning on a pr commit around localization.

> My only concern with the unit testing was that if it were to start failing for translations that it would become a "noise".

I don't think the test would cause needless noise or failures, It will let you know things that you forgot to add, i.e. a key in the localization files, which I think is a valid message, we aren't asking for you to translate them right now, I can help automate that in a future `github-action`. If we want the localization test results to be treated as a `WARNING` vs. `ERROR` that be done. I believe I have a flag that you can override the tests to hard fail vs. just doing a console output of the keys missing.

If we take this idea of new strings being added and never translated or added to the other supported locales it would be difficult for us to know otherwise that we have compromised the UI experience i.e. one string in english the other in localized string. This is fine for a short time but we should minimize that by having tests that let you know when you are committing something that puts us out of locale sync.

**What the test isn't checking:**
- Checking for every key to be translated.

**We aren't checking on each key/value is translated we are checking that the system works as expected.**
- Do localization respect plural, placeholders, etc.
- Do localization have matching keys across the locales (This should help prevents "drift" locale drift)
- Test a few common key strings with variations (simple, complex, plural, placeholder) for each locale we support to ensure those locales are being rendered correctly. I saw this issue when translating some languages where the placeholder caused translation issues.

**The test should only be checking:**
> Configuration: Key existence, expected values validation
> Structure: Locale parity, translation completeness
> Content: Non-empty translations, expected value matching
> Format: Placeholder consistency across locales

**Some other Script Ideas:**
With the base `Localization` tests, We could add additional dev commands to the JUST file, that can be leveraged to do things like:
1. Add a new key/value to all locales. - useful when you want to quickly add a new key/value to the 'en' and have it add to the rest of the locales to be translated.
2. Validate key across all locales - useful when you want to quickly ensure parity between one locale or many.
3. Add missing key/values in missing locales. - quick script to help quickly resolve missing key/values in the locale file.
4. are all key/values translated for a given locale - Helps when drift happens and can produce a CSV for quick resolution.
5. import/export KEY/VALUES for a given locale - this is the tool I use with LLM that has good reproducible workflow.

**In short I wanted to close out the Localization foundation work so we have some guardrails and then follow up with a Localization file update PR that fixes the issues, these seemed to be introduced by https://github.com/permissionlesstech/bitchat/commit/b873b1910452d8c624c2a3e80a861a0383612321**. After that I can help out where needed or get back to helping on some bugs, refactoring.

Curious of your thoughts, I can resubmit if needed before EOW and have an agent put together the locale update PR to review (missing/strings) - so in theory we could close these out maybe by Monday.

I appreciate you!

qalandarov:
Adding a pre-commit / pre-push git hook or even a github action would probably be better because the issue would be isolated to a single contributor/PR without affecting others.

Script-based scenario:
If person A creates a PR with a new string, only that PR would have a "failed" CI, and if that PR gets merged with a follow-up task, then we have only 1 affected PR = very isolated.

Unit-test scenario:
if person A creates a PR with a new string and if that gets merged even with a follow-up task, then `main` would be "broken" and all future PRs would have a "failed" CI because of unrelated changes = noise.

PR #602, opened 2025-09-13
Modularization: Extract Tor into a separate module

qalandarov:

PR #768, opened 2025-10-06
Add Tor’s xcframework and select “Do not embed”

qalandarov:
Tor's xcframework used to be treated in the .xcodeproj file as a static library with "Do Not Embed" before #602 - and it was removed in #602. This caused testflight uploads to show a warning, but still successfully upload.
Adding it again with the same "Do Not Embed" removes the testflight warning and works as before.

| Before | After |
| - | - |
| Screenshot 2025-10-07 at 12 44 18 AM | Screenshot 2025-10-07 at 12 43 26 AM |


qalandarov:
TestFlight build:

Screenshot 2025-10-07 at 12 53 27 AM

jackjackbits (approved):

PR #767, opened 2025-10-06
SwiftTesting: Enable in GitHubActions + peer uuids

qalandarov:
Using UUIDs for peer ID's would help us have unique ids across any test of the same or other test files

PR #769, opened 2025-10-07
PeerID 18/n: `BitchatDelegate` + Tests

qalandarov:


qalandarov:
When #767 is merged, the base branch of this PR should automatically update to `main` and the commit should still show `1`

qalandarov:
> does this add much string↔PeerID conversion overhead?

Since structs are almost free, there shouldn't be too much performance impact and once everything is converted to a static type the there won't be any conversions - just passing around.

PR #781, opened 2025-10-08
Convert the remaining tests to Swift Testing

### LINKED ISSUE #623: Migrate XCTests to Swift Testing
The latest [Swift Testing](https://developer.apple.com/documentation/testing/migratingfromxctest) offers a lot of benefits and the migration is not difficult

qalandarov:
There're a couple of `measure`-ment tests, but Swift Testing doesn't support performance testing + those tests don't really have a "baseline" plists to compare against.

---
Closes #623


bitchatTests/Integration/IntegrationTests.swift
```diff
@@ -62,370 +41,336 @@ final class IntegrationTests: XCTestCase {
...
         let messagesPerNode = 25
-        let expectedTotal = messagesPerNode * nodes.count * (nodes.count - 1)
-        var receivedTotal = 0
-        let expectation = XCTestExpectation(description: "High load handled")
-        
-        // Each node tracks messages
-        for (_, node) in nodes {
-            node.messageDeliveryHandler = { _ in
-                receivedTotal += 1
-                if receivedTotal >= (expectedTotal - 2) {
```
qalandarov, line 407: This was making the tests flaky and checking against the exact `expectedTotal` was always passing.
Can you double check this @jackjackbits ?

bitchatTests/Noise/NoiseProtocolTests.swift
```diff
@@ -468,89 +479,43 @@ final class NoiseProtocolTests: XCTestCase {
...
         let testResynced = "Resynced".data(using: .utf8)!
-        let encryptedResync = try aliceManager.encrypt(testResynced, for: TestConstants.testPeerID2)
-        let decryptedResync = try bobManager.decrypt(encryptedResync, from: TestConstants.testPeerID1)
-        XCTAssertEqual(decryptedResync, testResynced)
-    }
-    
-    // MARK: - Performance Tests
-    
-    func testHandshakePerformance() throws {
-        measure {
```
qalandarov, line 507: also, removed these measurements because we were not testing against any benchmark anyways (and they're not available on Swift Testing, unless we do some custom logic) - how should we go about them?

PR #784, opened 2025-10-10
Allow private message content to exceed 255 bytes

hamzaozturk:
## Summary
- Fix TLV encoding for private message content so payloads larger than 255 bytes are accepted.
- Keep messageId constrained to the original single-byte length while expanding only the content field to a two-byte length when needed.

### DIFF
--- a/bitchat/Protocols/Packets.swift
+++ b/bitchat/Protocols/Packets.swift
@@ -114,20 +114,19 @@ struct PrivateMessagePacket {
     }
 
     func encode() -> Data? {
-        var data = Data()
-        data.reserveCapacity(2 + min(messageID.count, 255) + 2 + min(content.count, 255))
-
-        // TLV for messageID
         guard let messageIDData = messageID.data(using: .utf8), messageIDData.count <= 255 else { return nil }
+        guard let contentData = content.data(using: .utf8), contentData.count <= 65535 else { return nil }
+
+        var data = Data()
+        let estimatedMessageLength = 1 + 1 + messageIDData.count
+        let estimatedContentLength = contentData.count > 255 ? 1 + 3 + contentData.count : 1 + 2 + contentData.count
+        data.reserveCapacity(estimatedMessageLength + estimatedContentLength)
+
         data.append(TLVType.messageID.rawValue)
         data.append(UInt8(messageIDData.count))
         data.append(messageIDData)
 
-        // TLV for content
-        guard let contentData = content.data(using: .utf8), contentData.count <= 255 else { return nil }
-        data.append(TLVType.content.rawValue)
-        data.append(UInt8(contentData.count))
-        data.append(contentData)
+        appendContentTLV(value: contentData, into: &data)
 
         return data
     }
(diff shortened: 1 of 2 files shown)


jackjackbits:
this is a protocol breaking change that we need to sync with android. it also has an error when message is exactly 255 chars. and no tests.

hamzaozturk:
Is there any particular reason why private messages are limited to 255 characters? Just wondering if that’s intentional.
Also, I’ve fixed the 255-char issue you mentioned and added the relevant tests for it.

jackjackbits:
arbitrary. i’ll sync with android team.

jackjackbits (changes requested):
Still relevant — the 255-byte guard is unchanged on main — and still blocked on the Android coordination we discussed in October, which hasn't happened yet. Beyond that, there's a wire-compat defect the earlier fix didn't cover:

1. **Exactly-255 cross-version misparse.** The old encoder legally emits a single-byte length of exactly 255 (`count <= 255`). The new decoder treats a first byte of 0xFF as the extended-length marker, so a valid 255-byte DM from any existing iOS/Android client is corrupted or dropped by a patched client — and extended-length payloads misparse on old clients in the other direction. The October fix made new-encoder↔new-decoder round-trip at 255, but interop with deployed clients at exactly 255 is still broken. In-place reinterpretation of length byte 0xFF can't be made safe; this needs a versioned/negotiated format or a new TLV type — which is also what the Android sync would have to specify.

2. **Non-canonical encodings accepted:** the decoder accepts 0xFF + u16 < 255, creating two valid wire forms for the same payload. A protocol decoder should reject non-canonical lengths.

3. Nit: the `encode()` reserve-capacity estimate for the short form is off (type + 1 length byte = 2) and the line has broken indentation — harmless, but worth cleaning while you're in there.

Sorry this has sat so long; the honest status is it stays held until the cross-platform format is agreed, and the exactly-255 issue needs solving in whatever format that ends up being.

vincenzopalazzo (commented):
**Production Readiness: NO-GO** — one wire-compat edge in the decoder and a test gap worth closing before merge.

Thanks for tackling this — the 0xFF escape design is the right shape, and the encode/decode split with `appendContentTLV`/`readContentLength` reads cleanly. Two things:

### Issues & Suggestions

- **Medium:** `bitchat/Protocols/Packets.swift` (`readContentLength`) — the new decoder treats a first length byte of `0xFF` as an escape for a two-byte length, but the *current* encoder accepts `contentData.count <= 255` and writes exactly-255-byte content as a bare `0xFF` with no extension bytes. A packet from today's release with 255-byte content will be misparsed by this decoder (the first two content bytes become a bogus length, decode returns nil, message dropped). Please either handle the ambiguous case (e.g. sanity-check the extended length against remaining bytes and fall back to literal 255) or call out the compat break explicitly so Android can match the decision.
- **Medium:** `bitchatTests/Protocol/PrivateMessagePacketTests.swift` — all tests are round-trip (encode → decode → compare). That cannot detect a symmetric bug where encoder and decoder are wrong in the same way (e.g. both little-endian, both off-by-one). For a wire-format change, please add golden vectors: hardcoded known-good byte sequences for (a) 254-byte content, (b) 255-byte content, (c) 300-byte content, ideally shared with Android.

Note for context, not a defect in this PR: old clients decoding the new ≥255-byte format will misparse (they read `0xFF` as a literal length and swallow the two length bytes as content). That's inherent to expanding this TLV without version negotiation, but worth a line in the PR description since wire compat is the point of the change.

### Summary

Right approach; needs the 0xFF ambiguity resolved and at least one golden vector before merge.

---

PR #792, opened 2025-10-11
Fix BLE queue usage and tighten actor isolation

jackjackbits:
## Summary
- route every CoreBluetooth operation through , assert queue affinity in debug builds, and keep  synchronous so LEAVE and cleanup finish before termination
- harden concurrency: mark / as , hop delegate callbacks back to the main actor, and rewrite  throttling to avoid non-Sendable captures; lock ’s cached 
- fix the crashy peer lookup by snapshotting  on its queue; trim the nickname setter/notification handlers to avoid duplicate announces and off-main access

## Testing
- Not run (sandbox blocked xcodebuild run); please build locally

### DIFF
--- a/bitchat/Services/BLEService.swift
+++ b/bitchat/Services/BLEService.swift
@@ -406,23 +406,35 @@ final class BLEService: NSObject {
     // MARK: Lifecycle
     
     func startServices() {
-        // Start BLE services if not already running
-        if centralManager?.state == .poweredOn {
-            centralManager?.scanForPeripherals(
-                withServices: [BLEService.serviceUUID],
-                options: [CBCentralManagerScanOptionAllowDuplicatesKey: false]
-            )
+        performOnBLEQueue { service in
+            // Start BLE services if not already running
+            if service.centralManager?.state == .poweredOn {
+                service.centralManager?.scanForPeripherals(
+                    withServices: [BLEService.serviceUUID],
+                    options: [CBCentralManagerScanOptionAllowDuplicatesKey: false]
+                )
+            }
         }
-        
-        // Send initial announce after services are ready
-        // Use longer delay to avoid conflicts with other announces
+
+        // Send initial announce after services are ready.
         messageQueue.asyncAfter(deadline: .now() + TransportConfig.bleInitialAnnounceDelaySeconds) { [weak self] in
             self?.sendAnnounce(forceSend: true)
         }
     }
     
     func stopServices() {
-        // Send leave message synchronously to ensure delivery
+        if DispatchQueue.getSpecific(key: bleQueueKey) != nil {
+            stopServicesOnBLEQueue()
+        } else {
+            bleQueue.sync {
+                self.stopServicesOnBLEQueue()
+            }
+        }
+    }
+
+    private func stopServicesOnBLEQueue() {
+        assertOnBLEQueue()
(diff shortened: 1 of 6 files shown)


bitchat/Services/BLEService.swift
```diff
@@ -406,23 +406,35 @@ final class BLEService: NSObject {
...
     
     func stopServices() {
-        // Send leave message synchronously to ensure delivery
+        if DispatchQueue.getSpecific(key: bleQueueKey) != nil {
+            stopServicesOnBLEQueue()
+        } else {
+            bleQueue.sync {
+                self.stopServicesOnBLEQueue()
+            }
+        }
```
qalandarov, line 432: I just did a similar refactor in #791 where I moved this logic of queue checking inside the actual function (in this case `stopServicesOnBLEQueue`) which should internally check the queue and call on the correct one so the caller would be "thread safe"

bitchat/Services/BLEService.swift
```diff
@@ -432,43 +444,45 @@ final class BLEService: NSObject {
...
             }
         }
-        
-        // Give leave message a moment to send
-        Thread.sleep(forTimeInterval: TransportConfig.bleThreadSleepWriteShortDelaySeconds)
-        
-        // Clear pending notifications
+
+        let delay = TransportConfig.bleThreadSleepWriteShortDelaySeconds
+        if delay > 0 {
```
qalandarov, line 460: Given that it's a constant, wouldn't we already know the answer - that it is always above 0?

bitchat/Views/ContentView.swift
```diff
@@ -19,7 +19,12 @@ import UIKit
 
 // MARK: - Main Content View
 
+@MainActor
```
qalandarov, line 22: Isn't it already the case? 🤔

bitchat/Views/ContentView.swift
```diff
@@ -79,6 +85,30 @@ struct ContentView: View {
...
+            scrollThrottleTask = nil
+            return
+        }
+        let nanoseconds = UInt64(max(delay, 0) * 1_000_000_000)
+        scrollThrottleTask = Task { [target] in
+            try? await Task.sleep(nanoseconds: nanoseconds)
+            await MainActor.run {
+                lastScrollTime = Date()
+                enqueueScroll(target: target)
+                scrollThrottleTask = nil
```
qalandarov, line 108: can we not add the `lastScrollTime` and `scrollThrottleTask` logic inside the `enqueueScroll` for consistency?

bitchat/Services/BLEService.swift
```diff
@@ -1517,13 +1517,29 @@ extension BLEService {
...
 extension BLEService {
     
-    /// Notify UI on the MainActor to satisfy Swift concurrency isolation
-    private func notifyUI(_ block: @escaping () -> Void) {
-        // Always hop onto the MainActor so calls to @MainActor delegates are safe
-        Task { @MainActor in
-            block()
+    /// Hop to the MainActor and hand `self` to the caller for UI delegate updates.
+    private func notifyUI(_ block: @MainActor @escaping (BLEService) -> Void) {
+        Task { @MainActor [weak self] in
```
qalandarov, line 1522: if the block has already @MainActor, do we need to wrap it in this `Task { @MainActor`? 🤔

PR #796, opened 2025-10-12
Refactor Nostr ID Bridge & Keychain Helper

qalandarov:
- Extract Nostr types into separate files
- Refactor Nostr ID Bridge from all-static functions and properties to be instance based
- Keychain Helper to be behind a protocol so we can easily mock so tests don't require keychain access


qalandarov:
Created #797 to eventually combine keychain services.

jackjackbits (approved):

PR #798, opened 2025-10-12
Align gossip sync stale cleanup with Android client

jackjackbits:
## Summary
- add stale-peer cleanup cadence to `GossipSyncManager` so announcements/messages beyond `STALE_PEER_TIMEOUT` are pruned with their packets
- reject incoming announce packets that are already past the stale timeout to avoid resurrecting offline peers
- add debug hooks and unit coverage verifying stale cleanup and broadcast pruning work end-to-end

## Testing
- xcodebuild -project bitchat.xcodeproj -scheme \"bitchat (iOS)\" -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.5' -only-testing:bitchatTests_iOS/GossipSyncManagerTests test

### DIFF
--- a/bitchat.xcodeproj/project.pbxproj
+++ b/bitchat.xcodeproj/project.pbxproj
@@ -312,6 +312,7 @@
 				ne,
 				"pt-BR",
 				ru,
+				tr,
 				uk,
 				"zh-Hans",
 			);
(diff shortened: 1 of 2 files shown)


callebtc:
looks like the cleanup of the _messages_ from stale peers is missing (only cleans up old announcements).

PR #803, opened 2025-10-13
Fix people sheet dismiss gestures

jackjackbits:
## Summary
- dismiss the people sheet when tapping the close button so showSidebar state cannot hold it open
- add a right-swipe gesture on the DM sheet that animates back to the people roster and clears DM state

## Testing
- xcodebuild -project bitchat.xcodeproj -scheme "bitchat (iOS)" -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.5' build

### DIFF
--- a/bitchat/Services/BLEService.swift
+++ b/bitchat/Services/BLEService.swift
@@ -1513,6 +1513,22 @@ extension BLEService {
             block()
         }
     }
+
+    /// Safely fetch the current direct-link state for a peer using the BLE queue.
+    private func linkState(for peerID: PeerID) -> (hasPeripheral: Bool, hasCentral: Bool) {
+        let computeState = { () -> (Bool, Bool) in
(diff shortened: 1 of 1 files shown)


bitchat/Views/ContentView.swift
```diff
@@ -1029,6 +1031,20 @@ struct ContentView: View {
         }
         .background(backgroundColor)
         .foregroundColor(textColor)
+        .gesture(
+            DragGesture(minimumDistance: 25, coordinateSpace: .local)
+                .onEnded { value in
+                    let horizontal = value.translation.width
+                    let vertical = abs(value.translation.height)
+                    guard horizontal > 80, vertical < 60 else { return }
```
SkinnnyJay, line 1039: I'm assuming this hardcoded won't matter?

jackjackbits: shouldn't

PR #804, opened 2025-10-14
Serialize fragment assembly on BLE message queue

jackjackbits:
## Summary
- ensure fragment intake always runs on the BLE message queue, using a barrier dispatch when invoked from other queues
- keep the pre-existing fragment logic intact in a private helper so shared state (incoming fragments, metadata) is mutated safely

## Testing
- xcodebuild -project bitchat.xcodeproj -scheme "bitchat (iOS)" -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.5' build


jackjackbits:
@qalandarov found a crash i think introduced by the ble queue pr

PR #805, opened 2025-10-14
Guard BLE link state lookups on BLE queue

jackjackbits:


jackjackbits:
@qalandarov another crash

qalandarov:
> @qalandarov another crash 

Can you give me some steps to reproduce? Maybe we can also writesome unit tests.

jackjackbits:
i’ve mainly just let an iPhone and mac app run for some time.

qalandarov:
Without sending any messages or files? Do both of them crash or only one I'll give it a try

Is it expected that some of those functions that were called specifically from the queue could be called from outside of the queue as well? Was that by design or are we uncovering the hidden issues now?

jackjackbits:
just iPhone. no sending of anything. just sitting there.

qalandarov:
Have the apps running for ~10 mins (connected to debugger) - no crash yet... will keep it running...

Speaking of crashes... is there a plan to integrate some kind of crash/log analytics?

Also, would be nice to have some kind of project dashboard to prioritize and tackle tasks (something like a github project / linear.app / trello...)

jackjackbits:
do you have suggestions for crash/log analytics?

qalandarov:
besides google's products, I've seen people use Sentry, but its quite large. I've used https://github.com/aptabase/aptabase (self-hosted) - only for analytics, don't think they have crash reporting though

PR #807, opened 2025-10-14
Reuse an existing session instead of recreating

qalandarov:
Can we reuse an existing session by just resetting it (just in case) - instead of creating a new one?

### DIFF
--- a/bitchat/Noise/NoiseSessionError.swift
+++ b/bitchat/Noise/NoiseSessionError.swift
@@ -6,10 +6,9 @@
 // For more information, see <https://unlicense.org>
 //
 
-enum NoiseSessionError: Error {
+enum NoiseSessionError: Error, Equatable {
     case invalidState
     case notEstablished
     case sessionNotFound
-    case handshakeFailed(Error)
     case alreadyEstablished
 }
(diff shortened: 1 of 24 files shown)


jackjackbits:
don't believe this is right. double checked with claude 4.5: 

This PR has a critical role mismatch bug.

  The Problem:

  NoiseSession has an immutable role field:
  let role: NoiseRole  // .initiator or .responder - set once at init

  Buggy Scenario:

  1. Alice initiates handshake → Bob receives it
  2. Bob's handleIncomingHandshake() creates session with role: .responder
  3. Handshake fails/times out (session left in non-established state)
  4. Later, Bob wants to initiate to Alice
  5. Bob calls initiateHandshake(with: alice)
  6. NEW code reuses the responder session:
  if let existingSession = sessions[peerID] {
      session = existingSession  // Still has role = .responder!
      session.reset()  // Doesn't change role
  }
  7. Calls session.startHandshake() which checks:
  if role == .initiator {
      return handshakeMessage  // Should do this
  } else {
      return Data()  // ❌ Returns EMPTY for responder!
  }
  8. Result: Empty handshake message sent = protocol violation

  Why Tests Don't Catch It:

  Current tests don't simulate this sequence:
  1. Peer A receives handshake (becomes responder)
  2. Handshake fails
  3. Same peer A tries to initiate to same peer B

qalandarov:
That's a great catch!

Would be nice if it can generate some test for it to "document" this

PR #822, opened 2025-10-18
Fix verification persistence on iOS when app backgrounds

### LINKED ISSUE #785: People verification lost
Maybe it's an intended behavior, but I noticed that, when closing the app, verifications for nicknames are lost and they need to be verified again.

jackjackbits:
## Summary
Fixes #785 - People verification was being lost when the app was closed on iOS.

## Problem
Verification status was not being saved when the iOS app entered background state, causing verified contacts to lose their verification status after the app was closed. This happened because:
- iOS can terminate backgrounded apps without calling `applicationWillTerminate`
- The debounced save timer in `SecureIdentityStateManager` may not have fired before termination
- No save was explicitly triggered when entering background state

## Solution
- **Added** `saveIdentityState()` method to force-save identity data (verifications, favorites, etc.) without stopping services
- **Updated** iOS background handler (`.background` scene phase) to call `saveIdentityState()` before the app might be terminated
- **Refactored** `applicationWillTerminate()` to use the new method for consistency
- **Fixed** selector name typo on both iOS and macOS (`appWillTerminate` → `applicationWillTerminate`)

## Technical Details
The verification data is now properly persisted to encrypted keychain storage whenever the app backgrounds on iOS, ensuring verification status persists across app launches.

The fix maintains:
- ✅ Encrypted keychain storage for verification data
- ✅ Background BLE mesh operation (doesn't stop services)
- ✅ Existing macOS behavior
- ✅ All existing persistence mechanisms

## Testing
- [x] Code compiles successfully
- [x] Both iOS and macOS selector references fixed
- [x] No warnings or errors

Reviewers should test:
1. Verify a contact via QR code
2. Background/close the app immediately (iOS)
3. Reopen the app
4. Confirm verification status persists

### DIFF
--- a/bitchat/BitchatApp.swift
+++ b/bitchat/BitchatApp.swift
@@ -84,9 +84,6 @@ struct BitchatApp: App {
                         }
                         // Proactively disconnect Nostr to avoid spurious socket errors while Tor is down
                         NostrRelayManager.shared.disconnect()
-                        // Force save identity state (verifications, favorites, etc) before backgrounding
-                        // iOS may terminate the app from background without calling applicationWillTerminate
-                        chatViewModel.saveIdentityState()
                         didEnterBackground = true
                     case .active:
                         // Restart services when becoming active
@@ -191,6 +188,10 @@ final class AppDelegate: NSObject, UIApplicationDelegate {
     
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
         return true
+    }
+    
+    func applicationWillTerminate(_ application: UIApplication) {
+        chatViewModel?.applicationWillTerminate()
     }
 }
 #endif
--- a/bitchat/ViewModels/ChatViewModel.swift
+++ b/bitchat/ViewModels/ChatViewModel.swift
@@ -4461,6 +4461,7 @@ final class ChatViewModel: ObservableObject, BitchatDelegate {
         
         // Update secure storage with verified status
         identityManager.setVerified(fingerprint: fingerprint, verified: true)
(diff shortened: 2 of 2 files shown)


bitchat/ViewModels/ChatViewModel.swift
```diff
@@ -2977,24 +2977,21 @@ final class ChatViewModel: ObservableObject, BitchatDelegate {
...
     }
     
-    @objc func applicationWillTerminate() {
-        // Send leave message to all peers
-        meshService.stopServices()
-        
+    /// Save identity state without stopping services (for backgrounding)
+    func saveIdentityState() {
         // Force save any pending identity changes (verifications, favorites, etc)
         identityManager.forceSave()
```
qalandarov, line 3257: Why not call this when the identity is actually verified? Because it's a user-initiated process of scanning the QR adding a small delay until all the verification-related process is completed would be an acceptable delay

bitchat/BitchatApp.swift
```diff
@@ -84,6 +84,9 @@ struct BitchatApp: App {
                         }
                         // Proactively disconnect Nostr to avoid spurious socket errors while Tor is down
                         NostrRelayManager.shared.disconnect()
+                        // Force save identity state (verifications, favorites, etc) before backgrounding
+                        // iOS may terminate the app from background without calling applicationWillTerminate
+                        chatViewModel.saveIdentityState()
```
qalandarov: instead of waiting for the scenePhase change we could listen to the "will terminate" notification directly - https://stackoverflow.com/a/67917232

Or, given that we already have an AppDelegate, we could utilize that as well for any before-termination logic (instead of depending on the notification system)

App could be terminated without going to background

PR #824, opened 2025-10-18
Add binary transfer protocol primitives

qalandarov:


qalandarov:
@jackjackbits  do you think this is covering the basics of data-transfer protocol?

PR #828, opened 2025-10-19
Add localizable strings in show commands view

Viniciuss-Moreira:
- Localizable strings are implemented in show commands (when type "/")
- Refactor ContentView for duplicate strings declarations
- Enum with localizable strings added in ChatViewModel

### DIFF
--- a/bitchat/Views/Components/CommandSuggestionsView.swift
+++ b/bitchat/Views/Components/CommandSuggestionsView.swift
@@ -72,11 +72,19 @@ struct CommandSuggestionsView: View {
 
 @available(iOS 17, macOS 14, *)
 #Preview {
-    @Previewable @State var messageText: String = "/m"
+    @Previewable @State var messageText: String = "/"
+    let keychain = KeychainManager()
+    let viewModel = ChatViewModel(
+        keychain: keychain,
+        idBridge: NostrIdentityBridge(),
+        identityManager: SecureIdentityStateManager(keychain)
+    )
+    
     CommandSuggestionsView(
         messageText: $messageText,
         textColor: .green,
         backgroundColor: .primary,
         secondaryTextColor: .secondary
     )
+    .environmentObject(viewModel)
 }


bitchat/ViewModels/ChatViewModel.swift
```diff
@@ -89,6 +89,34 @@ import UIKit
 #endif
 import UniformTypeIdentifiers
 
+// MARK: - Enum Commands Info
+
+struct CommandInfo {
```
qalandarov: What do you think about an enum?
Also, might as well extract into a separate file to "minify" the chatVM

bitchat/Views/ContentView.swift
```diff
@@ -772,6 +678,38 @@ struct ContentView: View {
         .background(backgroundColor.opacity(0.95))
     }
 
+    private func commandRow(for info: CommandInfo) -> some View {
+        Button(action: {
+            // Escolher o primeiro comando como default ao clicar
```
qalandarov: :)

bitchat/Views/ContentView.swift
```diff
@@ -772,6 +678,38 @@ struct ContentView: View {
...
+            messageText = info.commands.first! + " "
```
qalandarov: let's avoid force unwrapping

bitchat/Views/ContentView.swift
```diff
@@ -772,6 +678,38 @@ struct ContentView: View {
...
+            commandSuggestions = []
+        }) {
+            HStack {
+                Text(info.commands.joined(separator: ", "))
+                    .font(.bitchatSystem(size: 11, design: .monospaced))
+                    .foregroundColor(textColor)
+                    .fontWeight(.medium)
+                
+                if let syntaxKey = info.syntax {
+                    Text( NSLocalizedString(syntaxKey, comment: "") )
```
qalandarov: I think it's better to do the localization right where the string is declared to avoid issues in case if someone forgets this

bitchat/Views/ContentView.swift
```diff
@@ -772,6 +678,38 @@ struct ContentView: View {
...
+                        .font(.bitchatSystem(size: 10, design: .monospaced))
+                        .foregroundColor(secondaryTextColor.opacity(0.8))
+                }
+                Spacer()
+                
+                Text(NSLocalizedString(info.descriptionKey, comment: ""))
```
qalandarov: same as above

bitchat/ViewModels/ChatViewModel.swift
Viniciuss-Moreira: sure! a enum is a better choice indeed

fixed

bitchat/Views/ContentView.swift
Viniciuss-Moreira: fixed

qalandarov:
I think you need to rebase and force-push to clean up the history

bitchat/Views/ContentView.swift
Viniciuss-Moreira: fixed

bitchat/Views/ContentView.swift
Viniciuss-Moreira: fixed

bitchat/Views/ContentView.swift
Viniciuss-Moreira: fixed

Viniciuss-Moreira:
Rebased, thx @qalandarov for the review!

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+
+import Foundation
+
+// MARK: - CommandInfo Enum
+
+enum CommandInfo: String, CaseIterable, Identifiable {
+    case block = "/block"
+    case clear = "/clear"
+    case hug = "/hug"
+    case message = "/m, /msg"
```
qalandarov: I think this is wrong

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+    case slap = "/slap"
+    case unblock = "/unblock"
+    case who = "/w"
+    case favorite = "/fav"
+    case unfavorite = "/unfav"
```
qalandarov: I'd probably name the cases as the commands without having to declare the actual string values (as to avoid future syncing issues) - and then just add a `/` prefix when comparing

Also, missing `join` commands?

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+    case favorite = "/fav"
+    case unfavorite = "/unfav"
+
+    var id: String { rawValue }
+
+    var localizedSyntax: String? {
+        switch self {
+        case .block, .hug, .message, .slap, .unblock, .favorite, .unfavorite:
+            return NSLocalizedString("content.input.nickname_placeholder", comment: "")
+        default:
```
qalandarov: Instead of the `default` it's better to list the remaining cases so compiler will help us in case if we add more cases in the future

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+    var localizedSyntax: String? {
+        switch self {
+        case .block, .hug, .message, .slap, .unblock, .favorite, .unfavorite:
+            return NSLocalizedString("content.input.nickname_placeholder", comment: "")
+        default:
+            return nil
+        }
+    }
+
+    var localizedDescription: String {
```
qalandarov: I think these variable names could be confusing: what's the difference between `localizedSyntax` and `localizedDescription`

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+        case .message: return NSLocalizedString("content.commands.message", comment: "about message command.")
+        case .slap: return NSLocalizedString("content.commands.slap", comment: "about slap command.")
+        case .unblock: return NSLocalizedString("content.commands.unblock", comment: "abbout unblock command.")
+        case .who: return NSLocalizedString("content.commands.who", comment: "about who command.")
+        case .favorite: return NSLocalizedString("content.commands.favorite", comment: "about favorite command.")
+        case .unfavorite: return NSLocalizedString("content.commands.unfavorite", comment: "about unfavorite command.")
+        }
+    }
+
+    static func all(isGeoPublic: Bool, isGeoDM: Bool) -> [CommandInfo] {
```
qalandarov, line 51: if we have a custom `all` do we really need `CaseIterable`?

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+    case who = "/w"
+    case favorite = "/fav"
+    case unfavorite = "/unfav"
+
+    var id: String { rawValue }
+
+    var localizedSyntax: String? {
+        switch self {
+        case .block, .hug, .message, .slap, .unblock, .favorite, .unfavorite:
+            return NSLocalizedString("content.input.nickname_placeholder", comment: "")
```
qalandarov: looks like "syntax" in the old code sometimes is "<nickname>" and sometimes "[nickname]" - also, not sure if it's localized

bitchat/Views/ContentView.swift
```diff
@@ -606,67 +607,10 @@ struct ContentView: View {
...
-                                        .foregroundColor(secondaryTextColor)
-                                }
-                                .padding(.horizontal, 12)
-                                .padding(.vertical, 3)
-                                .frame(maxWidth: .infinity, alignment: .leading)
-                            }
-                            .buttonStyle(.plain)
-                            .background(Color.gray.opacity(0.1))
-                        }
+                    ForEach(filteredCommands, id: \.id) { info in
```
qalandarov: since the commands are identifiable - you could remove the `, id: \.id`

bitchat/Views/ContentView.swift
```diff
@@ -677,6 +621,7 @@ struct ContentView: View {
                 .padding(.horizontal, 12)
             }
 
+
```
qalandarov, line 612: redundant?

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,60 @@
...
+        switch self {
+        case .block: return NSLocalizedString("content.commands.block", comment: "about block command.")
+        case .clear: return NSLocalizedString("content.commands.clear", comment: "about clear command.")
+        case .hug: return NSLocalizedString("content.commands.hug", comment: "about hug command.")
+        case .message: return NSLocalizedString("content.commands.message", comment: "about message command.")
+        case .slap: return NSLocalizedString("content.commands.slap", comment: "about slap command.")
+        case .unblock: return NSLocalizedString("content.commands.unblock", comment: "abbout unblock command.")
+        case .who: return NSLocalizedString("content.commands.who", comment: "about who command.")
+        case .favorite: return NSLocalizedString("content.commands.favorite", comment: "about favorite command.")
+        case .unfavorite: return NSLocalizedString("content.commands.unfavorite", comment: "about unfavorite command.")
```
qalandarov: Why are we not reusing the same `String(localized...` is there a benefit in this approach vs the old code?

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: what commands ``join`` ? 
in localizable there are only those from the enum

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: make sense

fixed

bitchat/Views/ContentView.swift
Viniciuss-Moreira: really

fixed

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: fixed

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: changed

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: fixed

bitchat/Views/ContentView.swift
Viniciuss-Moreira: fixed

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: really before was better, I ended up running over myself in the middle of way 😅

fixed

Viniciuss-Moreira:
Rebased. Thx again @qalandarov for review, I started recently in iOS development, is being very enlightening!

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Foundation
+
+// MARK: - CommandInfo Enum
+
+enum CommandInfo: Identifiable {
```
qalandarov: `enum CommandInfo: String` would eliminate the need for `name` below

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+    case block, clear, hug, message, slap, unblock, who, favorite, unfavorite
```
qalandarov: I think listing these in separate lines is better for adding/removing easily

bitchat/Models/CommandInfo.swift
qalandarov, line 1: `CommandInfo`? (without an "s")

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+        case .who:
+            return ["/w", "/who"]
+        default:
+            return ["/\(name)"]
+        }
+    }
+    
+    var primaryAlias: String { aliases.first ?? "" }
+    
+    var commandsPlaceholderKey: String? {
```
qalandarov: the prefix "commands" is redundant here because the use case could be `<some command>.placeholderKey` (also, should it be private?)

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+        case .message:
+            return ["/m", "/msg"]
+        case .who:
+            return ["/w", "/who"]
+        default:
+            return ["/\(name)"]
+        }
+    }
+    
+    var primaryAlias: String { aliases.first ?? "" }
```
qalandarov: the whole point of using an enum is to enumerate all the possible cases to avoid unnecessary optionality and string conversions as much as possible :)

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+enum CommandInfo: Identifiable {
+    case block, clear, hug, message, slap, unblock, who, favorite, unfavorite
+    
+    var id: String { name }
+
+    var name: String {
+        String(describing: self)
+    }
+    
+    var aliases: [String] {
```
qalandarov: how about having an enum that takes a value so we always deal with an enum case vs dealing with string conversions? something like:

```
init?(str: String) {
    switch str {
    case "m", "msg":
        self = .message
    case "w", "who":
        self = .who
    default:
        self.init(rawValue: str)
    }
}
```

(typing without an IDE, sorry for the typos)

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+    
+    var primaryAlias: String { aliases.first ?? "" }
+    
+    var commandsPlaceholderKey: String? {
+        switch self {
+        case .block, .hug, .message, .slap, .unblock, .favorite, .unfavorite:
+            return "content.input.nickname_placeholder"
+        case .clear, .who:
+            return nil
```
qalandarov, line 33: can we not have placeholders for these too?

bitchat/Models/CommandsInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    var commandsDescription: String {
+        NSLocalizedString(commandsDescriptionKey, comment: "about \(name) command")
+    }
+    
+    static func all(isGeoPublic: Bool, isGeoDM: Bool) -> [CommandInfo] {
+        let mandatory: [CommandInfo] = [.block, .clear, .hug, .message, .slap, .unblock, .who]
+        let optional: [CommandInfo] = [.favorite, .unfavorite]
+        
+        return mandatory + (isGeoPublic || isGeoDM ? [] : optional)
```
qalandarov: the parenthesis should be around combined bool check: `(isGeoPublic || isGeoDM) ? ... : ...` - otherwise the `?` could be confusing

We can probably simplify it a bit like this to be more explicit:

```
if isGeoPublic || isGeoDM {
    return mandatory + [.favorite, .unfavorite]
}

return mandatory
```

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: ```clear``` and ```who``` are commands that don't need to mention a user, no?

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: sure, make sense

fixed

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: fixed

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: I tried to improve the example a bit and added

bitchat/Models/CommandsInfo.swift
Viniciuss-Moreira: sure

fixed

Viniciuss-Moreira:
Rebased again. @qalandarov do u think there is anything else to change ?

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,99 @@
...
+    
+    var primaryAlias: String {
+        aliases[0]
+    }
+    
+    private var placeholderKey: String? {
+        switch self {
+        case .clear, .who:
+            return nil
+        default:
```
qalandarov: try to never use the `default` if we can list all the other cases because new cases could be "lost" accidentally

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,99 @@
...
+    var description: String {
+        NSLocalizedString(descriptionKey, comment: "about \(rawValue) command")
+    }
+    
+    static func all(isGeoPublic: Bool, isGeoDM: Bool) -> [CommandInfo] {
+        let mandatory: [CommandInfo] = [.block, .clear, .hug, .message, .slap, .unblock, .who]
+        let optional: [CommandInfo] = [.favorite, .unfavorite]
+        let includeOptional = !(isGeoPublic || isGeoDM)
+        
+        return mandatory + (includeOptional ? optional : [])
```
qalandarov: why not just we keep a very simple and an explicit `if`? :)
include optional with "not or" makes it even more confusing than before :)

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+// MARK: - Custom Init
+
+extension CommandInfo {
+    init?(from input: String) {
+        let cleaned = input.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
+        let rawInput = cleaned.hasPrefix("/") ? String(cleaned.dropFirst()) : cleaned
```
qalandarov: if it doesn't have a `/` prefix, shouldn't we just return nil?

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,99 @@
...
+        
+        for command in CommandInfo.all(isGeoPublic: false, isGeoDM: false) {
+            let normalizedAliases = command.aliases.map { $0.lowercased().replacingOccurrences(of: "/", with: "") }
+            if normalizedAliases.contains(rawInput) {
+                self = command
+                return
+            }
+        }
```
qalandarov: We have `all`, `aliases`, `primaryAlias`... the whole purpose of an enum to simplify the usage of it without having to know the implementation details (which should also be as simple as possible)

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// MARK: - Custom Init
+
+extension CommandInfo {
+    init?(from input: String) {
+        let cleaned = input.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
+        let rawInput = cleaned.hasPrefix("/") ? String(cleaned.dropFirst()) : cleaned
+        
+        for command in CommandInfo.all(isGeoPublic: false, isGeoDM: false) {
```
qalandarov: why are we always passing false for the geo bools?

bitchat/Models/CommandInfo.swift
```diff
@@ -0,0 +1,99 @@
...
+        case .slap: return "content.commands.slap"
+        case .unblock: return "content.commands.unblock"
+        case .who: return "content.commands.who"
+        case .favorite: return "content.commands.favorite"
+        case .unfavorite: return "content.commands.unfavorite"
+        }
+    }
+    
+    var description: String {
+        NSLocalizedString(descriptionKey, comment: "about \(rawValue) command")
```
qalandarov: I'm not sure where my comment went, but I wanted to know if there was any benefit of using `NSLocalizedString` over `String(localized` version?

also, given that your description and placeholder are [almost] one-liners, why not just inline the logic with their `key`s?

bitchat/Views/ContentView.swift
```diff
@@ -606,75 +607,17 @@ struct ContentView: View {
             }
 
             // Command suggestions
-            if showCommandSuggestions && !commandSuggestions.isEmpty {
+            if showCommandSuggestions {
```
qalandarov: I think you can also update this logic: you can try to see if you can initialize a command from the input, if so, render the commands, else ignore. this way we could get ride of that extra `@State` variable + extract this whole thing out as a view builder (or modifier, forgot which one), something like:

`showCommandSuggestionsIfNeeded(for: input)`

```
@ViewBuilder (or modifier)
func showCommandSuggestionsIfNeeded(for input: String) -> some View {
    if let command = CommandInfo(str: input) {
        ...
    }
}
```

qalandarov (commented):
Sorry that I keep sending it back to you :) - also, please make sure to rebase each time you update the PR

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: oh yes, now I understand what you meant in that comment

fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: sure

fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: I was confused if the ```if``` wouldn't make understanding worse, but you're right

fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: make sense!

fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: I tried to improve this as much as possible

fixed

bitchat/Models/CommandInfo.swift
Viniciuss-Moreira: really

corrected

bitchat/Views/ContentView.swift
Viniciuss-Moreira: I added something along these lines inside the ```ContentView```, but wouldn't it be better to leave the logic only in the ```viewmodel```?

changed

Viniciuss-Moreira:
> Sorry that I keep sending it back to you :) - also, please make sure to rebase each time you update the PR

No problems :) these reviews are being very good

@qalandarov can take a look when u can

qalandarov (approved):

qalandarov:
I cleaned up a bit, changed `/m` -> `/dm` & `/w` -> `/who`, moved `/unblock` right after `/block` + showing fave-info only when it's a location channel

| Mesh Commands | Location Commands |
| - | - |
| Simulator Screenshot - iPhone 14 Pro - 2025-10-29 at 17 37 29 | Simulator Screenshot - iPhone 14 Pro - 2025-10-29 at 17 37 38 |

PR #834, opened 2025-10-19
Fix camera icon in DM sheets by consolidating sheet presentations

jackjackbits:
## Summary

Fixes the issue where tapping the camera icon in a mesh DM would show the error "Currently, only presenting a single sheet is supported" repeatedly, and the image picker wouldn't appear until after closing the DM sheet.

## Root Cause

ContentView had **9 competing `.sheet()` modifiers** across the view hierarchy:
- 6 on main body (people/chat, app info, fingerprint, image picker, mac picker, preview)
- 3 on mainHeaderView (verification, location channels, location notes)

SwiftUI only allows **one sheet presentation per view hierarchy**, so these modifiers created presentation conflicts. When tapping the camera icon inside a DM, the image picker couldn't take over because the DM sheet had higher priority.

### Part 1: Consolidated Sheet Management (91ac72fc)
- Created `SheetType` enum to represent all possible sheets with associated data
- Implemented `currentSheet` computed property with priority-based selection
- Added `sheetBinding` custom Binding to map between enum and state variables
- Consolidated all 9 `.sheet()` modifiers into a single `.sheet(item: sheetBinding)`
- Clear priority order: peopleOrChat > verification > locationChannels > locationNotes > appInfo > fingerprint > imagePreview

### Part 2: fullScreenCover for Image Picker (e58e5321)
- Removed image picker from `SheetType` enum (not a "sheet" anymore)
- Changed iOS image picker to use `.fullScreenCover()` instead of `.sheet()`
- This allows the picker to present OVER existing sheets (e.g., from within a DM)
- macOS keeps using `.sheet()` as it doesn't have the same nesting limitations

### Part 3: Prevent Duplicate fullScreenCover Conflicts (75770620)
- Initially had TWO fullScreenCovers both watching `$showImagePicker` (main body + peopleSheetView)
- Made them mutually exclusive using conditional `Binding`:
  - Main body: only active when NOT in a sheet
  - peopleSheetView: only active when IN a sheet
- Ensures only ONE fullScreenCover is active at any time

## Technical Details

**Before:**
```
9 competing .sheet() modifiers
└─ Priority conflicts
└─ Image picker couldn't override DM sheet
└─ "Currently, only presenting a single sheet is supported"
```

**After:**
```
1 consolidated .sheet() for regular sheets
└─ Clear priority system
└─ No conflicts

2 mutually-exclusive .fullScreenCover() for image picker
├─ Main body: for main view camera
└─ peopleSheetView: for DM camera
```

## Testing

Tested scenarios:
- ✅ Camera icon in mesh DM (tap = library, long press = camera)
- ✅ Camera icon in main #mesh view
- ✅ All other sheets still work (people list, app info, verification, etc.)
- ✅ No error messages
- ✅ Clean sheet transitions
- ✅ iOS and macOS compatibility

## Files Changed

- `bitchat/Views/ContentView.swift` (+322, -135)

## Related Issues

Fixes the camera icon issue in mesh DMs where image picker wouldn't appear.

### DIFF
--- a/bitchat/Views/ContentView.swift
+++ b/bitchat/Views/ContentView.swift
@@ -203,7 +203,15 @@ struct ContentView: View {
             }
         }
 #if os(iOS)
-        .sheet(isPresented: $showImagePicker) {
+        // Only present image picker from main view when NOT in a sheet
+        .fullScreenCover(isPresented: Binding(
+            get: { showImagePicker && !showSidebar && viewModel.selectedPrivateChatPeer == nil },
+            set: { newValue in
+                if !newValue {
+                    showImagePicker = false
+                }
+            }
+        )) {
             ImagePickerView(sourceType: imagePickerSourceType) { image in
                 showImagePicker = false
                 if let image = image {
@@ -223,7 +231,15 @@ struct ContentView: View {
         }
 #endif
 #if os(macOS)
-        .sheet(isPresented: $showMacImagePicker) {
+        // Only present Mac image picker from main view when NOT in a sheet
+        .sheet(isPresented: Binding(
+            get: { showMacImagePicker && !showSidebar && viewModel.selectedPrivateChatPeer == nil },
+            set: { newValue in
+                if !newValue {
+                    showMacImagePicker = false
+                }
+            }
+        )) {
             MacImagePickerView { url in
                 showMacImagePicker = false
                 if let url = url {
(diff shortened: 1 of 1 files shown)


jackjackbits:
@qalandarov is there a better way to do this?

qalandarov:
> @qalandarov is there a better way to do this?

I just noticed it when trying out `FingerprintView` in #833.

One way to handle it is to let the presented view to present another view (parent-child relationship) vs of all presentations originating from `ContentView` (siblings)

qalandarov:
I feel like AI is still over-engineering this solution... I think it's a matter of having a few internal states in different views that present other screens while being presented themselves. Then we can eventually refactor the views and have dedicated view models for better state management.

jackjackbits:
will go for this for now

PR #841, opened 2025-10-19
Simplify validation, compression heuristics, and notification scheduling

jackjackbits:
## Summary
- streamline user string validation to trimming and simple control-character checks
- replace byte-frequency map in compression heuristics with a lightweight uniqueness check
- remove unnecessary main-thread dispatch when scheduling notifications

## Testing
- swift test *(fails: Tor package requires FoundationNetworking on Linux)*

------
https://chatgpt.com/codex/tasks/task_e_68f5342837f88331aee8c10ea60aab15

### DIFF
--- a/bitchat/Models/RequestSyncPacket.swift
+++ b/bitchat/Models/RequestSyncPacket.swift
@@ -8,6 +8,14 @@ struct RequestSyncPacket {
     let p: Int
     let m: UInt32
     let data: Data
+    let types: SyncTypeFlags?
+
+    init(p: Int, m: UInt32, data: Data, types: SyncTypeFlags? = nil) {
+        self.p = p
+        self.m = m
+        self.data = data
+        self.types = types
+    }
 
     func encode() -> Data {
         var out = Data()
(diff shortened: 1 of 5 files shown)


bitchat/Services/NotificationService.swift
```diff
@@ -84,24 +78,18 @@ final class NotificationService {
...
+        content.sound = .default
+        content.interruptionLevel = .timeSensitive  // Make it more prominent
+
+        let request = UNNotificationRequest(
+            identifier: identifier,
+            content: content,
+            trigger: nil // Deliver immediately
+        )
+
+        UNUserNotificationCenter.current().add(request)
```
qalandarov: can we not somehow combine these 2 into a function?

qalandarov (approved):

bitchat/Services/NotificationService.swift
iSephix: why though?isnt it more scalable, especially in regards to future additions, if we don't..?

qalandarov: We can "tidy up" now and extend in the future when the need arises

PR #858, opened 2025-10-23
Add new relay URL to NostrRelayManager

Spl0itable:
Add the blaster relay so that outgoing messages in geohash channels for teleported users are sent to bridged bitchat clients. Perhaps, also add as a blaster for all geohash channels that leverage the GPS-related relays with some regex pattern to ensure all locations use the `wss://sendit.nosflare.com` blaster relay when `relays/online_relays_gps.csv` is used.


jackjackbits:
i believe this would fragment the network.

PR #863, opened 2025-10-23
Remove redundant conditional compilation directive

jithin-renji:
Removed the compilation directive that checked for iOS/macOS in `BitchatApp.swift` since it was redundant. Both branches were setting the same `chatViewModel` object.


jithin-renji:
Hi @qalandarov, thanks for approving the changes! It would be great if you could merge this when you get time, since it's a simple change.

PR #928, opened 2026-01-04
fix: clear DH shared secrets after Noise handshake operations

jackjackbits:
## Summary
- Fixed 6 locations in `NoiseProtocol.swift` where `secureClear()` was missing after DH key agreement operations
- Added `TrackingMockKeychain` to verify secureClear calls in tests
- Added 3 new tests to verify proper secret clearing

## Problem
4 of 8 DH operations in the Noise protocol weren't clearing shared secrets from memory after use. This breaks forward secrecy guarantees - if memory is later compromised, past session keys could be recovered.

## Solution
Added `secureClear()` calls for all DH operations:
- `writeMessage()`: .es (initiator/responder), .se (initiator/responder)
- `performDHOperation()`: .ee and .ss operations

Pattern applied:
```swift
let shared = try localEphemeral.sharedSecretFromKeyAgreement(with: remoteStatic)
var sharedData = shared.withUnsafeBytes { Data($0) }
symmetricState.mixKey(sharedData)
keychain.secureClear(&sharedData)
```

## Test plan
- [x] All existing NoiseProtocolTests pass
- [x] New `secureClearCalledDuringHandshake` test verifies secureClear is called at least 6 times
- [x] New `encryptionWorksAfterSecureClear` test verifies encryption still works
- [x] New `secureClearCalledInBothWriteAndReadPaths` test verifies both paths clear secrets

### DIFF
--- a/bitchat/Services/MessageDeduplicationService.swift
+++ b/bitchat/Services/MessageDeduplicationService.swift
@@ -12,6 +12,8 @@ import Foundation
 
 /// Generic LRU (Least Recently Used) cache for deduplication.
 /// Uses an efficient O(1) lookup with periodic compaction.
+/// Thread-safe via @MainActor - all callers are already on main actor.
+@MainActor
 final class LRUDeduplicationCache<Value> {
     private var map: [String: Value] = [:]
     private var order: [String] = []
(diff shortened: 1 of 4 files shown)


ColbySerpa:
Happy new year! :-)

PR #934, opened 2026-01-05
chore: pin swift-secp256k1 to exact version 0.21.1

csjones:
Change dependency requirement from upToNextMajorVersion to exactVersion to ensure consistent builds and prevent unexpected breaking changes. This change makes the Xcode project more similar to the [SPM manifest](https://github.com/permissionlesstech/bitchat/blob/aff700a15e1e5fcc7bb33844eb65046ca866e0ea/Package.swift#L21). 

image


callebtc (approved):
LGTM

PR #938, opened 2026-01-12
feat: Source routing v2

callebtc:
implements same source routing feature as on android: https://github.com/permissionlesstech/bitchat-android/pull/445

- announce direct neighbors in `ANNOUNCE` packet for gossip
- source-route packets with a valid route to the recipient
- forward source-routed packets if we're an intermediate hop

changes:
- only v2 packets support source routing
- v2 packets have a `route` field in the header
- `route` includes only intermediate hops, first is `senderID`, last is `recipientID`
- clients use an edge for routing only if both edges have confirmed the edge: for `A <--> B` will only be used if both `A` and `B` have announced their edge
- this also guarantees backwards compat: only updated clients with support for source routing will have announced the edge, which means they will also be able to decode a v2 packet with a route

## TODO
- [ ] some interop tests with android still needed

### DIFF
--- a/bitchat/BitchatApp.swift
+++ b/bitchat/BitchatApp.swift
@@ -63,6 +63,10 @@ struct BitchatApp: App {
 
                     // Initialize network activation policy; will start Tor/Nostr only when allowed
                     NetworkActivationService.shared.start()
+                    
+                    // Start presence service (will wait for Tor readiness)
+                    GeohashPresenceService.shared.start()
+
                     // Check for shared content
                     checkForSharedContent()
                 }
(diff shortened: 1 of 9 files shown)


jackjackbits:
Minor Observations (Not Blockers)

  1. BFS inside sync block: The new computeRoute performs BFS while holding the queue lock. With maxHops: 10, this is fine for typical mesh sizes, but could briefly block other topology operations for large networks.
  2. Pruning interval (60s): Topology nodes are pruned after 60s without updates. This is conservative but may need tuning based on real-world usage patterns.
  3. Dynamic chunk sizing calculation:
  let overhead = 16 + 8 + 8 + routeSize + 13 + 16
  calculatedChunk = max(64, bleMaxMTU - overhead)
  3. The overhead calculation looks correct (HeaderV2 + SenderID + RecipientID + Route + FragmentHeader + padding buffer), but edge cases with very long routes could push chunks below the 64-byte minimum.

PR #972, opened 2026-01-18
fix: Rate limit iOS peer notifications to prevent flood

jackjackbits:
## Summary
- Remove aggressive `formIntersection` that cleared `recentlySeenPeers` when peers temporarily dropped, causing them to be treated as "new"
- Add 5-minute cooldown between notifications (aligns with Android approach)
- Use fixed notification identifier so iOS updates existing notification instead of creating new ones

## Problem
Users reported getting notifications every 10-30 seconds when multiple peers are around. Root cause was:
1. `formIntersection` immediately removed peers from tracking when they dropped
2. When peers returned, they were treated as "new" → notification fired
3. No rate limiting was applied (property existed but was unused)
4. Each notification used a unique ID, spamming the notification center

## Test plan
- [ ] Connect to multiple BLE peers
- [ ] Walk in/out of range to trigger peer count changes
- [ ] Verify only one notification appears per 5-minute window
- [ ] Verify notification updates its count instead of creating new notifications

🤖 Generated with [Claude Code](https://claude.ai/code)


jackjackbits:
Addressed Codex feedback: moved `recentlySeenPeers.formUnion(newPeers)` inside the cooldown check so peers arriving during cooldown are included in the next notification rather than being silently dropped.

PR #1019, opened 2026-02-05
Add timestamp validation for incoming Nostr events

j0KZ:
## Summary
- Validates event timestamps to reject clearly invalid events and mitigate replay attacks
- Different event types have different tolerance windows based on their nature

## Validation Rules
| Event Type | Kind | Window | Rationale |
|------------|------|--------|-----------|
| Ephemeral events | 20000-29999 | ±1 hour | Should be recent, used for presence/chat |
| Gift wraps (NIP-17) | 1059 | ±24 hours | NIP-17 uses randomized timestamps for privacy |
| Other events | * | Future ≤5 min | Historical events OK, reject clearly invalid futures |

## Changes
- Add `hasValidTimestamp()` method to `NostrEvent` struct
- Validate timestamp in `ParsedInbound` alongside signature validation
- Document the validation rules with clear comments

## Test plan
- [ ] Verify ephemeral events outside ±1 hour are rejected
- [ ] Verify gift wraps within ±24 hours are accepted
- [ ] Verify events >5 min in the future are rejected
- [ ] Verify historical events (e.g., old text notes) are accepted
- [ ] Verify normal operation of location channels and private messages

🤖 Generated with [Claude Code](https://claude.ai/code)


jackjackbits:
I think there is still a gap in the timestamp hardening for Nostr DMs. This PR validates the outer relay event in `ParsedInbound`, but the DM path later decrypts the gift wrap and uses `rumor.created_at` as the actual message timestamp. That inner timestamp is still accepted unchecked, so a crafted DM can carry an arbitrarily old or future rumor timestamp and still pass as long as the wrapper timestamp is in range. In practice that means the replay / clock-skew protection is incomplete for private messages, and future-dated rumors can still skew unread logic and ordering. I think the timestamp check needs to be applied after `decryptPrivateMessage(...)` as well, before converting `rumorTimestamp` into `messageTimestamp`.

jackjackbits:
Closing as stale — the hardening idea (timestamp validation on incoming Nostr events) is still valid, but NostrRelayManager has been substantially refactored since this was written (subscription generations, per-relay state, dedup changes), so the patch no longer applies meaningfully and we haven't heard back in four months. A fresh PR against current main would be welcome.

PR #1020, opened 2026-02-05
Add Turkish localization support

abdbali:
Motivation
Provide Turkish language support for the app UI and help/commands so Turkish-speaking users can use the app in their language.
Ensure the new localization is included in the project build configuration and Xcode project metadata so it is packaged with the app.
Description
Added bitchat/tr.lproj/Localizable.strings containing Turkish translations for the main UI, commands and AppInfo copy (≈95 keys).
Updated bitchat/Views/ContentView.swift to use a localized fallback for the default channel name via String(localized: "channel") and fixed self-label formatting to use string interpolation ("\(displayName) (you)").
Registered the localization in the build configuration by adding bitchat/tr.lproj to project.yml resources and updated bitchat.xcodeproj/project.pbxproj to add the tr region, file references, and resource entries.


jackjackbits:
already exists

PR #1041, opened 2026-02-24
fix: use country-level resolution for low-precision geohashes

### LINKED ISSUE #887: Geohash reverse decode differs
When you select rehash channels and bookmark them the location changes

See for example the image below. If we have the region as "United Kingdom" or "gc" in the bookmarks we see "gc" is decided to "England" instead (even though I'm actually in Scotland).  

image

malkovitc:
## Summary

Fixes #887 — When bookmarking a 2-character geohash channel (e.g., "gc" for United Kingdom), the resolved location name showed "England" instead of "United Kingdom".

**Root cause:** `resolveCompositeAdminName()` and `nameForGeohashLength()` in `LocationStateManager.swift` prioritized `administrativeArea` (province/state) over `country` for low-precision geohashes (≤2 characters). A 2-char geohash covers a region/country, so the country name should take precedence.

## Changes

- `resolveCompositeAdminName()` — swapped `if/else` priority: check `pm.country` first, fall back to `pm.administrativeArea`
- `nameForGeohashLength()` — changed `case 0...2` from `pm.administrativeArea ?? pm.country` to `pm.country ?? pm.administrativeArea`

## How it works now

| Geohash | Before | After |
|---------|--------|-------|
| `gc` (UK) | "England" or "England and Scotland" | "United Kingdom" |
| `u3` (France) | Administrative area name | "France" |

Higher-precision geohashes (3+ chars) are unaffected — they continue to resolve to province/city/neighborhood as appropriate.

## Test plan

- [ ] Bookmark a 2-char geohash channel → verify country name is displayed
- [ ] Bookmark a 3+ char geohash channel → verify province/city name still works
- [ ] Remove and re-bookmark to clear cached names

### DIFF
--- a/bitchat/Services/LocationStateManager.swift
+++ b/bitchat/Services/LocationStateManager.swift
@@ -460,12 +460,12 @@ final class LocationStateManager: NSObject, CLLocationManagerDelegate, Observabl
             geocoder.reverseGeocodeLocation(loc) { [weak self] placemarks, _ in
                 guard self != nil else { return }
                 if let pm = placemarks?.first {
-                    if let admin = pm.administrativeArea, !admin.isEmpty, !seenAdmins.contains(admin) {
-                        seenAdmins.insert(admin)
-                        uniqueAdmins.append(admin)
-                    } else if let country = pm.country, !country.isEmpty, !seenAdmins.contains(country) {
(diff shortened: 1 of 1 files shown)


malkovitc:

PR #1042, opened 2026-02-24
fix: disable macOS focus ring on chat input TextField

### LINKED ISSUE #905: When clicking on the text field to message, a grey box covers it. It looks fine when not focusing on the text box (MacOS)
Image

Image

malkovitc:
## Summary

Fixes #905 — On macOS, clicking into the message input TextField caused a grey box (system focus ring) to cover the field.

**Root cause:** The TextField uses `.textFieldStyle(.plain)` with a custom `.background(RoundedRectangle...)`, but macOS still renders its default focus ring overlay on top, creating a visual glitch. There was no platform-specific handling to suppress this.

## Changes

- Added `FocusEffectDisabledModifier` — a `ViewModifier` that calls `.focusEffectDisabled()` on macOS 14.0+ and is a no-op on earlier versions and iOS
- Applied the modifier to the **chat message input** TextField
- Applied the modifier to the **nickname input** TextField (same issue)

### Why a ViewModifier instead of a direct `#if os(macOS)` block?

`.focusEffectDisabled()` requires macOS 14.0+, but the project's deployment target is macOS 13.0. The modifier uses `@available` check to gracefully degrade — on macOS 13 the system focus ring remains (no crash), on macOS 14+ it's properly suppressed.

## Test plan

- [ ] macOS 14+: Click into chat input → no grey box overlay
- [ ] macOS 14+: Click into nickname input → no grey box overlay
- [ ] macOS 13: App still builds and runs without crash (focus ring may still appear)
- [ ] iOS: No visual change, text fields work as before

### DIFF
--- a/bitchat/Views/ContentView.swift
+++ b/bitchat/Views/ContentView.swift
@@ -27,6 +27,22 @@ private struct MessageDisplayItem: Identifiable {
     let message: BitchatMessage
 }
 
+/// On macOS 14+, disables the default system focus ring on TextFields.
+/// On earlier macOS versions and on iOS this is a no-op.
+private struct FocusEffectDisabledModifier: ViewModifier {
+    func body(content: Content) -> some View {
+        #if os(macOS)
+        if #available(macOS 14.0, *) {
+            content.focusEffectDisabled()
+        } else {
+            content
+        }
+        #else
+        content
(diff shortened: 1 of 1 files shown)


malkovitc:

PR #1043, opened 2026-02-24
fix: allow verification sheet in private chat via sidebar navigation

### LINKED ISSUE #988: Clicking Person Name in Private Chat Doesn't Open Security Verification Sheet
When I click on the person name in private chat the security verification sheet doesn't open.
This is mostly because SwiftUI doesn't allow multiple sheets to be presented at the same time.

https://github.com/user-attachments/assets/bc831c10-fcc0-4cac-b744-701103d91b94

malkovitc:
## Summary

Fixes #988 — Clicking on a person's name in a private chat did not open the security verification (fingerprint) sheet.

**Root cause:** SwiftUI does not allow multiple `.sheet()` modifiers to be active simultaneously on the same view. When a private chat is open, the sidebar sheet is already presented (`ContentView.swift:177`), so the fingerprint sheet (`.sheet` at line 196) was silently ignored by SwiftUI.

### 1. Guard the root-level fingerprint sheet
Added conditions to prevent the root-level fingerprint `.sheet` from activating while the sidebar is open:
```swift
get: { viewModel.showingFingerprintFor != nil && !showSidebar && viewModel.selectedPrivateChatPeer == nil }
```
This preserves existing behavior when the sidebar is closed.

### 2. Present fingerprint inside the sidebar via NavigationStack
Wrapped `peopleSheetView` content in a `NavigationStack` and added a `.navigationDestination` that pushes `FingerprintView` when `showingFingerprintFor` is set while the sidebar is active.

### How it works now

| Context | Before | After |
|---------|--------|-------|
| Sidebar open (private chat) | Fingerprint sheet silently blocked | Fingerprint pushed inside sidebar via navigation |
| Sidebar closed | Root-level sheet opens | Root-level sheet opens (unchanged) |

The user can navigate back from the fingerprint view using the standard SwiftUI back gesture/button.

## Design decision

**Why navigation inside the sidebar instead of dismiss-then-present?**
- Dismissing the sidebar first would disrupt the chat context (jarring UX)
- A single-sheet-with-enum approach would require refactoring all 9+ sheet modifiers (too invasive)
- NavigationStack inside sheets is an established SwiftUI pattern

## Test plan

- [ ] Open a private chat → tap sender name → tap verify → fingerprint view opens inside sidebar
- [ ] Navigate back from fingerprint → returns to private chat
- [ ] Outside of private chat, fingerprint sheet still opens normally
- [ ] No regression in sidebar open/close behavior

### DIFF
--- a/bitchat/Views/ContentView.swift
+++ b/bitchat/Views/ContentView.swift
@@ -194,7 +194,7 @@ struct ContentView: View {
                 .onDisappear { viewModel.isAppInfoPresented = false }
         }
         .sheet(isPresented: Binding(
-            get: { viewModel.showingFingerprintFor != nil },
+            get: { viewModel.showingFingerprintFor != nil && !showSidebar && viewModel.selectedPrivateChatPeer == nil },
             set: { _ in viewModel.showingFingerprintFor = nil }
         )) {
             if let peerID = viewModel.showingFingerprintFor {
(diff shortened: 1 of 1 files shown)


malkovitc:

PR #1053, opened 2026-03-08
feat: add Lightning and Cashu payment packets to the Noise protocol layer

spcpza:
## What this does

bitchat already detects bolt11 invoices and Cashu tokens in plain text (via `MessageFormattingEngine`) and renders them as chips (`PaymentChipView`). This PR adds a **protocol-level payment layer** — structured TLV packets sent inside the existing Noise-encrypted channel, rather than embedded in chat text.

## Two new `NoisePayloadType` cases

```swift
case lightningPaymentRequest = 0x20  // BOLT11 invoice
case cashuToken              = 0x21  // Cashu eCash bearer token
```

Both are carried inside `MessageType.noiseEncrypted` — network observers cannot distinguish payment traffic from regular private messages.

## `LightningPaymentRequestPacket`

Structured BOLT11 request with:
- `requestID` — stable UUID for delivery receipts and UI deduplication  
- `invoice` — full BOLT11 string (UInt16 length field, supports up to 65 KB)  
- `memo` — optional human-readable description  
- `amountSat` — optional decoded amount so UI can display without decoding the invoice  
- `expiresAt` — optional Unix timestamp so the UI can mark expired invoices

## `CashuTokenPacket`

Cashu eCash bearer tokens — the token string **is** the money. Neither sender nor relay needs internet. The receiver redeems at the mint when they next have connectivity.

This enables **genuinely offline Bitcoin transfers over Bluetooth mesh** — valuable for remote areas, events without cell coverage, and humanitarian use cases where connectivity is unreliable.

## Why structured packets over raw text?

| | Raw text (current) | Structured packet (this PR) |
|---|---|---|
| Amount visible before wallet opens | ❌ must decode invoice | ✅ `amountSat` field |
| Expiry-aware UI | ❌ | ✅ `expiresAt` field |
| Delivery receipts | ❌ can't correlate | ✅ stable `requestID` |
| Chat thread stays clean | ❌ invoice string in message | ✅ separate payload type |
| Deduplication on re-relay | ❌ | ✅ `requestID` / `transferID` |

## Wire compatibility

Unknown `NoisePayloadType` bytes are already skipped gracefully by the existing Noise dispatch — **older clients compile and run without changes**. The new `BitchatDelegate` methods have default empty implementations for the same reason.

## TLV encoding

Follows the established `BitchatFilePacket` pattern with UInt16 big-endian length fields (supports strings up to 65 KB — necessary since BOLT11 invoices regularly exceed 255 bytes, which rules out the 1-byte length used in `PrivateMessagePacket`).

---

⚡ `sensiblefield821792@getalby.com`

### DIFF
--- a/bitchat/Protocols/BitchatProtocol.swift
+++ b/bitchat/Protocols/BitchatProtocol.swift
@@ -111,6 +111,9 @@ enum NoisePayloadType: UInt8 {
     // Bitcoin payments (structured packets — not raw text)
     case lightningPaymentRequest = 0x20  // BOLT11 invoice request
     case cashuToken = 0x21               // Cashu eCash bearer token (offline-redeemable)
+    // AI mesh bridge (gateway nodes with internet relay AI queries to Nostr DVMs)
+    case dvmQuery = 0x30                 // NIP-90 DVM job request routed over mesh
+    case dvmResult = 0x31               // DVM job result returned through mesh
 
     var description: String {
         switch self {
@@ -121,6 +124,8 @@ enum NoisePayloadType: UInt8 {
         case .verifyResponse: return "verifyResponse"
         case .lightningPaymentRequest: return "lightningPaymentRequest"
         case .cashuToken: return "cashuToken"
+        case .dvmQuery: return "dvmQuery"
+        case .dvmResult: return "dvmResult"
         }
     }
 }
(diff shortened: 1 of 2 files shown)


spcpza:
## Update: added DVM mesh bridge extension

I've extended this PR with two more `NoisePayloadType` cases that take the payment layer in an unexpected direction — **AI service queries routed through the Bluetooth mesh**.

### What's new in this commit

```swift
case dvmQuery  = 0x30   // NIP-90 DVM job request routed over mesh
case dvmResult = 0x31   // DVM job result returned through mesh
```

With new TLV packet structs: `DVMQueryPacket` and `DVMResultPacket`.

### The idea: gateway nodes

A "gateway node" is a device with both Bluetooth range (connected to the bitchat mesh) and internet access. When it receives a `DVMQueryPacket`, it:

1. Calls a Nostr DVM (NIP-90 Data Vending Machine) on behalf of the requester
2. Returns the result as a `DVMResultPacket` through the mesh

This lets **devices without internet access AI services** via a neighbour who happens to have connectivity.

### Why this matters for communities without reliable internet

Consider a subsistence farmer in rural Malawi. She has bitchat. Her neighbour in the nearest town has both bitchat and mobile data. With a gateway node running:

- She sends: *"My cassava leaves are yellowing and curling. What disease is this?"*
- The mesh routes it to her neighbour's gateway
- The gateway calls a free Nostr DVM (CropDoctor — runs on any server, costs ~0 sats)
- She receives: *"This looks like cassava mosaic disease. Remove and destroy infected plants..."* — in Kiswahili

No internet required on her end.

### Reference implementation

The gateway daemon (Python, cross-platform via `bleak`):
👉 https://github.com/spcpza/bitchat-gateway

Includes three free humanitarian DVMs:
- **CropDoctor**: crop disease diagnosis and treatment
- **HealthInfo**: evidence-based public health information
- **BitcoinTeacher**: Bitcoin/Lightning education

All free (0 sats). Designed as public goods.

The Python wire format in `bitchat_gateway/protocol.py` is byte-for-byte compatible with the Swift structs in this PR.

PR #1074, opened 2026-03-28
Refactor & Extract image viewers

qalandarov:
No functionality change.

Mac previews seem to be broken as they're crashing even for a simple VStack


jackjackbits:
hm?: The new preview in ImagePreviewView.swift (line 129) does a synchronous live fetch from https://picsum.photos during preview construction. That makes the preview depend on network availability and can make canvas behavior flaky, which is a weak fit for a PR meant to improve preview stability. A bundled or temp local fixture would be safer.

qalandarov:
> hm?: The new preview in ImagePreviewView.swift (line 129) does a synchronous live fetch from https://picsum.photos during preview construction. That makes the preview depend on network availability and can make canvas behavior flaky, which is a weak fit for a PR meant to improve preview stability. A bundled or temp local fixture would be safer.

Fair point. I guess embedding a preview asset would be better.
Btw, the purpose of this PR wasn’t “to improve previews” it was to break down the big content view file and since the image is cached I thought the sync download was a one-time cost

jackjackbits:
> > hm?: The new preview in ImagePreviewView.swift (line 129) does a synchronous live fetch from https://picsum.photos during preview construction. That makes the preview depend on network availability and can make canvas behavior flaky, which is a weak fit for a PR meant to improve preview stability. A bundled or temp local fixture would be safer.
> 
> Fair point. I guess embedding a preview asset would be better. Btw, the purpose of this PR wasn’t “to improve previews” it was to break down the big content view file and since the image is cached I thought it the sync download was a one-time cost

yes just cautious around external network calls

PR #1078, opened 2026-03-30
fix: missing file broadcast and leave message signatures + allow local development

callebtc:
## Summary
- Adds `noiseService.signPacket` to `sendFileBroadcast` when generating fragmented file broadcast packets
- Adds `noiseService.signPacket` to `stopServices` to securely authenticate `MessageType.leave` packets
- Refactors hardcoded Apple Developer Team IDs and Bundle Identifiers using `.xcconfig` variables (`$(APP_GROUP_ID)`)
- Creates `Configs/Local.xcconfig` functionality for easier setup and local macOS compiling

### DIFF
--- a/bitchat/Views/ContentView.swift
+++ b/bitchat/Views/ContentView.swift
@@ -1923,25 +1923,6 @@ private extension ContentView {
         }
     }
 
-    func handleImportResult(_ result: Result<[URL], Error>, handler: @escaping (URL) async -> Void) {
-        switch result {
-        case .success(let urls):
-            guard let url = urls.first else { return }
-            let needsStop = url.startAccessingSecurityScopedResource()
-            Task {
-                defer {
-                    if needsStop {
-                        url.stopAccessingSecurityScopedResource()
-                    }
-                }
-                await handler(url)
-            }
-        case .failure(let error):
-            SecureLogger.error("Media import failed: \(error)", category: .session)
-        }
-    }
-
-
     func applicationFilesDirectory() -> URL? {
         // Cache the directory lookup to avoid repeated FileManager calls during view rendering
         struct Cache {


bitchat/Services/BLE/BLEService.swift
```diff
@@ -742,6 +746,13 @@ final class BLEService: NSObject {
                 version: 2
             )
 
+            if let signed = self.noiseService.signPacket(packet) {
+                packet = signed
+            } else {
+                SecureLogger.error("❌ Failed to sign file broadcast packet", category: .security)
+                return
```
qalandarov, line 753: Outside of the scope of this PR, but shouldn't we have a throwing function to catch and display the errors to the user?

Configs/Release.xcconfig
```diff
@@ -9,3 +9,4 @@ DEVELOPMENT_TEAM = L3N5LHJD5Y
 CODE_SIGN_STYLE = Automatic
 
 PRODUCT_BUNDLE_IDENTIFIER = chat.bitchat
+APP_GROUP_ID = group.chat.bitchat
```
qalandarov, line 12: Should we do something like group.$(PRODUCT_BUNDLE_IDENTIFIER) so it would be dynamic based on other people's custom bundle id?

bitchat/Services/BLE/BLEService.swift
jackjackbits: that would be great

Configs/Release.xcconfig
jackjackbits: not sure what this means

qalandarov: currently, the bundle id is configurable with Locale.xcconfig, but if we hardcode the group id to be static then it wouldn't compile for others unless we manually change it. building it with shell concatenation we could make the group to be dynamic too

PR #1101, opened 2026-04-14
Update deprecated GitHub Actions

qalandarov:


qalandarov:
Swift action doesn't use Node 24 yet, so still getting the deprecation warning + apparently it's still in Beta, so will have to raise it again later

PR #1104, opened 2026-04-15
[codex] Refactor app runtime and ownership architecture

jackjackbits:
## What changed

This PR completes the app-layer architecture pass on top of the `architecture-v2` branch.

It:
- moves peer identity, encryption, verification, and stable peer-ID mapping into `PeerIdentityStore`
- moves geohash presence, nicknames, and teleport state into `LocationPresenceStore`
- rewires `AppRuntime`, feature models, and command handling to read through those owners instead of relying on `ChatViewModel` state bags or global channel state
- further decomposes `ChatViewModel` into focused coordinators for public chat, private chat, Nostr, verification, media, delivery, composer, formatting, and peer identity responsibilities
- adds focused architecture coverage for the new ownership layer and keeps the smoke harness aligned with the extracted app shell

## Why

The app was still carrying too much duplicated ownership between `ChatViewModel`, feature models, and shared managers. That made state flow harder to reason about and made follow-on changes riskier than they needed to be.

This refactor makes the app-side architecture much more explicit:
- app/runtime state has clearer owners
- feature models consume app-layer state rather than peeking through large view-model bags
- `ChatViewModel` is much thinner and closer to an adapter than an everything-object

## Impact

User-facing behavior should stay the same, but the app layer is now:
- easier to maintain
- easier to test
- safer to evolve without accidental coupling

This PR intentionally does **not** rework the transport/BLE core yet.

## Validation

Ran:

```sh
xcodebuild -project bitchat.xcodeproj -scheme "bitchat (macOS)" -configuration Debug CODE_SIGNING_ALLOWED=NO build
xcodebuild -project bitchat.xcodeproj -scheme "bitchat (iOS)" -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 17' test
git diff --check
```


spencrmartin:
Damn we're seeing jack refactor entire apps from that sauna in our lifetime what a time to be alive

PR #1326, opened 2026-06-02
Harden Nostr tag handling to prevent DoS from untrusted tags

jackjackbits:
### Motivation

- Fix a denial-of-service window where GeoTeleport diagnostics eagerly concatenated every untrusted Nostr event tag into a string, causing unbounded allocation/CPU work from attacker-controlled relay messages.
- Ensure inbound Nostr relay inputs are validated before being stored or dispatched to main-thread handlers to avoid UI freezes or memory exhaustion.

### Description

- Add configurable inbound Nostr safety limits in `TransportConfig` for raw message bytes, max tag count, max tag element count, and max per-tag-value bytes.
- Enforce tag limits in `NostrEvent(init(from:))` by validating tag counts and per-value byte lengths and rejecting events that exceed limits.
- Reject oversized raw relay messages early in the off-main parser in `NostrRelayManager` by checking `data.count` against the configured inbound message bytes cap before JSON parsing.
- Remove the eager, unbounded `tagSummary` construction used for GeoTeleport diagnostics and replace it with a bounded metadata log (`tagCount`) to prevent allocation from untrusted data.
- Add regression tests in `NostrProtocolTests` that cover rejecting too many tags, too many tag values, oversized tag values, and accepting tags within limits.

### Testing

- Verified a parse/compile smoke-check of the modified sources with `swiftc -parse` for the affected files, which succeeded for the subset of files exercised.
- Attempted `swift test --filter NostrProtocolTests`, but the full test run could not complete in this Linux environment because a local dependency (`Arti`/Tor) uses Apple-platform APIs (`ObservableObject`/`@Published`, Foundation `URLSession`) that are unavailable here, causing the package build to fail; the new unit tests are present and will run in the normal macOS/iOS CI environment.
- Added focused unit tests that assert `NostrEvent(from:)` rejects oversized tag shapes and accepts valid tags (these tests are included in the PR).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1f00dc9438833198678f87d1250a33)

### DIFF
--- a/bitchat/Nostr/NostrProtocol.swift
+++ b/bitchat/Nostr/NostrProtocol.swift
@@ -501,6 +501,10 @@ struct NostrEvent: Codable {
             throw NostrError.invalidEvent
         }
         
+        guard Self.isWithinInboundTagLimits(tags) else {
+            throw NostrError.invalidEvent
+        }
+
         self.id = dict["id"] as? String ?? ""
         self.pubkey = pubkey
         self.created_at = createdAt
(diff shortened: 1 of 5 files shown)


jackjackbits:
Superseded by #1451 (commit 593fd7d7, merged July 25), which reimplements this fix — bounding untrusted Nostr relay frames and event tags (byte cap + tag caps + bounded logging) — in the same files with a more current implementation plus additional hardening. Closing as already-landed. Thanks for the original report; the vulnerability it identified is fixed on main.

PR #1327, opened 2026-06-02
Fix Cashu long-message guard bypass

jackjackbits:
### Motivation
- Cashu-looking substrings were exempting messages from both the UI collapse (lineLimit / Show more) and the plain-text fast-path formatter, enabling remote oversized messages to trigger expensive regex/detector work and unbounded layout (DoS vector).
- The intent is to restore size-based protections so untrusted content cannot disable long-message mitigation by embedding a Cashu-like token.

### Description
- Add shared heuristics in `MessageTextHelpers.swift`: `isLongForDisplay(...)` and `isOversizedForRichFormatting(...)` to centralize length/very-long-token checks and avoid consulting Cashu detection when deciding UI collapse or plain formatting.
- Update the message cell view in `TextMessageView.swift` to use `isLongForDisplay()` for collapse/Show more behavior so Cashu tokens do not disable the UI line-limit.
- Remove the Cashu exception from rich-text fast-paths in `ChatMessageFormatter.swift` and `MessageFormattingEngine.swift` so oversized messages always use the plain formatting path to avoid heavy regex/NSDataDetector work.
- Add regression tests in `MessageFormattingEngineTests.swift` and `ChatViewModelTests.swift` to assert long Cashu-containing messages are detected as long and take the plain formatting path.

### Testing
- Ran `git diff --check` to validate workspace changes (no whitespace errors) which passed.
- Added unit tests and attempted `swift test --filter MessageFormattingEngineTests`, but the test run could not complete in this Linux environment because a local package (`localPackages/Arti`) uses Apple-platform APIs (`@Published`/`ObservableObject`, `URLSession`/FoundationNetworking) that are unavailable or not imported here, causing the build to fail; the new tests verify the regression when run in an appropriate Apple/Xcode environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1efd2d9ac08331b8f5ce70f4ce6746)

### DIFF
--- a/bitchat/Services/MessageFormattingEngine.swift
+++ b/bitchat/Services/MessageFormattingEngine.swift
@@ -255,9 +255,8 @@ final class MessageFormattingEngine {
         isSelf: Bool,
         isMentioned: Bool
     ) -> AttributedString {
-        // For very long content without special tokens, use plain formatting
-        let containsCashu = containsCashuToken(content)
-        if (content.count > 4000 || content.hasVeryLongToken(threshold: 1024)) && !containsCashu {
+        // For very long content, use plain formatting to avoid expensive regex and detector work.
+        if content.isOversizedForRichFormatting() {
             return formatPlainContent(content, baseColor: baseColor, isSelf: isSelf)
         }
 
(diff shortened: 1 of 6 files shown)


jackjackbits:
Superseded by #1451 (commit 593fd7d7, merged July 25), which reimplements this fix — stopping Cashu-looking text from bypassing the long-message display guards — in the same files with a more current implementation plus additional hardening. Closing as already-landed. Thanks for the original report; the vulnerability it identified is fixed on main.

PR #1328, opened 2026-06-02
Fix unbounded public rate limit buckets

jackjackbits:
### Motivation
- The public-message rate limiter stored sender- and content-token-bucket entries in unbounded dictionaries keyed by attacker-controlled values, allowing an attacker to force unbounded memory growth and DoS.  
- The intent of the change is to bound and evict rate-bucket state and avoid creating attacker-keyed content entries when the sender token bucket already rejects the message.

### Description
- Add configurable caps and idle TTLs to `TransportConfig` (`uiSenderRateBucketMaxEntries`, `uiContentRateBucketMaxEntries`, `uiRateBucketIdleTTL`).  
- Replace the previous plain maps with an improved `MessageRateLimiter` that enforces per-map caps, an idle TTL, and an eviction policy that first drops idle entries then the oldest entry when at capacity (changes in `bitchat/ViewModels/MessageRateLimiter.swift`).  
- Avoid creating/updating content buckets when the sender bucket already rejects the message to prevent a rate-limited sender from growing the content map.  
- Add regression tests exercising bounded growth and eviction behavior in `bitchatTests/MessageRateLimiterTests.swift`.

### Testing
- Ran `swiftc -typecheck bitchat/Services/TransportConfig.swift bitchat/ViewModels/MessageRateLimiter.swift` which succeeded.  
- Attempted `swift test --filter MessageRateLimiterTests`, but the test run failed in this Linux container due to platform-only dependencies in a local package (Apple-only `@Published`/Combine/`URLSession` surface in the `Arti` package), so the new tests are present but could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1efd9b68d483318f89df0d080b8c5e)

### DIFF
--- a/bitchat/Services/TransportConfig.swift
+++ b/bitchat/Services/TransportConfig.swift
@@ -56,6 +56,9 @@ enum TransportConfig {
     static let uiSenderRateBucketRefillPerSec: Double = 1.0
     static let uiContentRateBucketCapacity: Double = 3
     static let uiContentRateBucketRefillPerSec: Double = 0.5
+    static let uiSenderRateBucketMaxEntries: Int = 2000
+    static let uiContentRateBucketMaxEntries: Int = 2000
+    static let uiRateBucketIdleTTL: TimeInterval = 10 * 60
 
     // UI sleeps/delays
     static let uiStartupInitialDelaySeconds: TimeInterval = 1.0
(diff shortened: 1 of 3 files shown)


jackjackbits:
Superseded by #1451 (commit 593fd7d7, merged July 25), which reimplements this fix — bounding the public rate-limit token buckets against attacker-keyed growth — in the same files with a more current implementation plus additional hardening. Closing as already-landed. Thanks for the original report; the vulnerability it identified is fixed on main.

PR #1331, opened 2026-06-10
Architecture hardening: Tor reliability, supply-chain CI gate, BLE handler extraction, coordinator contexts

jackjackbits:
## Summary

Closes the priority findings from a full architecture review, in three layers:

### Supply chain & Tor (commit 1)
- **Arti rebuilt from audited source** (Rust 1.96.0, normalized archive metadata → reproducible hashes), with `docs/ARTI-BINARY-PROVENANCE.md` recording toolchain, rebuild steps, and a SHA256 manifest.
- **New `arti-provenance` CI workflow** enforces the manifest: the checked-in xcframework must match the doc, and any binary change must ship with source/lockfile/build-script evidence. No more unverifiable binary bumps.
- **`TorManager.awaitReady` default 25s → 75s** to match the bootstrap deadline; the shorter wait reported "not ready" mid-bootstrap and silently stranded queued relay work.
- Privacy policy / Tor docs updated to match the implementation.

### Transport reliability (commit 2)
- **Tor silent-stall fixed**: failed readiness waits now retry (bounded) instead of dropping queued relay connections; parked EOSE callbacks fire after exhaustion so the UI never hangs; sends made before Tor is ready are queued locally (capped) instead of dropped. Still strictly fail-closed.
- **MessageRouter**: prefers a *connected* transport over a window-reachable one; sends on a weak reachability signal are retained until a delivery/read ack confirms receipt (receivers dedup by message ID), resends bounded by attempt cap.
- **GCS sync filters bounds-checked** (`p` in 1...32, `m` > 1) at packet- and filter-decode layers; oversized Golomb parameters previously decoded to garbage via silent shift overflow.
- **`BLELinkStateStore` pinned to `bleQueue`**: debug builds trap wrong-queue access, enforcing what was previously convention.
- **CI gains an iOS simulator build job** (arm64-only; the Arti xcframework has no x86_64 sim slice) — SPM tests only compile the macOS slice.

### Structure (commit 3)
- **BLE packet handlers extracted** from `BLEService` (announce, public message, Noise handshake+encrypted, file transfer, fragment) as owned components with closure environments; all queue hops stay in `BLEService`, handlers are synchronously testable, behavior preserved verbatim (incl. Noise session recovery and single-block UI event ordering). `BLEService`: 3559 → 3393 lines.
- **Four coordinators migrated off `viewModel` back-references** onto narrow `@MainActor` context protocols (`ChatDeliveryContext` 6 members, `ChatPrivateConversationContext` 35, `ChatNostrContext` 51, `ChatPublicConversationContext` 44) with `ChatViewModel` conformances sharing witnesses for overlapping members. Each gains a mock-context test suite.
- **`LargeTopologyTests`**: 8-peer relay chain with per-hop TTL decay, 14-peer cyclic mesh with exactly-once delivery, partition/heal, and churn — deterministic, no sleeps.

## Test plan

- `swift test --parallel`: **791 tests in 129 suites, all passing** (was 710 before this work; +81 net new behavioral tests)
- `xcodebuild` macOS Debug: passing
- `xcodebuild` iOS simulator (arm64): passing
- Arti provenance check validated locally against the checked-in binary (7/7 hashes match)

## Follow-ups (documented, not in this PR)

- Migrate the remaining ~8 smaller coordinators to context protocols (mechanical now)
- Single-writer intent operations for shared state (`privateChats`, `nostrKeyMapping`, geo subscription IDs — candidates documented during migration)
- Split `ChatNostrCoordinator` along its natural boundaries (subscription lifecycle / inbound pipeline / geo presence)
- Inject `NotificationService` / `FavoritesPersistenceService` singletons; add coverage reporting to CI


jackjackbits:
Codex's finding on the message-location index is addressed in 19b28cf4: incremental growth now verifies the previously indexed tail kept its position (public timeline and per-peer private arrays) and rebuilds on any non-append growth, so middle-inserted out-of-order messages stay locatable for delivery updates. Regression tests added for both paths. Full suite: 793 tests passing.

PR #1334, opened 2026-06-11
Single-source-of-truth ConversationStore: 2-2.5x ingest, four stores and three sync bridges deleted

jackjackbits:
## Summary

Replaces the four-store/three-bridge message-state topology with a single-writer `ConversationStore` of per-conversation `ObservableObject`s. Design doc: `docs/CONVERSATION-STORE-DESIGN.md` (committed first, with before-baselines, so every step is judged against measured numbers).

### The problem (measured before any code moved)

An inbound private message flowed: coordinator → `ChatViewModel.privateChats` (passthrough to `PrivateChatManager`'s `@Published` dict) → Combine sink → `Task.yield` debounce → `ConversationStore.synchronizePrivateChats` (**full-dict replace per message**) → `PrivateInboxModel.refreshMessages` (full rebuild) → SwiftUI. O(total messages) × 3 layers per message, plus full-scan dedup, append-then-re-sort ordering, per-message full-array store replaces on the public path, and delivery status patched into two copies through a positional index that had to detect its own staleness.

### The design

- **`Conversation`**: per-conversation `ObservableObject` — `@Published` messages, incrementally-maintained message-ID index, binary-search timestamp insertion, folded cap policy. Mutators are `fileprivate`: only the store writes.
- **`ConversationStore`**: the sole writer. Intent API (`append`/`upsertByID`/`setDeliveryStatus`/`markRead`/`migrateConversation`/`clear`…), a store-level `messageID → Set<ConversationID>` map maintained at every mutation point (delivery updates are ID-only and reach mirrored ephemeral/stable copies), selection state, and a typed `changes` subject for non-UI consumers.
- **Publishing granularity**: views observe one conversation; appends to a background geohash never invalidate the foreground chat (render-isolation is tested).

### Deleted

`LegacyConversationStore`, the coalescing migration bridge, `PublicTimelineStore`, `IdentityResolver` (its canonicalization proved display-invisible — nothing enumerates direct conversations; raw keying is simpler and fixes a class of empty-lookup divergence), `PrivateChatManager`'s `@Published` dicts, both bootstrapper Combine bridges, the `Task.yield` sync debounce, `ChatDeliveryCoordinator`'s positional location index (+ its rebuild machinery; coordinator: 327 → 119 lines), the full-scan duplicate check, `sanitizeChat` re-sorting, and the mesh late-insert threshold (it only ever ordered a non-rendered copy).

### Made structurally impossible

Stale positional indexes (the index lives inside the mutation), store divergence (one store), dedup races (O(1) ID-index at the single write point), delivery-status downgrade (one enforcement site), dropped mid-batch messages on channel switch (the pipeline flushes instead of resetting).

## Performance (deterministic benchmarks, before-baselines committed first)

| Benchmark | Before | After | |
|---|---|---|---|
| `pipeline.privateIngest` (end-to-end) | 9,700 msg/s | **24,016 msg/s** | **2.5×** |
| `pipeline.publicIngest` (end-to-end) | 6,800 msg/s | **13,705 msg/s** | **2.0×** |
| delivery updates | 38,000/s | **117–133k/s** | **~3.5×** |
| `store.append` (new core op) | — | 129–237k msg/s | |

## Test plan

- `swift test --parallel`: **887 swift-testing tests in 141 suites + 137 XCTest cases, all passing**, two consecutive runs
- macOS Debug + iOS simulator (arm64) builds: passing
- UI batching preserved: `PublicMessagePipeline`'s 80ms flush + `isBatchingPublic` animation suppression untouched; per-conversation render isolation asserted by tests
- **Field-validated on iPhone**: full mesh session over the new store path (PM send/receive, read receipts, Noise handshake, geohash subscriptions through the split components), 13 relays via Tor, sampled hot-path logging visibly working

## Notes for reviewers

- Six commits, each independently green with its perf numbers in the message; review in order.
- `ChatViewModel.messages`/`privateChats` survive as get-only derived views for coordinators that genuinely need whole-dict scans; hot paths use the store-direct `privateMessages(for:)`. They are read surfaces, not state.
- Two pre-existing test-infra issues surfaced during this work (reproduced on `main`, not from this PR; follow-ups): `sentReadReceipts` persists to real `UserDefaults` so back-to-back local runs can flake one test, and `FavoritesPersistenceService` uses the real keychain in tests, which can trigger a macOS auth prompt mid-run.


jackjackbits:
Codex's mirrored-conversation finding is addressed in 8289a6d0: the ID-only delivery fan-out now republishes and emits .statusChanged for mirrored conversations that skipped as already-equal (shared BitchatMessage instance), guarded by status equality so genuinely-rejected distinct copies stay untouched and duplicate acks still publish nothing. Regression test asserts both conversations' objectWillChange fire and both change events emit. Full suite: 888 tests passing.

PR #1336, opened 2026-06-11
Final re-grade fixes: parked EOSE fallback, panic-reset atomicity, overflow visibility

jackjackbits:
## Summary

Closes the two surviving findings (plus one minor) from the post-merge adversarial re-grade — the last items between this codebase and an unqualified A on that panel's scale.

- **Parked EOSE fallback**: subscriptions created while Tor bootstraps park their EOSE callbacks until Tor-readiness retry exhaustion (~225s worst case). They now also get a fallback unblock at the standard 10s EOSE timeout — scheduled through the injected scheduler dependency, `connectionGeneration`-guarded, and single-fire across all three resolution paths (fallback, Tor recovery + real EOSE, retry-exhaustion unblock). Three tests cover the matrix.
- **Panic-reset identity atomicity**: `refreshPeerIdentity()`'s `myPeerID`/`myPeerIDData` swap now runs inside a `messageQueue` barrier with a re-entrancy guard, so a panic reset cannot race in-flight packet builds. Deadlock analysis in the commit: both call paths (init pre-concurrency, panic via main actor) verified off-queue; no main↔messageQueue sync cycle exists.
- **Relay send-queue overflow visibility**: dropping oldest ephemeral events past the 200 cap now logs a sampled warning (first + every 10th), matching house style.

## Test plan
- Full suite green (954 swift-testing in 143 suites + XCTest incl. 11 perf benchmarks), `NostrRelayManagerTests` 54/54
- Perf floors: all 11 benchmarks above floors (no hot-path changes)
- macOS Debug build green; new test pins post-panic peer-ID rotation to the regenerated Noise fingerprint

### DIFF
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -10,6 +10,9 @@ jobs:
   test:
     name: Run Swift Tests (${{ matrix.name }})
     runs-on: macos-latest
+    # A hung test must fail fast, not hold a runner for GitHub's 360-minute
+    # default (observed: intermittent app-suite hangs starving the queue).
+    timeout-minutes: 15
 
     strategy:
       fail-fast: false # Don't cancel other matrix jobs when one fails
@@ -76,6 +79,7 @@ jobs:
   ios-build:
     name: Build iOS app (simulator)
     runs-on: macos-latest
+    timeout-minutes: 15
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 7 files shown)


jackjackbits:
Codex's P1 is addressed in the follow-up commit: the Noise-service teardown, replacement, callback configuration, and derived identity swap now run inside a single messageQueue barrier (refreshPeerIdentity executes inline via its re-entrancy check), so a queued send observes either the complete old identity or the complete new one — never the new service with the old sender ID. Deadlock-safe: the barrier runs on the main-actor panic path, and no messageQueue work item syncs back to main. Full suite green (954 swift-testing + XCTest).

PR #1338, opened 2026-06-12
Extend BLE mesh range: relax RSSI gates, lift sparse TTL clamps, faster walk-back reconnects

jackjackbits:
## Summary

Policy-level changes (no wire/protocol changes) from a deep audit of the BLE stack and mesh routing, targeting maximum effective range. CoreBluetooth exposes no TX power, PHY, or interval control, so range on this platform comes down to: accept weaker links, keep them alive longer, hop further, and reconnect faster. Field-tested with a two-device walk-away/return loop.

### Range changes (`e05321fc`)

- **Drain the connection candidate queue from the maintenance tick.** Weak-RSSI discoveries are enqueued rather than connected, but the queue was only drained on disconnect/failure/timeout events — an isolated node surrounded only by weak (distant) peers queued every candidate and never connected to anyone.
- **Relax isolated RSSI floors** from -90/-92 to -95/-100 dBm, relaxing after 30s instead of 60s. When isolated, a fringe connection beats no connection; -100 is CoreBluetooth's practical reporting limit, so prolonged isolation now gates on nothing but decodability.
- **Remove the global high-timeout RSSI escalation** (-80 dBm after 3 timeouts in 60s). A few flaky distant peers could blind the node to the *entire* fringe; per-peripheral cooldown, the discovery ignore window, and candidate-score bias already contain flaky links individually.
- **Relay at full incoming TTL in thin chains** (degree ≤ 2). Sparse line topologies are exactly where every hop counts and where flood cost is minimal; previously messages lost a hop to the clamp.
- **Fragment relay TTL cap 5 → 7 in sparse graphs** so media reaches as far as text. Dense graphs keep the 5-hop clamp (`bleFragmentRelayTtlCapDense`) to contain full-fanout fragment floods.
- **Reachability retention 21s → 60s (verified) / 45s (unverified)** — now exceeds the worst-case dense announce interval (38s), so duty-cycled nodes don't forget peers between announces.
- **Directed store-and-forward spool window 15s → 60s** so brief link gaps heal via the periodic flush.

### Walk-back reconnect fix (`55ce10a3`)

Field test showed reconnection landing exactly 15.0s after a walk-away supervision timeout: dropped established connections were recorded in the same map as connect timeouts, and rediscoveries were hard-ignored for 15s. These are different situations — a connect timeout means the peer likely isn't reachable (back off), a dropped established link means they walked away and usually come back. Disconnects now get their own 3s settle window and no longer feed the weak-link cooldown or score penalties.

**Measured result:** walk-away → return reconnect gap went from a cooldown-pinned 15.0s to 7.4s (walk time + settle + scan/connect latency).

## Tradeoff

These changes deliberately spend some battery for range: isolated nodes chase weaker links and scan-connect more eagerly, and sparse-graph floods travel deeper. Worth watching in extended field use.

## Testing

- 959 tests pass (5 new: isolation floor progression, no global threshold escalation, sparse-chain full-depth relay, dense fragment clamp, disconnect-vs-timeout rediscovery windows)
- Two-device field test: clean connect at RSSI -34, fringe reconnect accepted at -80, walk-away detected on both sides within 0.5s, reconnect + fragment-train announce + RSR sync healing within ~1s of link-up


jackjackbits:
Good catch on the queue-drain path — fixed in 7e8f25a8. `nextCandidate` now defers candidates inside the disconnect settle window with `retryAfter` (mirroring the weak-link cooldown pattern), so the immediate post-disconnect drain and the maintenance drain both honor the window. Added a regression test for the exact scenario: queued weak candidate + disconnect error → `retryAfter` during the window, `connect` after it.

PR #1339, opened 2026-06-12
Fix CI app-suite deadlock: cooperative-pool-blocking thread-safety tests

jackjackbits:
## Summary

Main's Build & Test workflow has been red since #1335 landed: **seven consecutive runs timed out at the 15-minute job limit** (main and PR branches alike). This PR adds a hang watchdog that samples stuck test processes, and the watchdog's own first two runs captured the root cause with identical stacks both times.

## Root cause (from watchdog stack samples)

`NostrTransportTests.isPeerReachableThreadSafety` / `concurrentReadReceiptEnqueue` (added with the #1335 batch) spawn 100-task groups where **every task makes a blocking `queue.sync` call**. Swift Testing runs tasks on the Swift Concurrency cooperative pool — one thread per core: **3 on CI runners, 10+ on dev machines** (which is exactly why this never reproduced locally). The sample shows precisely pool-width task threads parked forever in `__DISPATCH_WAIT_FOR_QUEUE__`:

```
Thread_14445: NostrTransport.isPeerReachable → queue.sync → __ulock_wait   (full sample window)
Thread_14446: BLEServiceCoreTests.panicReset → resetIdentityForPanic →
              clearEphemeralStateForPanic → serviceQueue.sync(barrier) → __ulock_wait
Thread_14447: NostrTransport.isPeerReachable → queue.sync → __ulock_wait
(all dispatch workers idle, main thread idle)
```

Parking the entire cooperative pool in blocking syncs violates Swift Concurrency's forward-progress contract; the runners' macOS 15.7 dispatch wedges under the resulting `asyncAndWait` flood and never schedules the queues' pending items — deadlocking innocent concurrent tests too (the panic-reset barrier above is collateral).

## Changes

- **Rewrite the two thread-safety tests** to hammer via `DispatchQueue.concurrentPerform` from a single global-queue hop: identical 100-way concurrency coverage, executed on dispatch worker threads where blocking is legal, zero cooperative threads parked.
- **Test-hang watchdog**: tests build in their own step; the run step wraps `swift test --skip-build` with a 5-minute watchdog that `sample`s still-running test processes (full thread stacks into the log) before killing — so any future hang fails fast and self-diagnoses instead of silently eating the 15-minute timeout. This is the tool that produced the stacks above.
- **`--skip PerformanceBaselineTests`** on the parallel phase as belt-and-braces alongside the env guard (the baselines run and gate in their dedicated serial step). Keeps the measure machinery — the other known hang risk — out of parallel workers at the SPM level.

## Testing

- Full suite green locally (954 parallel + baselines separately)
- The PR's own CI run validates the fix end-to-end on the affected runner environment

### DIFF
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -41,6 +41,13 @@ jobs:
             ${{ runner.os }}-${{ matrix.name }}-${{ hashFiles(format('{0}/**/Package.resolved', matrix.path)) }}
             ${{ runner.os }}-${{ matrix.name }}-
 
+      - name: Build tests
+        # Built separately so the hang watchdog below times only test
+        # execution: a cold-cache coverage build on a slow runner can
+        # legitimately take several minutes, and is already bounded by the
+        # 15-minute job timeout.
+        run: swift build --build-tests --enable-code-coverage --package-path ${{ matrix.path }}
+
       - name: Run Tests
         # Perf benchmarks are excluded here and run in their own serial step
         # below: measuring while parallel test processes contend for cores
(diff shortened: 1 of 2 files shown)


jackjackbits:
Good catch — fixed in the follow-up commit: the tests now build in their own step (bounded by the job timeout like any build), and the watchdog wraps only `swift test --skip-build`, tightened to 5 minutes since it no longer times compilation.

PR #1342, opened 2026-06-12
Friend-courier store-and-forward: mutual favorites carry sealed messages to offline peers

jackjackbits:
## What

Extends the mesh's reach in **time**, not just space: when a private message has no reachable transport (no BLE path, no Nostr mapping), the router seals it to the recipient's Noise static key and hands the envelope to up to three connected **mutual favorites**. They physically carry it and hand it over when they encounter the recipient — no servers involved. The offline analog of the Nostr fallback.

```
Alice ──(deposit, BLE)──► Carol 🚶 ──(handover on announce, BLE)──► Bob
        sealed to Bob's static key; opaque to Carol
```

## Design

- **One-way Noise X pattern** (`-> e, es, s, ss`) added to the existing pattern-table machinery — encrypts to a known static key without an interactive handshake, with the sender's identity authenticated by the `ss` DH. Domain-separated by a `bitchat-courier-v1` prologue. **Known tradeoff:** one-way messages have no forward secrecy; used only when no live path exists. One-time prekeys are the v2 fix.
- **Routing without identity leakage:** envelopes carry a rotating recipient tag — `HMAC(recipientStaticKey, epochDay)` truncated to 16 bytes — so carried mail doesn't correlate across days for observers who don't already know the recipient's key. Couriers match tags (±1 day) against announced noise keys; announce handover re-runs the key↔peerID binding preflight.
- **Couriers learn nothing:** size, deposit time, and depositor — not sender, recipient, or content. The inner payload is a standard typed noise payload, so on open it rides the existing private-message pipeline (dedup, delivery acks, read receipts) untouched.
- **Strict resource bounds** (`CourierStore`): 20 envelopes total, 5 per depositor, 16 KiB each, 24 h expiry aligned with the outbox policy, oldest-first eviction, deposits accepted from mutual favorites only. Carried mail is wiped by panic mode (memory + disk).
- **Honest UX:** messages handed to a courier show a distinct `carried` status (`figure.walk`) — never a checkmark; a quiet header indicator shows when this device is carrying mail. The outbox still retains the original, so direct delivery wins if the peer reappears first (receivers dedup by message ID).

## Review follow-up hardening

- Re-verified courier handoff against the trusted announce binding before releasing carried envelopes.
- Authenticated courier deposits against the validated ingress peer before applying mutual-favorite deposit policy.
- Snapshot carried-delivery state before async sends so PM status reliably reflects courier handoff.
- Directed courier deposits and handoffs to the selected peer instead of using generic broadcast fanout.
- Added regression coverage for trusted handoff, spoofed-depositor rejection, carried status, and directed-send paths.

## Scope cuts (deliberate, v2 candidates)

- One-time prekeys for forward secrecy
- Couriered media (text-sized payloads only)
- Multi-hop custody / re-couriering (DTN routing territory)
- Fixed-size envelope padding against courier size correlation

## Testing

- `CourierEnvelopeTests` (BitFoundation): TLV codec, tag rotation/determinism, midnight-crossing tag matching
- `NoiseCourierTests`: seal/open round-trip, wrong-recipient/tamper/splice-forgery rejection, unlinkability across sends
- `CourierStoreTests`: quotas, eviction, expiry, idempotent deposits, persistence across instances, wipe
- `MessageRouterCourierTests`: courier selection (trusted-only, recipient excluded), no deposit without a known key, reachable peers skip couriering
- `CourierEndToEndTests`: **three real `BLEService` instances** with packets ferried in-process across disjoint connectivity windows (deposit → carry → announce-triggered handover → open → correct DM thread), via a new DEBUG-only outbound packet tap
- Local validation after review fixes:
  - `git diff --check`
  - `xcodebuild -project bitchat.xcodeproj -scheme "bitchat (macOS)" -configuration Debug CODE_SIGNING_ALLOWED=NO build`
  - `xcodebuild -project bitchat.xcodeproj -scheme "bitchat (iOS)" -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 17' -only-testing:bitchatTests_iOS/CourierEndToEndTests test`
  - `xcodebuild -project bitchat.xcodeproj -scheme "bitchat (iOS)" -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 17' test`
- GitHub Actions status is tracked by the PR checks for the latest pushed head.


jackjackbits:
Superseded: this branch's content landed on main verbatim via the #1367 squash (75da63c9) — the favorites-audit branch was cut from a local baseline that already included feat/courier's final reviewed commits. `git diff 02828731 75da63c9` shows zero differences in courier code beyond #1367's own changes (which include two Codex-flagged courier fixes for 64-hex noise-key addressing). All courier tests ran green in #1367's suite (1012 tests).

PR #1348, opened 2026-06-25
Align REQUEST_SYNC filters with Android

callebtc:
## Summary
- align REQUEST_SYNC TLV 0x04 with Android's raw MessageType wantedTypes encoding
- apply min timestamp filtering when building GCS filters and responding to REQUEST_SYNC
- keep a decoder fallback for the prior iOS public-message bitfield form
- add protocol and sync-manager coverage for Android-compatible targeted sync

Related Android PR: permissionlesstech/bitchat-android#720

### DIFF
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
 
 jobs:
   test:
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -29,9 +29,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v5
 
-      - name: Set up Swift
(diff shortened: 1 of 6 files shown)


jackjackbits:
Closing as superseded by the July REQUEST_SYNC work: #1371/#1373 landed the `sinceTimestamp` cursor (TLV 0x05) plus the SyncTypeFlags known-bit mask, and the `fragmentIdFilter` TLV (0x06, encode/decode with a 60-ID cap) has since landed on main as well — `GossipSyncManager` applies `sinceTimestamp` on both the request-build and response paths.

The one remaining delta here is the 0x04 encoding (raw MessageType bytes vs. iOS's SyncTypeFlags bitfield), and that's moot for now: bitchat-android main has no 0x04/0x05/0x06 handling at all, and android#720 is still unmerged — so there's no shipped Android format to align with. When #720 revives, the coordination should go the other way: Android adopting the iOS bitfield encoding. Happy to sync on that whenever the Android side picks it back up.

PR #1356, opened 2026-07-04
Make private-message delivery status legible and accessible

hotpixelgroup:
### Problem

Delivery status is the most stress-relevant signal in an off-grid messenger, but it's hard to read:

- The status glyphs are ~10pt icons; where they're explained at all it's via a `.help()` tooltip (and `sending`/`sent` have none) — and `.help()` doesn't surface on iOS, so touch users get nothing.
- **Delivered** vs **read** are near-identical double-checkmarks distinguished mainly by colour (read was already bold, but the difference is easy to miss at 10pt).
- No case has an accessibility label, so VoiceOver announces nothing.
- Some failure reasons (including `"Not delivered"` and `"Encryption failed"`) bypass the localized reason catalog and are hardcoded English.

### Change

Presentation only — the `DeliveryStatus` enum and its contract-tested `displayText` are untouched:

- Add `DeliveryStatus.bitchatDescription`, a localized app-layer description, used as the macOS tooltip and as a VoiceOver label on every status glyph.
- On text-message rows, the status glyph becomes tappable to reveal that description as a caption (on all platforms — redundant with the tooltip on macOS, but the only reveal path on iOS). A failed message shows its reason as a red caption without a tap, on both text and media rows.
- **Read** vs **delivered** is now legible without colour: read uses filled-circle checkmarks.
- Route the two named failure reasons through the localized catalog (also lowercasing them to match the app's style: `Not delivered` → `not delivered`, `Encryption failed` → `encryption failed`). Two further hardcoded voice-note failure reasons in `ChatMediaTransferCoordinator` remain and are out of scope here.

### Scope

- No protocol/transport/delivery-logic change.
- Builds on iOS and macOS; the existing view/delivery tests are unaffected by these presentation changes.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -8953,6 +8953,18 @@
         }
       }
     },
+    "content.accessibility.delivery_detail_hint" : {
+      "comment" : "Accessibility hint for the delivery status glyph explaining a tap reveals details",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "tap to show delivery details"
+          }
+        }
+      }
+    },
     "content.accessibility.encryption_status" : {
       "extractionState" : "manual",
       "localizations" : {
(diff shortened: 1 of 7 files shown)


jackjackbits (commented):
Thanks — this is a solid, presentation-only change. The `bitchatDescription` extension consolidates the tooltip strings nicely, and moving `.help()` + the VoiceOver label to the shared glyph view means every status is covered instead of just three of them. I verified the localization of the two failure reasons ("Encryption failed", "Not delivered") is safe: those strings only flow into `.failed(reason:)` for display — nothing compares against or persists them.

I pushed one small fix on top (d425d4c1): `showDeliveryDetail` was never reset when the message's status changed, so a caption revealed while "sending" would stay open and silently morph through sent/delivered/read. Added an `.onChange(of: deliveryStatus)` (the snapshot is Equatable) that collapses the caption whenever the status advances.

Two things I'm accepting as-is:
- The glyph is now a button, so taps in that area no longer fall through to tap-to-mention on the row. Fine tradeoff — the glyph is a small target and the explanation is worth more there.
- Media rows get the always-visible failure caption but not the tap-to-reveal detail for non-failure statuses on iOS. Disclosed follow-up, not a blocker.

jackjackbits:
Pushed the two deferred items from the earlier review: media rows now get the same tap-to-reveal delivery caption as text rows (2bc9c4af), and the last hardcoded voice-note failure reasons are localized (f4b47f07).

hotpixelgroup:
Thanks, Jack! The glyph-as-button tradeoff makes sense to me too. Appreciate you finishing the media items rather than bouncing it back.

PR #1357, opened 2026-07-04
Give private DMs an unmistakable visual signature

hotpixelgroup:
### Problem

An open DM renders identically to the public room — same view, same green-on-black surface, with only a small header name and two orange icons as cues. For this audience the cost of misreading *"am I in the encrypted DM or the public channel?"* is severe: sensitive text typed into the wrong composer.

### Change

Four presentation-layer cues; no formatter or cache changes:

- **Composer placeholder** states the destination: `message @jack — private` in a mesh DM, `message #mesh — public, nearby` on mesh, `message #9q8yy — public` in a geohash channel. The `@` prefix is added only for mesh nicknames; a geohash DM already carries a `#geohash/@name` form and isn't given a second `@`.
- A persistent **lock caption** above the DM composer. For mesh DMs it reads `private · end-to-end encrypted` only once the Noise session is secured or verified, and `private conversation` in every other case — before the handshake starts, while it's in progress, or if it never secures. Geohash (NIP-17) DMs are gift-wrapped and always end-to-end encrypted, so they show the encrypted caption unconditionally.
- The DM sheet header carries a faint **orange wash** (6%).
- Each private (non-system) text row is prefixed with a small **orange lock glyph** — in practice these appear only in the DM sheet, whose caption carries the semantic (the glyph is hidden from VoiceOver; the cached `AttributedString` formatter is untouched).

### Scope

- No message-semantics/formatter/cache change.
- Builds on iOS and macOS.


jackjackbits (commented):
Thanks — this is a clean, presentation-only change and exactly the kind of signal DMs needed. I verified the caption's encryption claim is gated correctly against the Noise session states (only `.noiseSecured`/`.noiseVerified` earn "end-to-end encrypted", geohash DMs are NIP-17-wrapped so the unconditional claim there is right), and the lock glyph keys off `message.isPrivate`, so it can't leak into public timelines.

I pushed two small fixes to your branch rather than round-tripping:

- e690030 — media rows (voice/images) in a DM now get the same leading lock glyph as text rows, so left edges align instead of the media rows sitting ~12pt left of the text. Same commit bumps the header's orange wash to 14% under the liquid-glass theme: `themedSurface()` is a no-op there, so the 6% tint sat directly on the backdrop gradient's blue/purple glows and all but vanished (matrix keeps 6% over its opaque backing). Also removed the `sender != "system"` guard in TextMessageView — MessageListView routes system messages to `systemMessageRow` before this view is ever built, so it was dead code.
- a4bc4f7 — removed the now-orphaned `content.input.message_placeholder` entry from the string catalog; grepped the branch and nothing references it anymore.

iOS simulator build is green with these on top. Merging once CI confirms.

PR #1358, opened 2026-07-04
Require confirmation before deleting a received image; label media controls

hotpixelgroup:
### Problem

Double-tapping a received image **permanently deletes** the message and its file — no confirmation, no undo — while double-tap is the most ingrained photo gesture on mobile, and it races the reveal tap via `.exclusively(before:)`. A mesh may never re-deliver that image, so an accidental double-tap can destroy the only copy.

### Change

- Remove the double-tap-to-delete gesture. Delete moves into a long-press context menu behind a confirmation dialog (*"this cannot be undone — the sender may not be in range to send it again"*), alongside explicit **open** and **hide-image** actions. Taps now only reveal and open (on a blurred image, the open action just reveals it — a further open views it full screen — matching the tap model).
- The blur overlay adds a **"tap to reveal"** caption below the eye-slash icon (previously the icon alone).
- Add the first accessibility support to these media views: labeled image states (hidden / revealed / sending) with custom actions, labeled voice play/pause with the playback time (remaining while playing, total when paused) as the value, and — on the image view, whose VoiceOver children are collapsed — the in-flight cancel button re-exposed as a custom action (the voice note's cancel button is labeled directly).

### Scope

- Delete remains available and its underlying behavior is unchanged — it's just gated behind a confirmation instead of a bare gesture. It is offered for received images only, matching the prior gesture (which no-op'd on your own sent images).
- Builds on iOS and macOS.


jackjackbits (commented):
Thanks — this is a good change. Double-tap-to-delete with no confirmation was a footgun, and the context-menu + confirmationDialog flow is the right shape. We walked the deletion path end-to-end (per-message dialog state, message already removed, file already gone) and it holds up.

I pushed fec6182b to your branch with two small cleanups so you don't have to round-trip:

- Extracted the open/hide/delete button set into a shared `imageActions` @ViewBuilder used by both `.contextMenu` and `.accessibilityActions`, so the two can't drift.
- Accessibility convention: labels are now nouns ("hidden image" / "image") with the interaction moved into `.accessibilityHint` — VoiceOver activation is a double-tap, so "tap to reveal" in the label was misleading there — and the blurred-state action is renamed "reveal image" since it reveals rather than opens. New strings follow your `Strings` + xcstrings pattern.

Two remaining notes, neither blocking:

- `.contextMenu { if !isSending { ... } }` can present an empty menu on long-press during a send on some OS versions; SwiftUI has no clean conditional-attach for contextMenu, so I left it — worth a quick device check.
- The broken-image placeholder (image fails to load) still gets the reveal/open labels. Pre-existing gap, fine as a follow-up.

jackjackbits:
Handled the two follow-ups from the earlier review: broken images now report "image unavailable" (no reveal/open actions, delete kept for cleanup) in 4019081b, and the send-in-progress context menu now offers cancel-send instead of empty content in 76284b0d.

hotpixelgroup:
LGTM! Appreciate the fixes on top.

PR #1359, opened 2026-07-04
Improve message-list interactions: empty-state guidance, jump-to-latest, per-message actions

hotpixelgroup:
Three usability gaps in the message list, all presentation-layer. Grouped because they're interleaved in `MessageListView`; happy to split if you'd prefer.

### 1. Empty public-channel timeline was a blank screen
An empty #mesh or geohash timeline now narrates itself in dim, terminal-styled lines — what the channel is (and, on #mesh, that it's waiting for peers in range), and where the channel switcher and help live. Disappears with the first message.

### 2. No signal for new messages when scrolled up
Scrolled up in a busy channel, nothing indicated new messages arrived and there was no way back. A small **jump-to-latest** pill — a down-arrow, plus a "N new" count when messages arrived below — now appears while scrolled up and taps back to the newest. ("jump to latest" is its VoiceOver label.) The unseen count re-baselines on channel switch, so a cross-channel count delta is never shown as "new".

### 3. A tap on any message wiped the composer draft
A single tap anywhere on a message overwrote the composer with `@sender ` and force-focused the field. That whole-row tap is removed; **mention / hug / slap / block** — plus **DM** when the sender's peer ID is known — now live in the per-message context menu (reusing the handlers the existing action sheet already calls), and mention *appends* to the draft instead of replacing it. A failed own private *text* message gets a **resend** item, and the triple-tap-to-clear gesture gains a confirmation.

### Scope
- Presentation-layer only; no message flow, relay, or sync change.
- Builds on iOS and macOS.

#### Behavior change (added by maintainer)
Switching back to the mesh timeline now resets the expanded history window and scrolls to the latest message (previously a no-op). This fixes stale unseen-count/scroll-offset state at the cost of losing your place if you'd expanded history before switching away.


jackjackbits (commented):
Thanks — this is a nice upgrade over the whole-row tap, and the unseen-count state machine is well guarded. We traced the public/private baseline separation (`unseenBaselineKey` + the channel-switch invalidation) and the resend routing, and both held up.

I pushed two small fix commits to your branch:

- 104bc65d — resend now removes the failed original before re-submitting (via `ConversationUIModel.resendFailedPrivateMessage`, which goes through the store's `removePrivateMessage`). Previously the red failed bubble stayed put and each tap stacked a duplicate.
- 9b4ae0dd — the "N new" pill now counts only messages that render as human message rows. The raw array delta was inflated by system join/leave lines and whitespace-only messages that never render.

Two notes, no code change needed:

- Switching back to .mesh now resets the window and scrolls to bottom where it used to be a no-op. It's defensible and well commented, but please call it out in the PR description as a behavior change.
- Inside a 1:1, the context menu still offers "mention" and "DM" for the person you're already talking to. It matches the existing action sheet, so fine to leave, but worth trimming as a follow-up.

Merging once CI is green on the new head.

jackjackbits:
Pushed 4d82b3c4: mention/DM are now hidden from the context menu inside 1:1 conversations (redundant there); public timeline unchanged. Also documented the mesh-switch window-reset behavior change in the description.

hotpixelgroup:
Thank you! Filtering the "N new" count to rendered message rows is cleaner than my raw array delta. And thanks for documenting the mesh-switch behavior change.

PR #1360, opened 2026-07-04
Make peer lists accessible and actionable; block mesh peers by stable identity

hotpixelgroup:
The mesh peer list's only actions were tap-to-DM and a favorite button, with verify hidden behind a double-tap that never fired and blocking available only via `/block`. And reachability — the app's most important fact — was encoded in unlabeled 10pt icons (at best a macOS-only tooltip; the reachability icons had none).

### Accessibility & clarity
- Both peer lists become real accessibility citizens. Mesh rows announce name, reachability, and favorite/unread/blocked state; geohash rows announce name, a "you" marker, presence (teleported / in-area), and blocked — each non-self row as one element with a button trait and custom actions for the gesture-only interactions (your own row is a plain, non-actionable element). Neither file previously had a single `.accessibility*` modifier (only `.help` tooltips).
- The teleported vs in-area pins are finally explained.
- The DM header's `.offline` state (previously `EmptyView()`) becomes a dimmed **"offline"** tag; a geohash DM — always Nostr-routed — no longer mislabels itself "offline".
- **App Info** gains a `symbols` legend defining the main status glyphs the lists and headers use; nothing defined them before.

### Actions
- Mesh rows gain a full context menu — **direct message, favorite, show fingerprint, block/unblock** — going beyond the geohash list, whose context menu only offered block/unblock. The fingerprint double-tap, previously shadowed by the single tap, is reordered so it fires.
- Mesh block/unblock now resolve by the peer's **stable Noise identity** instead of a `/block <displayName>` string, so the exact tapped row is affected and offline peers can be unblocked (with covering tests in `UnifiedPeerServiceTests`).

### Scope
- Presentation + block-resolution only; geohash blocking (already pubkey-based) is unchanged.
- Builds on iOS and macOS; `UnifiedPeerServiceTests` pass.


jackjackbits (commented):
Thanks — this is a solid change, and the important part checks out: block-by-identity is correctly keyed to the SHA-256 fingerprint of the peer's Noise static key (the same material `/block` uses), persists via the identity manager, and enforcement is unchanged — so a blocked peer can't evade it by rotating their ephemeral ID. Also confirmed the double-tap-before-single-tap gesture ordering matches the working precedent in ContentHeaderView.

I pushed a few small fixes to your branch rather than round-tripping review comments:

- acb26ba7 — `setMeshPeerBlocked` no longer silently no-ops when the identity can't be resolved (e.g. long-press-blocking an old public message from a sender who left); it now prints the same "cannot block X: not found or unable to verify identity" message the command path shows. Confirmation/failure output also routes through `addCommandOutput` (from #1363, which merged after you branched), so blocking from inside a private chat prints into that chat instead of invisibly into the public timeline. Confirmation text now matches `/block` ("blocked X. you will no longer receive messages from them").
- c56b2dd7 — removed the favorite button's unreachable `.accessibilityLabel` (the row-level `.accessibilityElement(children: .ignore)` swallows it, and the custom action covers it) and the unreachable `/unblock <name>` fallback in `ConversationUIModel.unblock` (its only caller always passes a mesh peerID).

`UnifiedPeerServiceTests` and the ChatViewModel suites pass locally, and the iOS simulator build is clean. CI has now been approved and will run on the new push — should be green before merge.

hotpixelgroup:
Appreciate the feedback! Surfacing a failure instead of a silent no-op is the right fix and routing it through addCommandOutput so it lands in the chat is a nice touch. Glad the identity-keyed blocking held up.

Issue #1088, opened 2026-04-03
Fix Linter Issues


qalandarov:
We have a few linter issues that could be "good first issues" when #1087 is merged.
They're currently disable to avoid build failures in .swiftlint.yml file.

jackjackbits:
Closing as resolved/obsolete against the current `main` and v1.7.1-era architecture. The implementation has changed materially since this report and the reported gap is now covered. If this still reproduces on a current release, please open a focused report with the app version, OS/device, exact steps, and relevant logs.

PR #1087, opened 2026-04-03
Fix trailing_whitespace violations

qalandarov:
Predominantly white-space removal with a script + a couple of manual fixes

PR #1086, opened 2026-04-03
Integrate SwiftLint + minor fixes + GH Actions

qalandarov:
This should help us have a consistent and clean codebase and helps us catch real bugs (force unwraps, implicit returns, retain cycles)

---
Also, introduced `-skipPackagePluginValidation` to avoid forcing people to install swiftlint binary just to be able to run the code so the initial pull + build "works out of the box"

PR #1361, opened 2026-07-05
Add SwiftLint as an advisory CI-only lint job (no Xcode plugin dependency)

ecgang:
### What

Adds SwiftLint to CI as an **advisory, informational-only** job — never a build breaker, never a merge gate:

- New `.swiftlint.yml` at the repo root (curated disabled-rule list).
- One new `lint` job in `.github/workflows/swift-tests.yml` that runs `swiftlint lint` in a **pinned** container (`ghcr.io/realm/swiftlint:0.65.0`) with `continue-on-error: true`.

No Xcode Build Tool Plugin. No changes to `bitchat.xcodeproj/project.pbxproj`. No `.swift` files touched. This cannot break any `xcodebuild`/`swift build` path and cannot block a merge — it only surfaces style violations in the Actions log.

### Why this shape

An earlier attempt (#1086) wired SwiftLint in as an Xcode Build Tool Plugin. As @jackjackbits noted on that PR, plugin wiring breaks the documented `xcodebuild` flow unless every invocation passes `-skipPackagePluginValidation`, and he pushed a CI-safe counter-approach on the branch `codex/swiftlint-ci-safe-followup` ("Keep SwiftLint in CI without Xcode plugin dependency") that was never opened as a PR and has since gone stale against `main`.

This PR **finishes that CI-safe approach against current `main`**: a standalone container-based lint job, zero project-file surface, so the linter is purely observational. The image tag is pinned (not `:latest`) so results are reproducible and version bumps are deliberate.

### Prior art & credit

- #1086 (@qalandarov) — original SwiftLint initiative and the disabled-rule curation this config builds on.
- #1087 (@qalandarov) — rule-set / whitespace cleanup on the `swiftlint` branch.
- The shipped `.swiftlint.yml` is @jackjackbits' variant from `codex/swiftlint-ci-safe-followup` (not @qalandarov's byte-exact text), so credit here is prose rather than `Co-authored-by:`.
- #1088 remains the open tracker for the disabled-rule backlog; the `# To be enabled as we fix the issues` block in `.swiftlint.yml` stays as the on-ramp for that work.

### Current violations on `main` (measured with SwiftLint 0.65.0, the pinned version)

The advisory job reports **109** violations today across the enabled rules — none large enough to warrant disabling:

| Rule | Count |
|---|---|
| non_optional_string_data_conversion | 45 |
| trailing_comma | 24 |
| comma | 7 |
| switch_case_alignment | 6 |
| todo | 5 |
| colon | 5 |
| redundant_string_enum_value | 4 |
| unused_optional_binding | 3 |
| trailing_newline | 3 |
| unused_closure_parameter | 2 |
| comment_spacing | 2 |
| unneeded_break_in_switch | 1 |
| static_over_final_class | 1 |
| opening_brace | 1 |

The heavier structural/style rules (function length, cyclomatic complexity, `trailing_whitespace`, etc.) stay disabled and tracked in #1088 so the advisory signal is honest rather than a flood.

---
This contribution was developed with AI assistance via [terminalhire](https://terminalhire.com). The author has reviewed the change and takes responsibility for its content.

### DIFF
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -153,10 +153,19 @@ jobs:
     name: SwiftLint (advisory)
     runs-on: ubuntu-latest
     timeout-minutes: 15
+    # This job runs a third-party container image, so give it the least
+    # privilege we can: a read-only token, and no credentials left in the
+    # checkout for the container to find.
+    permissions:
+      contents: read
     container:
-      image: ghcr.io/realm/swiftlint:0.65.0   # pinned tag; bump deliberately, never a floating tag
+      # Tag for readability, digest for immutability (tags can be repointed).
+      # Bump both together, deliberately — never a floating tag.
+      image: ghcr.io/realm/swiftlint:0.65.0@sha256:a482729f4b58741875af1566f23397f3f6db300372756fc31606d0a4527fab9e
     continue-on-error: true
     steps:
       - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
       - name: Run SwiftLint
         run: swiftlint lint --reporter github-actions-logging
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -145,3 +145,18 @@ jobs:
             ARCHS=arm64 \
             CODE_SIGNING_ALLOWED=NO \
             build
+
+  # Advisory only: SwiftLint reports style violations without ever failing the
+  # build. Runs in a pinned container (no Xcode plugin, no pbxproj changes) so
+  # it can never break the documented xcodebuild path or block a merge.
+  lint:
+    name: SwiftLint (advisory)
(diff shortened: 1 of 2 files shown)


jackjackbits:
Pushed 65244875 with two small follow-ups (maintainer edit):

- **Hardened the lint job**: since it runs a third-party container image, the job now gets `permissions: contents: read`, the checkout uses `persist-credentials: false`, and the image is pinned by digest as well as tag (tags are mutable).
- **Added an `excluded:` list to `.swiftlint.yml`** so local `swiftlint` runs in a working tree don't drown in `.build`/`DerivedData` artifacts. CI checkouts are fresh, so the advisory count is unchanged — verified locally at exactly 109 violations with SwiftLint 0.65.0, matching the table in the description.

Also approved the workflow run so CI can actually exercise the new job. Thanks for finishing this off the stale branch — the description's claims all checked out on independent verification.

PR #1375, opened 2026-07-06
Add capability negotiation bits to announce TLV

jackjackbits:
## Summary

Adds an optional **capabilities TLV (type 0x05)** to `AnnouncementPacket` so peers can advertise feature support, converting future protocol features from version-gated flag days into per-feature negotiation.

- New `PeerCapabilities` OptionSet in BitFoundation: little-endian bitfield, minimal-length encoding (grows only when high bits are assigned), unknown bits preserved on decode for forward compatibility.
- Assigned bits (reserved for in-flight feature PRs): 0=prekeys, 1=wifiBulk, 2=gateway, 3=groups, 4=board, 5=vouch, 6=meshDiagnostics.
- Peer capabilities stored in `BLEPeerRegistry` on verified announce; query via `BLEService.peerCapabilities(_:)`. Absent TLV decodes as `nil` (distinguishes legacy peers from peers advertising nothing).
- Local advertisement (`PeerCapabilities.localSupported`) is empty in this PR — each feature PR adds its bit when it ships.

## Compatibility

- The announce decoder already skips unknown TLVs, so old iOS/Android clients ignore this field; no version bump needed.
- Android parity: the Android app should add the same TLV + bit assignments to participate in negotiation; until then it's simply treated as a legacy peer.

## Testing

- BitFoundation package tests: minimal encoding, round-trip, unknown-bit and oversized-field tolerance.
- App tests: announce round-trip with capabilities, legacy decode → nil, future-bit survival.
- `xcodebuild` macOS build + PacketsTests/BLEPeerRegistryTests pass.

This is the base PR for the feature series (prekeys, vouching, groups, gateway, Wi-Fi bulk transport) — merge first.

### DIFF
--- a/bitchat/Protocols/Packets.swift
+++ b/bitchat/Protocols/Packets.swift
@@ -1,3 +1,4 @@
+import BitFoundation
 import Foundation
 
 // MARK: - Protocol TLV Packets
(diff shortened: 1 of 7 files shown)


jackjackbits:
PASS on-device. 3-node mesh (2x iPhone iOS 26.5 + Mac) formed with all gossip-sync types active (announce/message/group/fragment/file/prekey/board); no announce-parse breakage from the new capabilities TLV. Separately audited the Android Kotlin decoders — unknown TLVs/types/sync-bits are tolerated (skipped/ignored, unknown broadcast types still relayed), so this is backward-compatible. Recommend merging this first for the series.

PR #1376, opened 2026-07-06
Cashu ecash chips: detect, render, and redeem tokens + /pay command

jackjackbits:
First-class Cashu ecash token support in messages — detect, render, redeem — plus a `/pay` command. Content-level only: **no wire-protocol changes**.

### Detection
`cashuA…`/`cashuB…` tokens were already regex-detected (`MessageFormattingEngine.Patterns.cashu`) and rendered as a generic "pay via cashu" chip. This PR keeps that pipeline and makes it useful:
- `cashu:`/`cashu://` URI forms are handled (the token embedded after the scheme matches the existing word-boundary pattern; the decoder also normalizes URI-wrapped and percent-encoded input).
- `extractCashuLinks` now returns bare, deduplicated bearer strings instead of re-wrapped `cashu:` URIs, so the chip can decode what it renders.

### Chip rendering (`PaymentChipView` + new `CashuTokenDecoder`)
- **V3 (`cashuA`, base64url JSON):** decoded to `🥜 500 sat · mint.example.com`, memo shown as a caption line. Amount is the sum of all proofs across all token entries; unit defaults to `sat` per NUT-00.
- **V4 (`cashuB`, base64url CBOR):** a minimal, bounded CBOR reader (definite lengths only, depth ≤ 16, item budget, container caps) walks the NUT-00 TokenV4 map for amount/unit/mint/memo. Payloads it can't walk (indefinite lengths, exotic encodings) degrade to a generic Cashu chip without amount — that's the stated V4 limitation.
- **Adversarial input:** the decoder treats every token as attacker-controlled — size caps (60k chars), overflow-guarded summation, control-character stripping and length caps on displayed memo/unit/host, fail-closed on anything malformed. Fuzz-style tests cover truncated, garbage, huge, deeply nested, and overflow-adjacent inputs.
- Both themes look right: the chip keeps the existing `@ThemedPalette` styling; only its label content changed.

### Chip actions
- **Tap = redeem:** opens `cashu:<token>` for a wallet app; falls back to `https://redeem.cashu.me/?token=…` in the browser when nothing handles the scheme (iOS via `UIApplication.open` completion, macOS via `NSWorkspace.urlForApplication(toOpen:)`).
- **Context menu:** copy token · redeem in wallet · redeem on web.

### `/pay` command
- `/pay <token>` validates the token decodes, then sends the bare token as the message body (works in mesh DMs, geo DMs, and public channels).
- In a **public channel** it refuses without an explicit confirm — `/pay <token> public` — because tokens are bearer instruments and anyone reading the channel can redeem them. Sent via the normal outgoing path (`sendPublicMessage`, with local echo), unlike emotes which use `sendPublicRaw`.
- `/pay` with no/invalid args prints usage. Success output states the amount and repeats the bearer warning.
- **Suggestion gating:** `/pay` is suggested in mesh public and all DMs, but hidden in public geohash suggestions (same spirit as `/fav` gating) — it still executes there behind the `public` confirm.

## Security posture
- Cashu tokens are **bearer instruments**: possession = spendability, first redeemer wins. The UI says so whenever one is sent, and public sends require an explicit confirm.
- The app **never contacts a mint** and holds no keys/wallet state; it only detects, summarizes, copies, and relays token strings. Redemption is fully delegated to an external wallet (via `cashu:` URL) or the user's browser (redeem.cashu.me — the token only reaches that site if the user picks the web fallback).
- Decoding is display-only and fail-closed; malformed tokens can't crash rendering (fuzz-tested).

## Limitations
- V4/CBOR tokens using indefinite-length encoding render as a generic chip without an amount (deliberate: the reader is minimal and bounded rather than a full CBOR implementation).
- No embedded wallet, no receiving/claiming inside the app — v1 is deliberately detect/render/redeem-elsewhere.

## Testing
- `xcodebuild build` — macOS scheme, Debug: **succeeds**.
- Full `bitchatTests_macOS` suite (41 suites incl. new `CashuTokenDecoderTests` and extended `CommandProcessorTests`): **TEST SUCCEEDED**, 0 failures.
- SwiftLint clean on all touched files (0 violations).

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -15425,6 +15425,18 @@
         }
       }
     },
+    "content.commands.pay" : {
+      "comment" : "Autocomplete description for the /pay command that sends a Cashu ecash token",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "send a cashu ecash token in this chat"
+          }
+        }
+      }
+    },
     "content.commands.slap" : {
       "extractionState" : "manual",
       "localizations" : {
@@ -18469,6 +18481,18 @@
         }
       }
     },
+    "content.input.token_placeholder" : {
+      "comment" : "Placeholder shown after /pay in the command suggestion panel",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "token"
+          }
+        }
+      }
+    },
     "content.jump.new_count" : {
       "comment" : "Count of messages that arrived while scrolled up, shown in the jump-to-latest pill",
       "extractionState" : "manual",
@@ -19734,6 +19758,18 @@
         }
       }
     },
+    "content.payment.copy_token" : {
+      "comment" : "Context menu action copying a Cashu token to the pasteboard",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "copy token"
+          }
+        }
+      }
+    },
     "content.payment.lightning" : {
       "extractionState" : "manual",
       "localizations" : {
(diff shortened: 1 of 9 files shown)


jackjackbits:
**Codex P2 (strict V4 before sending payments) — fixed** in 953d24f7.

The permissive fallback turned any non-empty `cashuB…` base64 that failed CBOR parsing into a generic `TokenInfo`, so the `/pay` guard accepted base64 junk / truncated V4 tokens and relayed them with a success message.

Fix: added a `strict` flag to `CashuTokenDecoder.decode(_:strict:)`. In strict mode there is **no** permissive V4 fallback, and the token must cleanly decode to a known version **with a positive amount**, otherwise it returns nil. Rendering chips keeps the permissive path (an unknown chip is fine for display). The `/pay` SEND path now decodes with `strict: true` and surfaces `invalid cashu token — it doesn't decode to a known token with an amount, not sending it` instead of relaying.

Tests added:
- `/pay` with a truncated `cashuB` and with base64 junk under `cashuB` are both rejected (nothing sent).
- A valid V3 token and a valid definite-length V4 token still send (21 sat).
- Decoder-level strict-mode unit tests (accepts valid V3/V4, rejects unwalkable/truncated V4 and amountless tokens).

All CashuTokenDecoderTests + CommandProcessorTests pass on the macOS scheme (45 tests). SwiftLint clean on touched files.

jackjackbits:
PASS on-device. Token decode + chip render (amount/mint/memo) verified; `/pay` accepts a valid token and rejects malformed/truncated input (the strict-decode fix works); redeem hand-off opens the wallet/web.

PR #1377, opened 2026-07-06
Add mesh diagnostics: /ping, /trace, and topology map

jackjackbits:
Adds three mesh diagnostics in one PR: a `/ping` command with real round-trip measurement, a `/trace` command that estimates the path from gossiped topology, and a minimal topology map sheet.

## Wire format

Two new `MessageType` cases (0x23/0x24/0x25/0x28 remain reserved for other in-flight PRs):

- `ping = 0x26` — directed (recipientID set), unencrypted, unsigned, TTL 7
- `pong = 0x27` — same shape, addressed back to the prober

Shared 9-byte payload (`MeshPingPayload` in BitFoundation):

| bytes | field |
|---|---|
| 0–7 | random nonce (pong echoes the ping's nonce) |
| 8 | origin TTL — the TTL the packet was launched with |

Each side computes hop count as `originTTL − receivedTTL + 1` (direct link = 1 hop). Decoding tolerates trailing bytes so the format can grow. Ping/pong are unsigned; they never enter `toBinaryDataForSigning()` paths (that helper is only used when signing/verifying, which these packets skip) — verified they encode/relay like any other unsigned packet.

## Relay & abuse hardening

- Ping/pong join the deterministic directed-relay tier in `BLEReceivePipeline`/`RelayController` (always relay, 20–60 ms jitter, no TTL cap) so RTT/hop measurements reflect the real DM path, including source-routed forwarding (which also decrements TTL).
- Inbound pings are rate-limited per peer — 5 per 10 s sliding window (reuses `SyncResponseRateLimiter`) — so a directed unencrypted probe can't become an amplification primitive. A pong is only accepted if its echoed nonce matches an outstanding probe *and* the sender matches the probed peer.
- `SyncTypeFlags` maps both types to no sync bit: diagnostics never spread via gossip sync.

## Commands

- `/ping @name` — prints `pinging name…`, then `pong from name: 42 ms · 2 hops` (or `· direct (1 hop)`), or `no reply from name` after a 10 s timeout. Async output routes through the reusable `addCommandOutput` (now a `CommandContextProvider` requirement) so it lands in the conversation where the command was typed.
- `/trace @name` — BFS over the gossiped `directNeighbors` graph (reuses `MeshTopologyTracker.computeRoute`, bidirectionally-confirmed fresh claims; falls back to the live link for just-connected peers). Output is labeled as an estimate: `estimated path: you → alice → bob (2 hops)` or `no known path to bob`.
- Both are mesh-only: hidden from autocomplete in geohash contexts (like `/fav`) and rejected for Nostr/geohash identities.

## Topology map

New `MeshTopologyView` sheet: Canvas-drawn graph, self highlighted in the center, known peers on a ring, edges from gossiped neighbor claims (advisory — announces cap `directNeighbors` at 10), peer/link count summary, refresh button, localized strings, and both themes respected (monospace green under matrix, system under liquid glass). Entry point: a "network → mesh topology" row in the App Info sheet, following the existing `FeatureRow` pattern; the row only renders when a topology provider is wired (previews stay unchanged).

## Android parity

The Android app needs matching `MessageType` support (0x26/0x27, same 9-byte payload) for cross-platform ping. Degradation on old clients is graceful — verified in this codebase: `BinaryProtocol.decode` does not validate the type byte, and `handleReceivedPacket` hits the unknown-type branch (logs and ignores, no response). Old clients still *relay* unknown directed types, just under the generic broadcast clamp rather than the directed tier, so pings through old relays work but with slightly lower relay determinism.

## Tests

- BitFoundation: `MeshPingPayloadTests` — encode/decode round-trip, slice-index safety, truncation/nonce-length rejection, hop-count math (117/117 package tests pass).
- App: `MeshDiagnosticsTests` — /ping output (success, direct, timeout), mesh-only gating, /trace formats (direct, multi-hop with short-ID fallback, no path), edge normalization, layout determinism. BFS path computation was already covered by `MeshTopologyTrackerTests`.
- Full `bitchatTests_macOS` suite passes; macOS and iOS app targets build clean; SwiftLint reports no violations on changed files.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -59,6 +59,28 @@ final class AppChromeModel: ObservableObject {
         isAppInfoPresented = true
     }
 
+    /// Builds the mesh topology map model from the transport's gossiped
+    /// graph plus the live nickname table. Unknown nodes (heard about via a
+    /// neighbor claim but never announced to us) fall back to a short ID.
+    func meshTopologyDisplayModel() -> MeshTopologyDisplayModel {
+        let mesh = chatViewModel.meshService
+        guard let snapshot = mesh.currentMeshTopology() else { return .empty }
+        let nicknames = mesh.getPeerNicknames()
+
+        let nodes = snapshot.nodes.map { peerID -> MeshTopologyDisplayModel.Node in
+            let isSelf = peerID == snapshot.localPeerID
+            let label: String
+            if isSelf {
+                label = chatViewModel.nickname
+            } else {
+                label = nicknames[peerID] ?? "\(peerID.id.prefix(8))…"
+            }
+            return MeshTopologyDisplayModel.Node(id: peerID.id, label: label, isSelf: isSelf)
+        }
+        let edges = snapshot.edges.map { ($0.a.id, $0.b.id) }
+        return MeshTopologyDisplayModel(nodes: nodes, edges: edges)
+    }
+
     func triggerScreenshotPrivacyWarning() {
         showScreenshotPrivacyWarning = true
     }
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -5142,7 +5142,7 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "symbols"
+            "value" : "SYMBOLS"
           }
         }
       }
@@ -5178,7 +5178,7 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "network"
+            "value" : "NETWORK"
           }
         }
       }
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -5171,6 +5171,54 @@
         }
       }
     },
+    "app_info.network.title" : {
+      "comment" : "Section header for network diagnostics in the app info sheet",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "network"
+          }
+        }
+      }
+    },
+    "app_info.network.topology.description" : {
+      "comment" : "Row description for the mesh topology map",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "map of peers and links learned from mesh announces"
+          }
+        }
+      }
+    },
+    "app_info.network.topology.hint" : {
+      "comment" : "Accessibility hint for the mesh topology row",
+      "extractionState" : "manual",
+      "localizations" : {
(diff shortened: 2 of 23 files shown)


jackjackbits:
Pushed fd2dffdd addressing the CI failure and both Codex findings:

**CI failure — `sendImage_privateChatProcessesAndTransfersImage` (root cause: shared-filesystem race, not the diagnostics code).** `panicClearAllData` and `clearCurrentPublicTimeline` delete the real `~/Library/Application Support/files` tree in a detached utility-priority task, and the test process shares that tree. The wipe fires at a nondeterministic time after the panic/clear tests run, and on this run it landed inside the sendImage test's window between writing the processed JPEG to `files/images/outgoing` and re-reading it, so `prepareImagePacket` threw and the test burned its full 10s timeout (verified locally: a probe file planted in that directory is deleted by every full suite run). This PR's new suites merely shifted the parallel schedule enough to hit the window on CI. Both wipes are now skipped under tests via the existing `TestEnvironment.isRunningTests` pattern — which also stops unit-test runs from wiping the developer's real media files. Probe file now survives the full suite.

**Codex P1 — spoofable rate-limit key.** Ping packets are unsigned, so `packet.senderID` is attacker-controlled; keying the pong limiter on it let one connected peer rotate forged sender IDs past the 5-per-10s budget. `handleMeshPing` now takes the ingress link peer (the directly connected peer that delivered the packet, same as the courier-envelope path) and keys the limiter on that, while still replying to the claimed sender per protocol. Renamed `meshPingInboundMaxPerPeer` → `meshPingInboundMaxPerLink`. New regression test (`meshPingResponseBudget_isPerIngressLinkNotClaimedSender`) injects 10 pings with rotating forged senders over one link and asserts exactly 5 pongs — confirmed it fails (10 pongs) against the previous code.

**Codex P2 — async /ping output misrouted after a chat switch.** The origin conversation is now captured when the command is issued (`CommandOutputDestination`, via `currentCommandDestination()`) and the deferred pong/timeout output is routed to that destination instead of reading `selectedPrivateChatPeer` at callback time. A DM-issued result lands in the origin chat's history even if it's no longer selected (if the chat was cleared meanwhile, the result simply appears when it's reopened); a timeline-issued result pins to #mesh rather than whatever channel is active. Two new tests cover capture-then-switch in both directions.

Full macOS app suite: 1067 tests / 150 suites passing across repeated parallel runs; SwiftLint clean on all touched files.

jackjackbits:
App Info tweak added in this branch (d0704450): moved the NETWORK diagnostics section to sit directly under HOW TO USE and uppercased the NETWORK/SYMBOLS section headers to match the existing HOW TO USE styling. macOS build passes.

jackjackbits:
PASS on-device. `/ping` returned RTT + hop count, `/trace` showed an estimated path, and the topology map rendered all 3 nodes on the 3-node mesh (2x iPhone iOS 26.5 + Mac).

PR #1378, opened 2026-07-06
Originate v2 source routes and wire fragmentIdFilter targeted resync

jackjackbits:
## Part A — v2 source-route origination

Main already had the plumbing (`MeshTopologyTracker` BFS from announce `directNeighbors`, `applyRouteIfAvailable` in `broadcastPacket`, route-following relays, fragment planner MTU math), but origination was ungated: any packet with a recipient — including packets we were merely relaying — got a route attached whenever BFS found *any* path, with no version awareness and no failure recovery. This PR turns that into a deliberate policy.

**When a route is attached** (`BLESourceRouteOriginationPolicy`, all gates must pass):
- we authored the packet (relays no longer rewrite + re-sign someone else's packet — previously a relay would re-sign under its own key while `senderID` stayed the originator's, breaking §6 signature semantics);
- directed at a single peer (never broadcast), with TTL > 1 (link-local packets like REQUEST_SYNC never route);
- recipient **not** directly connected;
- BFS over confirmed fresh edges finds a complete path of **≤ 4 intermediate hops** where every hop **and the recipient** has been observed speaking protocol v2 (`MeshTopologyTracker.recordObservedVersion`, fed from decoded packets in the receive path; version observations expire with the same prune cadence as topology claims);
- routing to that recipient is not suppressed by a recent failure.

**Fallback behavior:** `BLESourceRouteFailureCache` — a routed send with no inbound packet authored by the recipient within 10 s counts as a route failure; subsequent directed sends to that recipient flood for 60 s before routes are attempted again. Inbound traffic confirms a pending send but deliberately does not lift an active suppression (it may have arrived via flood). No retransmission machinery added — MessageRouter keeps owning retries.

**Android/v1 parity:** routed sends only target v2-observed hops and v2-observed recipients, so v1-only peers never receive v2 frames from iOS and see unchanged flood behavior. Our own announces remain v1. Note the practical consequence: routes only form along nodes that have actually emitted v2 traffic (file transfers, routed packets, v2-originating Android builds); everything else keeps flooding exactly as before.

## Part B — REQUEST_SYNC `fragmentIdFilter` (TLV 0x06)

Previously encoded/decoded but never set and never read. Now:

- **Requester:** `BLEFragmentAssemblyBuffer` tracks per-assembly last-fragment time and reports stalled *broadcast* reassemblies (no new fragment for 5 s, re-requested at most every 10 s per stream). The BLE maintenance pass sends a `types=fragment` REQUEST_SYNC to each connected peer whose 0x06 filter names the stalled 8-byte fragment stream IDs (the ID prefixing every fragment payload — the requester cannot know the `PacketIdUtil` IDs of packets it never received, so the filter names streams, which the responder's store can match by payload prefix). Directed reassemblies are excluded: peers only archive broadcast fragments for gossip sync.
- **Responder:** when the filter is present, `GossipSyncManager` restricts the fragment diff to exactly the named streams and bypasses the since-cursor for them; the request's GCS filter still excludes pieces the requester already holds. `SyncResponseRateLimiter` (8/30 s), RSR marking, TTL 0, and REQUEST_SYNC's link-local rule (TTL 0, never relayed, suppressed by the forwarding policy) are unchanged.
- **Cap math:** at most **60 IDs** per request — each ID is 16 hex chars + 1 comma, so the largest value is 60 × 17 − 1 = **1019 bytes ≤ the 1024-byte decoder cap**; oversized 0x06 values are ignored rather than failing the whole request.

## Docs

- `docs/SOURCE_ROUTING.md`: status line corrected; new §8 documents the iOS origination policy and failure fallback.
- `docs/REQUEST_SYNC_MANAGER.md`: 0x05/0x06 moved from "future filters" to implemented, with requester/responder semantics and the cap math.

## Tests

52 tests across 6 suites (all passing), including: BFS no-path / stale-neighbor / v1-hop-blocks-route / v2-recipient-required / max-hop-cap; origination gating (relayed packet, broadcast recipient, TTL 0/1, connected recipient, suppression, empty route); failure-cache lifecycle (confirm, suppress, expire, prune); fragmentIdFilter wire round-trip, cap, malformed/oversized handling; responder honors the filter and bypasses the cursor; stalled-assembly detection/retry/completion. Full `bitchatTests_macOS` suite and BitFoundation package tests (111) pass; macOS app builds clean; SwiftLint clean on all touched files.

### DIFF
--- a/bitchat/Models/RequestSyncPacket.swift
+++ b/bitchat/Models/RequestSyncPacket.swift
@@ -1,16 +1,50 @@
+import BitFoundation
 import Foundation
 
 // REQUEST_SYNC payload TLV (type, length16, value)
 //  - 0x01: P (uint8) — Golomb-Rice parameter
 //  - 0x02: M (uint32, big-endian) — hash range (N * 2^P)
 //  - 0x03: data (opaque) — GR bitstream bytes (MSB-first)
+//  - 0x04: types (SyncTypeFlags) — packet types the filter covers
+//  - 0x05: sinceTimestamp (uint64, big-endian) — filter coverage cursor
+//  - 0x06: fragmentIdFilter (UTF-8) — comma-separated 16-hex-char (8-byte)
+//          fragment stream IDs; restricts the fragment diff to exactly those
+//          streams (targeted resync for stalled reassemblies)
 struct RequestSyncPacket {
+    /// Maximum fragment IDs one 0x06 filter may carry. Each ID encodes as
+    /// 16 hex chars plus a comma separator, so the largest encoded value is
+    /// 60 * 17 - 1 = 1019 bytes, which fits the 1024-byte decoder cap.
+    static let maxFragmentIdFilterCount = 60
+
     let p: Int
     let m: UInt32
     let data: Data
     let types: SyncTypeFlags?
     let sinceTimestamp: UInt64?
     let fragmentIdFilter: String?
+
+    /// Encodes 8-byte fragment stream IDs as the 0x06 filter string,
+    /// dropping malformed IDs and capping at `maxFragmentIdFilterCount`.
+    static func encodeFragmentIdFilter(_ fragmentIDs: [Data]) -> String? {
+        let tokens = fragmentIDs
+            .filter { $0.count == 8 }
+            .prefix(maxFragmentIdFilterCount)
+            .map { $0.hexEncodedString() }
+        guard !tokens.isEmpty else { return nil }
+        return tokens.joined(separator: ",")
+    }
+
+    /// Decodes a 0x06 filter string back into 8-byte fragment stream IDs,
+    /// ignoring malformed tokens and capping at `maxFragmentIdFilterCount`.
+    static func decodeFragmentIdFilter(_ filter: String?) -> Set<Data>? {
+        guard let filter else { return nil }
+        var ids: Set<Data> = []
+        for token in filter.split(separator: ",").prefix(maxFragmentIdFilterCount) {
+            guard token.count == 16, let id = Data(hexString: String(token)) else { continue }
+            ids.insert(id)
+        }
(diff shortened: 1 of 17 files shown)


jackjackbits:
Addressed both Codex P2 findings in 85112d8:

**1. Duplicate fragments no longer reset the stall clock** (`BLEFragmentAssemblyBuffer.append`)
`lastFragmentAt` now updates only when the fragment index was not already present. Relay duplicates of an already-held index (which bypass the packet deduplicator) previously kept an incomplete stream looking "fresh" forever, suppressing the targeted REQUEST_SYNC. New test `duplicateFragmentsDoNotResetStallClock` feeds periodic duplicates and asserts the stream still shows up in `stalledBroadcastFragmentIDs`.

**2. Rate-limit only the IDs that go on the wire** (`stalledBroadcastFragmentIDs`)
Candidates are now capped at `RequestSyncPacket.maxFragmentIdFilterCount` (the shared 60-ID wire cap, no duplicated literal) at the selection point, sorted oldest-stall-first with a deterministic tie-break, and only the selected streams get `lastResyncRequestAt` stamped. Overflow streams stay eligible and rotate in on the next pass instead of being silently suppressed for `retryAfter`. New test `overflowStalledStreamsRotateAcrossPasses` builds 70 stalled streams and verifies the first pass requests the 60 oldest and the second pass the remaining 10.

Full macOS app suite passes (1085 tests, 152 suites); SwiftLint clean on touched files.

jackjackbits:
PARTIAL PASS on-device. The `[ROUTE]` origination logic made correct decisions live — it flooded (rather than source-routed) to directly-connected peers, which is correct. Full multi-hop route origination wasn't force-tested (needs a deliberately separated A—relay—B topology); recommend exercising that before merge.

PR #853, opened 2025-10-21
Add type-aware REQUEST_SYNC sync rounds

jackjackbits:
## Summary
- add SyncTypeFlags bitfield support and include TLV 0x04 in REQUEST_SYNC packets
- track fragments and file transfers in GossipSyncManager and run typed sync rounds
- observe broadcast fragments in BLEService so they participate in sync

## Testing
- swift test --filter GossipSyncManagerTests (fails in sandbox: SwiftPM cache write permissions)

PR #1379, opened 2026-07-06
Geohash bulletin board: persistent signed notices over mesh sync

jackjackbits:
Turns ephemeral chat into a zero-infrastructure neighborhood board: persistent, signed, mesh-synced notices scoped to a geohash channel (or the mesh-local board), for campgrounds, festivals, and disaster zones.

## Wire format
- New `MessageType.boardPost = 0x23` (BitFoundation).
- Payload is a TLV (type u8, len u16 BE, value) in the `RequestSyncPacket` style, `bitchat/Protocols/BoardPackets.swift`:
  - post: kind, postID (16B random), geohash (UTF-8, empty = mesh-local board, max 12), content (1–512 bytes UTF-8), authorSigningKey (32B Ed25519), authorNickname (≤64 bytes), createdAt/expiresAt (u64 ms), flags (bit0 = urgent), signature (64B).
  - tombstone: kind, postID, authorSigningKey, deletedAt, signature.
- Signatures are Ed25519 over canonical length-prefixed bytes with domain prefixes `bitchat-board-v1` / `bitchat-board-del-v1`. Posts are **self-authenticating** — no dependency on the author's announce being present.
- Decode rejects: expiry more than 7 days past creation, expiry ≤ creation, content > 512 bytes, bad geohash charset, wrong field sizes. Unknown TLVs are skipped (forward compatible). Signatures are verified on ingest; invalid packets are dropped and **not relayed**.
- Only the author's key can delete: a tombstone must verify against its embedded key *and* that key must equal the stored post's author key.

## Retention and caps (`bitchat/Services/Board/BoardStore.swift`)
- Disk-persisted under Application Support `board/posts.json` (atomic writes, complete file protection on iOS), following the `CourierStore` pattern. The on-disk format is the raw signed packets (re-verified on load), plus each tombstone's retention deadline.
- Caps: 200 posts total, 5 per author key, oldest-evicted. Expiry sweep on every access. Tombstones are retained until the deleted post's **original expiry** so deletes keep outrunning stale copies (orphan tombstones capped at 7 days).
- Wired into `panicClearAllData()` alongside CourierStore/GossipMessageArchive.

## Sync-bitfield compatibility choice
`SyncTypeFlags` has been a **variable-length (1–8 byte) little-endian bitfield** since type-aware sync (#853), and `RequestSyncPacket.decode` accepts 1–8 bytes for TLV 0x04 while mapping unknown bits to no known type. So the board flag is simply **bit 8** — the first bit of the second byte — with no new TLV needed:
- Old clients decode a two-byte types field without error; the board bit maps to no known type, so they serve nothing for it (no amplification, preserving the #1371 hardening) and never request it.
- New clients interoperate with old one-byte encodings unchanged.
- Pinned by tests: `SyncTypeFlagsBoardTests` (unknown-bit tolerance, legacy one-byte decode, round trips).

Board posts get their own 60s gossip round with a 200-packet capacity and **multi-day retention**: rounds are served straight from the BoardStore (single retention owner), so posts sync until their own `expiresAt` instead of the 15-minute message window, and deleted/expired posts drop out of rounds immediately. The provider keeps the original signed packets, so packet IDs stay stable mesh-wide (no re-send loops).

## Relay behavior
- Board posts relay like broadcast messages (same clamp/jitter); posts with the **urgent flag get the announce-class preferred TTL cap (7 vs 6)** via a cheap TLV peek on the relay path.
- Old clients: the relay path is type-agnostic, so existing clients still **flood-relay** live board packets (with the ordinary broadcast clamp), but they do not store, serve, or backfill them via gossip sync — multi-day persistence and partition backfill require new-client paths. This is expected and degrades gracefully.

## UI
- Pin button in the header opens the board sheet for the current context: the geohash channel's board, or the mesh-local board in mesh chat.
- Urgent posts pinned on top, then newest-first; compose with urgent toggle and 1/3/7-day expiry picker; swipe-delete (plus inline delete button and VoiceOver action) on own posts sends a tombstone.
- Both themes via the existing palette; user-facing strings use `String(localized:defaultValue:comment:)`; VoiceOver labels on all tappables. No `minHeight` + `maxHeight: .infinity` header combinations.

## Nostr bridge (one-way, v1)
When a post targets a geohash, it is also published as a **kind-1 location note** through the existing geo-relay machinery (`GeoRelayDirectory` + `NostrProtocol.createGeohashTextNote` + `NostrRelayManager`) so online users see it. **Inbound Nostr→board merging is out of scope for v1** and left to v2.

## Capability bit
`PeerCapabilities` does not exist on this base branch, so the `.board` capability bit is **skipped**; follow-up once #1375 merges.

## Tests
- `BoardPacketsTests`: post/tombstone round trips, forged/tampered signature rejection, 7-day expiry and 512-byte content limits, geohash charset, unknown-TLV tolerance, truncation, urgent-flag peek.
- `BoardStoreTests`: caps + per-author/global eviction, expiry sweep, tombstone delete/propagation/forgery/race, persistence across restart, panic wipe.
- `SyncTypeFlagsBoardTests`: bitfield widening compat (see above).
- `GossipSyncBoardTests`: board rounds served from the provider, type isolation, provider-gated maintenance rounds.

## Build/test results
- `bitchat (macOS)` scheme builds; `bitchat (iOS)` builds for simulator (arm64).
- All 4 new suites pass (32 tests), plus regression sweep: GossipSyncManagerTests, RelayControllerTests, BLEReceivePipelineTests, ProtocolContractTests, PacketsTests, RequestSyncManagerTests, SyncResponseRateLimiterTests, GCSFilterTests, BLEServiceCoreTests, ViewSmokeTests, AppArchitectureTests — all green.
- SwiftLint: no violations on changed files.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -18,6 +18,9 @@ final class AppChromeModel: ObservableObject {
     private let chatViewModel: ChatViewModel
     private var cancellables = Set<AnyCancellable>()
 
+    /// Bulletin-board coordinator, created on first use of the board sheet.
+    private(set) lazy var boardManager = BoardManager(transport: chatViewModel.meshService)
+
     init(chatViewModel: ChatViewModel, privateInboxModel: PrivateInboxModel) {
         self.chatViewModel = chatViewModel
         self.nickname = chatViewModel.nickname
--- a/bitchat/Protocols/BoardPackets.swift
+++ b/bitchat/Protocols/BoardPackets.swift
@@ -1,0 +1,348 @@
+//
+// BoardPackets.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import CryptoKit
+import Foundation
+
+// MARK: - Board wire format (MessageType.boardPost payloads)
+//
+// TLV layout (type u8, length u16 big-endian, value), matching REQUEST_SYNC:
+//  - 0x01: kind (u8) — 0x01 post, 0x02 tombstone
+//  - 0x02: postID (16B random)
+//  - 0x03: geohash (UTF-8, empty = mesh-local board, max 12 chars)
+//  - 0x04: content (UTF-8, 1...512 bytes) [post]
+//  - 0x05: authorSigningKey (32B Ed25519 public key)
+//  - 0x06: authorNickname (UTF-8, max 64 bytes)
+//  - 0x07: createdAt (u64 big-endian, ms) [post]
+//  - 0x08: expiresAt (u64 big-endian, ms, max 7 days after createdAt) [post]
+//  - 0x09: flags (u8, bit0 = urgent) [post]
+//  - 0x0A: signature (64B Ed25519)
+//  - 0x0B: deletedAt (u64 big-endian, ms) [tombstone]
+// Unknown TLVs are skipped for forward compatibility.
+
+enum BoardWireConstants {
+    static let postIDLength = 16
+    static let signingKeyLength = 32
+    static let signatureLength = 64
+    static let contentMaxBytes = 512
+    static let nicknameMaxBytes = 64
+    static let geohashMaxLength = 12
+    /// Posts may live at most 7 days past their creation timestamp.
+    static let maxLifetimeMs: UInt64 = 7 * 24 * 60 * 60 * 1000
+    static let postSigningContext = "bitchat-board-v1"
+    static let tombstoneSigningContext = "bitchat-board-del-v1"
+    static let geohashAlphabet = Set("0123456789bcdefghjkmnpqrstuvwxyz")
+}
+
+private enum BoardTLVType: UInt8 {
+    case kind = 0x01
+    case postID = 0x02
+    case geohash = 0x03
+    case content = 0x04
+    case authorSigningKey = 0x05
+    case authorNickname = 0x06
+    case createdAt = 0x07
+    case expiresAt = 0x08
+    case flags = 0x09
+    case signature = 0x0A
+    case deletedAt = 0x0B
+}
+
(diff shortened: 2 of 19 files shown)


jackjackbits:
Pushed 6dd8dd0 addressing both Codex review findings:

**P1 — unbounded orphan tombstones** (`BoardStore.ingestTombstoneLocked`)
- Retention for a tombstone with no matching post is now clamped by receive time as well as the claimed `deletedAt`: `min(deletedAt + 7d, now + 7d + 1h skew)`. Since no post can outlive 7 days from receipt, a far-future `deletedAt` can no longer pin a tombstone.
- Orphan tombstones are now capped like posts: 100 globally, 5 per author signing key, oldest-received evicted first. Matched tombstones (post was present) and disk restores are exempt and keep their existing behavior, so legitimate delete-then-sync flows are unchanged.

**P2 — future-dated posts** (`BoardStore.ingestPostLocked`)
- The ingest chokepoint (which covers radio receive, gossip sync, and disk restore — all funnel through `BoardStore.ingest`) now rejects `createdAt > now + 1h skew` and `expiresAt > now + 7d + 1h skew`, so forged timestamps can't sort ahead of honest posts or hold the 200 global slots without pruning. The decoder's createdAt→expiresAt span rule is unchanged; the receive-time policy lives in one place.

**Tests**: added coverage for the skew boundary (accept within, reject beyond), far-future `expiresAt`, receive-time clamping of orphan retention, per-author and global orphan caps with eviction (evicted tombstone's post becomes ingestable again, survivors still suppress), and matched-tombstone exemption from the caps. Full suite green locally: 1092 tests in 153 suites, plus SwiftLint clean on touched files.

jackjackbits:
PASS on-device. Posting, urgent-pin, cross-device propagation, and delete (signed tombstone) all verified across 3 devices.

PR #1380, opened 2026-07-06
Transitive verification: vouch for verified peers over Noise

jackjackbits:
Serverless web-of-trust: peers you verified can vouch for the peers *they* verified, giving those strangers a distinct **vouched** tier — "verified by people you verified" — with no servers, no directory, and no extra round trips. Based on #1375 (capability bits).

## Attestation format

`NoisePayloadType.vouch = 0x12` (0x04–0x0F left reserved for in-flight PRs) carries a batch: `[count u8]` then per attestation `[len u16 BE][TLV]`, capped at 16 both ways (the sender-declared count is not trusted).

Each attestation (TLV, unknown types skipped for forward compat):

| TLV | Field | Size |
|---|---|---|
| 0x01 | voucheeFingerprint (SHA-256 of Noise static key) | 32 B |
| 0x02 | voucheeSigningKey (Ed25519, anchors the vouch to a concrete identity) | 32 B |
| 0x03 | timestamp (ms since 1970, BE) | 8 B |
| 0x04 | Ed25519 signature by the **voucher's** signing key | 64 B |

Signature covers `"bitchat-vouch-v1" | voucheeFingerprint | voucheeSigningKey | timestamp`. The voucher's identity is deliberately implicit: attestations only arrive inside an authenticated Noise session, so receivers verify against the session peer's announce-bound signing key and store the vouch keyed by that peer's fingerprint — a relayed/captured attestation is useless to anyone whose key doesn't match. Attestations expire receiver-side after 30 days (plus a 1h future-skew bound).

## Exchange policy (automatic, bounded)

On `onPeerAuthenticated` (same additive hook family the verification coordinator uses — `BLEService.addPeerAuthenticatedObserver` appends to `NoiseEncryptionService`'s handler array without touching the single-slot handshake callback), `ChatVouchCoordinator` sends a batch iff:

- the peer's fingerprint is in **my** `verifiedFingerprints`, and
- the peer advertises the `.vouch` capability bit (now in `PeerCapabilities.localSupported`), and
- no batch went to that fingerprint in the last 24 h (persisted in the identity cache; stamped only after an actual send).

Batch contents: up to 16 *other* verified fingerprints, most recently verified first (new `verifiedAt` map; pre-existing verifications sort last), skipping fingerprints whose announce-bound signing key we don't know this session (nothing to anchor).

## Accept policy

Inbound `vouch` payloads are processed only when the **sending** peer is verified-by-me; each attestation must verify under the sender's announce-bound Ed25519 key and be within the validity window. Storage (`SecureIdentityStateManager.recordVouch`) additionally drops self-vouches, vouches for peers I already verified, and enforces a cap of 8 vouchers per vouchee (most recent kept). Vouch payloads arriving via the Nostr path are ignored — the trust anchoring is defined against the mesh Noise session.

## Trust tier + persistence compatibility

`TrustLevel` gains `vouched` between `casual` and `trusted`. Two compat facts, verified by tests:

- `TrustLevel` persists by **String raw value**, not ordinal, so inserting a case mid-ladder cannot corrupt stored values — every existing case keeps its exact raw string.
- `vouched` is **derived, never persisted**: `effectiveTrustLevel(for:)` recomputes it on read (≥1 vouch whose voucher is *currently* in `verifiedFingerprints` and unexpired). Unverifying a voucher therefore silently retires their vouches with no cascade delete, and re-verifying restores them.
- The three new `IdentityCache` fields (`vouchesByVouchee`, `vouchBatchSentAt`, `verifiedAt`) are Optionals so pre-existing encrypted caches decode via `decodeIfPresent` instead of tripping the discard-unreadable-cache recovery path.

Panic wipe: vouch state lives inside `IdentityCache`, so `clearAllIdentityData()` wipes it with the rest (covered by test).

## UI

- **Peer list**: unfilled `checkmark.seal` badge for vouched peers (connected or offline); the filled seal stays exclusive to verified. Tooltip + VoiceOver state label ("vouched") added to the row's spoken description. Colors ride the row's palette color, so both themes work unchanged.
- **FingerprintView**: when not verified but vouched, a teal `checkmark.seal` + "✓ VOUCHED" badge with "vouched for by N people you verified" (proper plural via xcstrings substitution) and the voucher nicknames where known, shown regardless of session state.
- Four new `Localizable.xcstrings` keys (en), following the manual-extraction pattern.

## Android parity note

Android needs: the `.vouch` capability bit (1 << 5), payload type 0x12, the TLV/batch format and signing context above, the same accept gates (verified sender, announce-bound key, 30-day expiry, self-vouch/already-verified drops, 8-voucher cap), the 24h/16-entry exchange bounds, and an unfilled-seal tier below explicit verification. Until then Android peers simply don't advertise `.vouch` and neither send nor receive vouches — no interop risk.

## Tests

33 new tests across three suites (Swift Testing):
- `VouchAttestationTests`: encode/decode round-trips, unknown-TLV skip, truncation/size rejection, signature valid/forged/tampered, expiry window, batch cap both directions, malformed-entry tolerance.
- `SecureIdentityStateManagerVouchTests`: accept gates (unverified voucher, self-vouch, verified vouchee, stale/future timestamps), 8-voucher cap semantics, voucher-unverified invalidation + restore, read-time expiry, trust ladder derivation, recent-verified ordering, batch rate-limit persistence, panic wipe, `TrustLevel` raw-value stability, legacy-cache decode, vouch-state reload.
- `ChatVouchCoordinatorContextTests`: exchange policy (verified+capable gate, 24h rate limit, cap/exclusion, signing-key skip, no rate-stamp on empty batch) and accept policy (unverified/unknown sender, forged signature, expired attestation, missing sender key) against a mock context with real Ed25519 keys.

Build: `bitchat (macOS)` Debug builds clean (`-derivedDataPath /tmp/dd-vouch`). Full macOS test bundle: **1088 tests in 152 suites, 0 failures**. SwiftLint clean on all touched files.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -18,6 +18,9 @@ final class AppChromeModel: ObservableObject {
     private let chatViewModel: ChatViewModel
     private var cancellables = Set<AnyCancellable>()
 
+    /// Bulletin-board coordinator, created on first use of the board sheet.
+    private(set) lazy var boardManager = BoardManager(transport: chatViewModel.meshService)
+
     init(chatViewModel: ChatViewModel, privateInboxModel: PrivateInboxModel) {
         self.chatViewModel = chatViewModel
         self.nickname = chatViewModel.nickname
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -14,6 +14,9 @@ struct MeshPeerRow: Identifiable, Equatable {
     let isMutualFavorite: Bool
     let encryptionStatus: EncryptionStatus
     let showsVerifiedBadgeWhenOffline: Bool
+    /// Vouched-for by someone I verified, without an explicit verification of
+    /// mine — rendered as the unfilled seal (verified gets the filled one).
+    let showsVouchedBadge: Bool
 
     var id: String { peerID.id }
 }
@@ -183,13 +186,12 @@ final class PeerListModel: ObservableObject {
         let myPeerID = chatViewModel.meshService.myPeerID
         let meshRows = allPeers.map { peer in
             let isMe = peer.peerID == myPeerID
-            let verifiedBadge: Bool
-            if !isMe && !peer.isConnected,
-               let fingerprint = chatViewModel.getFingerprint(for: peer.peerID) {
-                verifiedBadge = peerIdentityStore.isVerified(fingerprint)
-            } else {
-                verifiedBadge = false
-            }
+            let fingerprint = isMe ? nil : chatViewModel.getFingerprint(for: peer.peerID)
+            let isVerifiedFingerprint = fingerprint.map { peerIdentityStore.isVerified($0) } ?? false
+            let verifiedBadge = !peer.isConnected && isVerifiedFingerprint
+            // Vouched is subordinate to verified: never show both seals.
+            let vouchedBadge = !isVerifiedFingerprint
+                && (fingerprint.map { chatViewModel.isVouchedFingerprint($0) } ?? false)
 
             return MeshPeerRow(
                 peerID: peer.peerID,
@@ -202,7 +204,8 @@ final class PeerListModel: ObservableObject {
                 isReachable: peer.isReachable,
                 isMutualFavorite: peer.isMutualFavorite,
                 encryptionStatus: chatViewModel.getEncryptionStatus(for: peer.peerID),
-                showsVerifiedBadgeWhenOffline: verifiedBadge
+                showsVerifiedBadgeWhenOffline: verifiedBadge,
+                showsVouchedBadge: vouchedBadge
             )
         }
 
--- a/bitchat/App/VerificationModel.swift
+++ b/bitchat/App/VerificationModel.swift
@@ -150,6 +150,17 @@ final class VerificationModel: ObservableObject {
                 self?.objectWillChange.send()
             }
             .store(in: &cancellables)
+
+        // Vouch state changes (ChatVouchCoordinator.notifyPeerTrustChanged)
+        // are signalled via this notification rather than a published
+        // property, so an open fingerprint sheet refreshes its vouched badge
+        // live when a vouch batch is accepted.
+        NotificationCenter.default.publisher(for: Notification.Name("peerStatusUpdated"))
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.objectWillChange.send()
+            }
+            .store(in: &cancellables)
     }
 
     private func resolveDisplayName(for peerID: PeerID, statusPeerID: PeerID) -> String {
--- a/bitchat/App/VerificationModel.swift
+++ b/bitchat/App/VerificationModel.swift
@@ -9,6 +9,14 @@ struct FingerprintPresentationState: Equatable {
     let theirFingerprint: String?
     let myFingerprint: String
     let isVerified: Bool
+    /// Number of currently-valid vouches from peers the user verified
+    /// (0 when the peer is explicitly verified — the stronger badge wins).
+    let voucherCount: Int
+    /// Display names of the (verified) vouchers, where known.
+    let voucherNames: [String]
+
+    /// Vouched for by ≥1 peer the user verified (and not explicitly verified).
+    var isVouched: Bool { voucherCount > 0 }
 
     var canToggleVerification: Bool {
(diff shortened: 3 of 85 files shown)


jackjackbits:
Pushed 48035872 + 385e9e2a addressing the CI failure and the Codex P2 finding.

**CI root cause — "Run Swift Tests (app)" (exit 137):** not a failing assertion but a process-wide deadlock. The run's watchdog sample shows all three Swift Concurrency cooperative-pool threads parked in `SecureIdentityStateManager` syncs (`recordVouch` → `queue.sync(.barrier)` at SecureIdentityStateManager.swift:621, `effectiveTrustLevel` → `queue.sync`) while the main thread and the one dispatch worker sat idle. `SecureIdentityStateManagerVouchTests` was a nonisolated Swift Testing suite, so its tests ran in parallel on the cooperative pool; each one enqueues a `queue.async(.barrier)` write (`setVerified`) and immediately blocks in a `queue.sync` behind it. On the few-core CI runner that saturated every cooperative-pool thread, the pending barriers never got a dispatch worker, and the suite hung until the workflow watchdog SIGKILLed the test process (`Killed: 9` → exit 137). Locally the wider pool masks it — which is why the full suite ran green on the author's machine.

**Fix 1 (48035872):** made the suite `@MainActor` (with a comment explaining why). That matches production isolation — the vouch API is driven by the main-actor `ChatVouchCoordinator` — and keeps the blocking `queue.sync` calls off the cooperative pool, the same profile as every other green test that exercises this class.

**Fix 2 (Codex P2, stale fingerprint sheet, 48035872):** `VerificationModel.bind()` now subscribes to the `"peerStatusUpdated"` notification that `ChatVouchCoordinator.notifyPeerTrustChanged()` posts (the same signal `PeerListModel` refreshes on) and forwards it to `objectWillChange`, so an open fingerprint sheet re-renders its vouched badge live when a vouch batch is accepted. Added a regression test in `AppArchitectureTests` that pins VerificationModel's own subscription — it releases the `PrivateConversationModel` first, because that model's re-assignment of its published selection on the same notification was incidentally masking the gap. Verified the test fails without the fix and passes with it.

**Fix 3 (385e9e2a, latent CI flake):** ported the shared-filesystem race fix from #1377: `panicClearAllData` / `clearCurrentPublicTimeline` delete the real `~/Library/Application Support/files` tree in detached utility-priority tasks, and `ChatViewModelTests` invoke both — under parallel scheduling the wipe can land mid-test and delete media a concurrent test just wrote. Both wipes now `guard !TestEnvironment.isRunningTests` (existing pattern). This wasn't the mechanism of this run's hang (the thread sample pins that on the vouch-suite deadlock) but it's the same class of nondeterministic CI failure and needs a per-branch copy until one lands on main.

**Verification:** vouch suites (`SecureIdentityStateManagerVouchTests`, `ChatVouchCoordinatorContextTests`, `VouchAttestationTests`) and the full macOS app suite pass locally (1089 tests / 152 suites, plus the XCTest half, via the same `swift test --parallel` invocation CI uses). SwiftLint clean on touched files.

jackjackbits:
Ported the vouch capability-race fix onto this branch as `dcc05f3f` (cherry-pick of `b8adcbe9` from the integration test branch — clean, no conflicts).

**What it fixes:** vouch attestations were only sent from `peerAuthenticated`, gated on the peer's `.vouch` capability. That capability arrives on the peer's *announce*, processed independently of the Noise handshake, so at auth time the set was usually still empty → gate failed → the vouch was silently skipped and never retried.

**The fix adds two more triggers plus a race-tolerant gate:**
- `peersUpdated` — fires after every verified announce (peer-list update), so the batch goes out once the `.vouch` bit actually arrives.
- `vouchToConnectedVerifiedPeers` — runs on local verification (manual + QR), covering verify-while-already-connected and propagating the new identity onward.
- Capability gate relaxed: an empty/unknown capability set is now eligible (the Noise `0x12` payload is harmlessly ignored by non-supporting peers); only a non-empty set that explicitly lacks `.vouch` is skipped.

**On-device confirmed:** the transitive vouch propagated as expected once the send was triggered on verify / announce arrival rather than auth alone.

**Validation on this branch:**
- `bitchat (macOS)` build: SUCCEEDED
- `bitchat (iOS)` generic sim arm64 build: SUCCEEDED
- Verified the new code is actually compiled in (not a stale build): `verify-triggered vouch pass sent to` present in the fresh `bitchat.debug.dylib`.
- Tests: `ChatVouchCoordinatorContextTests`, `ChatVerificationCoordinatorContextTests`, `SecureIdentityStateManagerVouchTests` — 35 tests, all passing.

PR #1381, opened 2026-07-06
Prekey bundles: forward-secret async first contact for courier mail

jackjackbits:
Closes the biggest gap in the courier protocol: Noise XX is interactive-only, and store-and-forward envelopes fell back to one-way Noise X against the recipient's **long-lived static key** — no forward secrecy, so a later compromise of that key exposed every envelope ever captured in transit. This PR adds one-time prekey bundles so async first contact is forward secret.

Based on #1375 (capability bits); retargets to main when it merges.

## Crypto design

**Bundle** (`PrekeyBundle`, MessageType `0x24`): 8 one-time Curve25519 public prekeys `[(prekeyID u32, pubkey 32B)]` + owner's Noise static key + `generatedAt` (ms), signed with the owner's Ed25519 key over domain-prefixed canonical bytes (`"bitchat-prekey-bundle-v1"` context, owner key, count, each id‖key pair, generatedAt). On ingest the signature is verified against the **announce-bound** signing key for that noise key (live registry entry with matching 32-byte key, else the identity persisted from a verified announce); unverifiable bundles are dropped and never enter our gossip store, so nodes only help spread bundles they can attribute.

**Sealing** (`sealPrekeyPayload`/`openPrekeyPayload`): Noise **X** where the responder static is the *one-time prekey*, not the identity key, with prologue `"bitchat-prekey-v1" ‖ prekeyID (u32 BE)` — domain-separated from interactive XX, from static-sealed courier mail (`"bitchat-courier-v1"`), and from every *other* prekey (a ciphertext can't be replayed against a different ID). The initiator's static rides encrypted inside and is bound by the `ss` DH, so the recipient authenticates the sender exactly as with v1 envelopes — including the blocked-sender fingerprint check in `openCourierEnvelope`, which both paths share. Once the recipient consumes the prekey and its grace window lapses, the private key is deleted: captured ciphertext becomes undecryptable even if the identity static leaks later.

## Gossip / sync integration

- Bundles broadcast as signed packets and ride their own REQUEST_SYNC round (60s cadence, 200-owner cap, 24h freshness) — `SyncTypeFlags` bit 9 (`prekeyBundle`); bit 8 stays reserved for the board PR. No wire-format widening was needed: the flags field is already a 1–8-byte little-endian u64 and `RequestSyncPacket`'s decoder tolerates any length, so old clients decode the wider flags and simply never match the new bit; unknown bits are ignored on both sides.
- Latest bundle per owner wins (`generatedAt`); like announces, bundles are exempt from the since-cursor (at most one per owner) and deliberately survive leave/stale-peer cleanup — they exist precisely to reach owners who are gone.
- Our own bundle re-broadcasts alongside announces at most hourly, and immediately when it changes.

## Consumption + 48h grace tradeoff

Local privates persist in the Keychain. Opening marks the prekey consumed, but spray-and-wait means the *same* ciphertext (or a re-seal of the same message to the same ID) can arrive via several couriers days apart — and a redelivery is indistinguishable from a fresh ciphertext to the same prekey. Simplest sound policy, implemented here: **retain the consumed private for 48h, then delete**. Tradeoff: the forward-secrecy clock starts at deletion, not first open; the window is short and fixed. When the unconsumed count drops below 3 the batch tops back up (surviving unconsumed keys stay valid) and the refreshed bundle re-gossips. Prekey privates are wiped in panic mode along with cached peers' bundles.

Sender side, one message consumes exactly **one** prekey: deposit retries and extra couriers for the same message ID reuse its assignment (persisted, so restarts can't double-spend); distinct messages get distinct prekeys; when a bundle is exhausted or older than 7 days, sealing falls back to static.

## Envelope v2 compat

`CourierEnvelope` gains an optional `prekeyID` TLV (0x05) as the format discriminator; absent = v1 static-sealed, and v1 envelopes encode byte-identically to before. Old clients skip the unknown TLV (verified: the decoder `continue`s on unknown types), carry and hand over v2 envelopes opaquely, and if one is addressed to them the static-key open fails and is dropped quietly — same handling as a tag collision today. Routing gates on the recipient's advertised `.prekeys` capability as a veto for on-mesh peers, but the real signal is a verified cached bundle.

## Tests

38 new tests, and the full suite (1094 tests / 155 suites) plus BitFoundation package tests (113) pass on `bitchat (macOS)`:
- bundle sign/verify (valid, forged signature, tampered contents, wrong signing key), TLV round-trip, unknown-TLV skip, duplicate-ID rejection
- prekey seal/open round-trip, wrong-ID/wrong-recipient/tampered failures, prologue domain separation vs static seals, unlinkability
- consumed-prekey redelivery grace (crypto-level and end-to-end via a second courier packet), replenishment trigger + generation bump, keychain persistence, wipe
- bundle store: latest-wins ingest, per-message assignment reuse, exhaustion → static fallback, top-up consumption carry-over, 7-day expiry, peer cap, disk persistence
- envelope v2 compat (v1 byte-identical, v1 still opens end-to-end), and the full three-node courier flow with a prekey-sealed envelope, plus rejection of unverifiable/forged gossiped bundles

## Android parity

Android needs: MessageType `0x24`, SyncTypeFlags bit 9, the two prologues and the `"bitchat-prekey-bundle-v1"` canonical signing bytes, the envelope `prekeyID` TLV 0x05, and the same 48h-grace / replenish-below-3 / one-prekey-per-message policies. Until then Android peers simply don't advertise `.prekeys` and never gossip bundles, so iOS senders fall back to static sealing — no interop break.

## Scope exclusions

QR-embedded prekeys and DM-anyone-from-/who UX are follow-ups; this PR is the crypto + gossip + courier integration.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -18,6 +18,9 @@ final class AppChromeModel: ObservableObject {
     private let chatViewModel: ChatViewModel
     private var cancellables = Set<AnyCancellable>()
 
+    /// Bulletin-board coordinator, created on first use of the board sheet.
+    private(set) lazy var boardManager = BoardManager(transport: chatViewModel.meshService)
+
     init(chatViewModel: ChatViewModel, privateInboxModel: PrivateInboxModel) {
         self.chatViewModel = chatViewModel
         self.nickname = chatViewModel.nickname
@@ -57,6 +60,28 @@ final class AppChromeModel: ObservableObject {
 
     func presentAppInfo() {
         isAppInfoPresented = true
+    }
+
+    /// Builds the mesh topology map model from the transport's gossiped
+    /// graph plus the live nickname table. Unknown nodes (heard about via a
+    /// neighbor claim but never announced to us) fall back to a short ID.
+    func meshTopologyDisplayModel() -> MeshTopologyDisplayModel {
+        let mesh = chatViewModel.meshService
+        guard let snapshot = mesh.currentMeshTopology() else { return .empty }
+        let nicknames = mesh.getPeerNicknames()
+
+        let nodes = snapshot.nodes.map { peerID -> MeshTopologyDisplayModel.Node in
+            let isSelf = peerID == snapshot.localPeerID
+            let label: String
+            if isSelf {
+                label = chatViewModel.nickname
+            } else {
+                label = nicknames[peerID] ?? "\(peerID.id.prefix(8))…"
+            }
+            return MeshTopologyDisplayModel.Node(id: peerID.id, label: label, isSelf: isSelf)
+        }
+        let edges = snapshot.edges.map { ($0.a.id, $0.b.id) }
+        return MeshTopologyDisplayModel(nodes: nodes, edges: edges)
     }
 
     func triggerScreenshotPrivacyWarning() {
--- a/bitchat/App/LocationChannelsModel.swift
+++ b/bitchat/App/LocationChannelsModel.swift
@@ -12,20 +12,26 @@ final class LocationChannelsModel: ObservableObject {
     @Published private(set) var bookmarkNames: [String: String]
     @Published private(set) var locationNames: [GeohashChannelLevel: String]
     @Published private(set) var userTorEnabled: Bool
+    @Published private(set) var gatewayEnabled: Bool
 
     private let manager: LocationChannelManager
     private let network: NetworkActivationService
+    private let gateway: GatewayService
     private var cancellables = Set<AnyCancellable>()
 
     init(
         manager: LocationChannelManager? = nil,
-        network: NetworkActivationService? = nil
+        network: NetworkActivationService? = nil,
+        gateway: GatewayService? = nil
     ) {
         let manager = manager ?? .shared
         let network = network ?? .shared
+        let gateway = gateway ?? .shared
 
         self.manager = manager
         self.network = network
+        self.gateway = gateway
+        self.gatewayEnabled = gateway.isEnabled
         self.permissionState = manager.permissionState
         self.availableChannels = manager.availableChannels
         self.selectedChannel = manager.selectedChannel
@@ -96,6 +102,10 @@ final class LocationChannelsModel: ObservableObject {
         network.setUserTorEnabled(enabled)
     }
 
+    func setGatewayEnabled(_ enabled: Bool) {
+        gateway.setEnabled(enabled)
+    }
+
     func refreshMeshChannelsIfNeeded() {
         guard case .mesh = selectedChannel,
               permissionState == .authorized,
@@ -160,6 +170,10 @@ final class LocationChannelsModel: ObservableObject {
         network.$userTorEnabled
             .receive(on: DispatchQueue.main)
             .assign(to: &$userTorEnabled)
+
+        gateway.$isEnabled
+            .receive(on: DispatchQueue.main)
+            .assign(to: &$gatewayEnabled)
     }
 
     private func level(forLength length: Int) -> GeohashChannelLevel {
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -14,6 +14,9 @@ struct MeshPeerRow: Identifiable, Equatable {
     let isMutualFavorite: Bool
     let encryptionStatus: EncryptionStatus
     let showsVerifiedBadgeWhenOffline: Bool
+    /// Vouched-for by someone I verified, without an explicit verification of
+    /// mine — rendered as the unfilled seal (verified gets the filled one).
+    let showsVouchedBadge: Bool
 
     var id: String { peerID.id }
 }
@@ -183,13 +186,12 @@ final class PeerListModel: ObservableObject {
         let myPeerID = chatViewModel.meshService.myPeerID
         let meshRows = allPeers.map { peer in
             let isMe = peer.peerID == myPeerID
-            let verifiedBadge: Bool
-            if !isMe && !peer.isConnected,
-               let fingerprint = chatViewModel.getFingerprint(for: peer.peerID) {
-                verifiedBadge = peerIdentityStore.isVerified(fingerprint)
-            } else {
-                verifiedBadge = false
-            }
+            let fingerprint = isMe ? nil : chatViewModel.getFingerprint(for: peer.peerID)
+            let isVerifiedFingerprint = fingerprint.map { peerIdentityStore.isVerified($0) } ?? false
+            let verifiedBadge = !peer.isConnected && isVerifiedFingerprint
+            // Vouched is subordinate to verified: never show both seals.
+            let vouchedBadge = !isVerifiedFingerprint
+                && (fingerprint.map { chatViewModel.isVouchedFingerprint($0) } ?? false)
 
             return MeshPeerRow(
                 peerID: peer.peerID,
@@ -202,7 +204,8 @@ final class PeerListModel: ObservableObject {
                 isReachable: peer.isReachable,
                 isMutualFavorite: peer.isMutualFavorite,
                 encryptionStatus: chatViewModel.getEncryptionStatus(for: peer.peerID),
-                showsVerifiedBadgeWhenOffline: verifiedBadge
+                showsVerifiedBadgeWhenOffline: verifiedBadge,
+                showsVouchedBadge: vouchedBadge
             )
         }
 
--- a/bitchat/App/VerificationModel.swift
+++ b/bitchat/App/VerificationModel.swift
@@ -9,6 +9,14 @@ struct FingerprintPresentationState: Equatable {
     let theirFingerprint: String?
     let myFingerprint: String
     let isVerified: Bool
+    /// Number of currently-valid vouches from peers the user verified
+    /// (0 when the peer is explicitly verified — the stronger badge wins).
+    let voucherCount: Int
+    /// Display names of the (verified) vouchers, where known.
+    let voucherNames: [String]
+
+    /// Vouched for by ≥1 peer the user verified (and not explicitly verified).
+    var isVouched: Bool { voucherCount > 0 }
 
     var canToggleVerification: Bool {
         encryptionStatus == .noiseSecured || encryptionStatus == .noiseVerified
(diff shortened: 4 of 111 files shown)


jackjackbits (commented):
Adversarial review of the prekey-bundle crypto/gossip/courier integration. I read the full diff plus the base-branch announce/identity trust model and `main`'s `CourierEnvelope` decoder. Three findings below, severity-ordered, followed by what I probed and ruled out.

## Findings

### 1. HIGH — Bundle authentication inherits the announce trust model, which never proves possession of the Noise static key. An active attacker can publish a verifiable bundle whose prekeys they control for a victim's noise key, then decrypt/redirect courier mail addressed to that victim.

`BLEService.handlePrekeyBundle` verifies the bundle's inner Ed25519 signature against `announceBoundSigningKey(forNoiseKey: bundle.noiseStaticPublicKey)` (BLEService.swift, `announceBoundSigningKey`). That binding comes from a "verified" announce, but announce verification does **not** prove the announcer holds the noise static private key:

- `BLEAnnouncePreflightPolicy.evaluate` (BLEAnnounceHandlingPolicy.swift:32-35) derives `peerID = PeerID(publicKey: announcement.noisePublicKey)` — the noise key is public info, so any sender can set `senderID` to match.
- `BLEAnnounceTrustPolicy.evaluate` (BLEAnnounceHandlingPolicy.swift:70-90) only checks that the packet is signed by the **self-asserted** `signingPublicKey`; it never checks that key against an authority. The `keyMismatch` guard compares `existingNoisePublicKey` (keyed by `PeerID(noiseKey)`, so it always equals the announced noise key) — it never compares the signing key.
- `SecureIdentityStateManager.upsertCryptographicIdentity` (SecureIdentityStateManager.swift:320) overwrites the stored `signingPublicKey` for a noise-key fingerprint on every verified announce (last-writer-wins).

Attack (victim V offline — the exact courier/prekey use case): attacker broadcasts an announce `{noisePublicKey: Kv, signingPublicKey: attackerSign, senderID: PeerID(Kv)}` signed by `attackerSign`. It is accepted as verified and persisted, binding `Kv → attackerSign`. The attacker then gossips a `PrekeyBundle{noiseStaticPublicKey: Kv, prekeys: attacker-owned, signature by attackerSign}`. `signableBytes()` covers `Kv` and the prekeys, so the signature is valid under the poisoned binding and `verifyPrekeyBundleSignature` passes. Alice caches it, and `sendCourierMessage`/`assignRecipientPrekey` seals her mail to the attacker's prekey. The `recipientTag` is derived from `Kv` (looks addressed to V), but the ciphertext is decryptable only by the attacker, who also learns Alice's authenticated static identity. V, if it ever receives the envelope, cannot open it.

This is partly a property of the pre-existing announce model, but the PR's confidentiality guarantee rests entirely on the announce-bound signing key, so it is squarely in scope. Note the fallback is safe *only* when no binding exists (`announceBoundSigningKey` returns nil → static seal to the real key); the attack works precisely because the attacker supplies the binding. Consider requiring proof-of-possession of the noise key before a signing-key binding is trusted for bundle attribution, or refusing to prekey-seal to a binding that wasn't established through an authenticated (handshake/QR) channel.

### 2. MEDIUM — Gossip prekey-bundle store is keyed by the unauthenticated `packet.senderID`, so an on-mesh attacker can spray one valid bundle under up to 200 fake sender IDs and fill/exhaust the per-node cache, blocking legitimate bundles.

`GossipSyncManager.onPublicPacketSeen` `case .prekeyBundle` (GossipSyncManager.swift, added block ~line 246) stores under `PeerID(hexData: packet.senderID)`. Nothing verifies the outer packet signature or that `senderID` matches the bundle owner — `handleReceivedPacket` has no generic outer-signature gate (each handler self-verifies, and `handlePrekeyBundle` only checks the *inner* bundle signature, which is keyed on `bundle.noiseStaticPublicKey`, independent of `senderID`). The dedup key `"\(senderID)-\(timestamp)-\(type)-\(digest)"` (BLEReceivePipeline.swift:21) varies with `senderID`, so spoofed copies are not deduplicated.

An attacker takes any peer V's genuinely-signed bundle (inner signature valid) and re-broadcasts it in up to 200 packets with distinct fabricated `senderID`s. All pass `handlePrekeyBundle`; each creates a distinct `latestPrekeyBundleByPeer` entry. Once the cap (`prekeyBundleCapacity = 200`) is reached, `latestPrekeyBundleByPeer[sender] != nil || count < capacity` rejects *new* owners' bundles, so this node stops caching and serving other peers' bundles over sync. `PrekeyBundleStore` (keyed by noise key) is unaffected, so local sealing still works — impact is a targeted DoS on this node's bundle propagation. Announces don't have this exposure because their handler requires `senderID == PeerID(noiseKey)` and a valid signature.

### 3. LOW — Bundle top-up can produce two ciphertexts for one message ID, violating the "one message = one prekey" invariant.

`PrekeyBundleStore.ingest` carries forward assignments only for IDs the new bundle still offers (`carriedAssignments = ... filter { newIDs.contains($0.value) }`, PrekeyBundleStore.swift ~line 777). If message `m1` was assigned prekey id=0, deposited to couriers (sealed to id=0, in flight), and the owner then rotates id=0 out in a top-up, `assignPrekey(m1, …)` on a later re-deposit drops the stale assignment and picks a fresh id (covered by `topUpBundleKeepsConsumptionStateForSurvivingIDs`). Now two ciphertexts for `m1` exist (id=0 and id=1). This wastes a prekey and, if id=0's grace window lapses before delivery, makes the first ciphertext undeliverable — delivery still succeeds via the reassigned prekey, so this is a minor efficiency/robustness issue, not a correctness break.

## Probed and ruled out

- **Remote prekey drain / forced static downgrade:** `openPrekeyPayload` calls `markConsumed` only *after* a successful `readMessage` (NoiseEncryptionService.swift ~line 413-417), so garbage ciphertext to a valid prekey ID throws before consuming. Sender-side `usedIDs`/assignments are local per-sender state, so one party cannot exhaust another's view or the victim's actual prekeys remotely. No remote-forced downgrade.
- **Prekey-ID substitution / cross-prekey replay:** the prologue `"bitchat-prekey-v1" ‖ prekeyID(BE)` is mixed into the Noise transcript and the responder static is the prekey itself; relabeling the envelope's TLV 0x05 to another valid ID makes the DH/MAC fail. Domain separation vs static-sealed courier mail confirmed.
- **Signature substitution across owners:** `signableBytes()` covers `noiseStaticPublicKey`, so a signed prekey set cannot be moved under a different owner key.
- **Replay of an old bundle after rotation:** `ingest` rejects `existing.generatedAt >= bundle.generatedAt`; no downgrade.
- **Envelope v2 compat on `main`:** verified against `origin/main`'s `CourierEnvelope.decode` — TLV 0x05 hits `case nil: continue` (unknown, skipped), the v1 fields still parse, and a v1 client's static-key open fails and is dropped. v1 encoding stays byte-identical. Old clients handle bit 9 / MessageType 0x24 by ignoring the unknown sync bit and warning on the unknown packet type.
- **Panic wipe completeness:** `LocalPrekeyStore.wipe()` (keychain blob `prekeysV1`) via `clearPersistentIdentity`, plus `PrekeyBundleStore.shared.wipe()` in `ChatViewModel`; `lastPrekeyBundleSentAt` reset sits inside the `collectionsQueue` barrier (BLEService.swift:327), same guard as `sendPrekeyBundle`, so no race.
- **Amplification via since-cursor exemption:** the `.prekeyBundle` sync response is gated by the same `responseRateLimiter` (8/30s per peer) and bounded by the 200 cap; it mirrors the existing announce exemption rather than introducing a new class.
- **Concurrency of `assignRecipientPrekey`:** `collectionsQueue.sync` for registry lookup then a separate `PrekeyBundleStore` queue — no nested same-queue deadlock.

jackjackbits:
Pushed `b481a70` addressing the CI failure, both Codex findings, and the adversarial review. Summary mapping each item to its fix:

## CI failure — `Run Swift Tests (app)`
**Root cause (real bug, not just a test artifact):** the mesh receive queue `mesh.message` is a **concurrent** `DispatchQueue`. `handleReceivedPacket` dispatches an announce and its prekey bundle as two separate `.async` blocks, so the bundle handler can run *before* the announce populates the signing-key binding in the peer registry. The old `handlePrekeyBundle` dropped any bundle it couldn't yet attribute (`announceBoundSigningKey == nil`) and never retried it, so under CI parallel load the bundle was permanently lost — `verifiedBundleEntersGossipStore()` and `prekeySealedMailTravelsViaCourierAndOpens()` failed on `hasUsableBundle == false`. Locally the announce almost always won the race, so it passed.

**Fix:** a bundle that arrives before its binding is now retained per-owner in a bounded `pendingPrekeyBundles` map and re-attempted from `handleAnnounce` when the verified announce lands. The signing-key lookup and the stash happen in **one** `collectionsQueue` barrier, and the drain runs in a barrier ordered after the announce's registry write, so there's no lost-update window against the concurrent handler. Full macOS app suite (1098 tests) now green.

## Codex P2 + Review MEDIUM — authenticate the outer packet (done together)
`handlePrekeyBundle` now, in addition to the inner bundle signature:
- requires `PeerID(hexData: packet.senderID) == PeerID(publicKey: bundle.noiseStaticPublicKey)` — the owner's genuine bundle (direct or relayed/gossip-resent) always carries the owner's senderID + outer signature, so a fabricated senderID is rejected;
- verifies the **outer** packet Ed25519 signature (`toBinaryDataForSigning` covers senderID + timestamp) against the owner's announce-bound signing key — so a valid bundle can't be replayed under a fresh timestamp (defeating freshness) or a spoofed sender.

Gossip side (`GossipSyncManager._onPublicPacketSeen` `.prekeyBundle`): the store/dedup is now keyed by the **decoded bundle's `noiseStaticPublicKey`** (the authenticated identity), not the unauthenticated `packet.senderID`. One valid bundle sprayed under N fabricated sender IDs collapses to a single owner-keyed entry, so it can't exhaust the 200-owner cap and starve legit bundles. Cap remains 200 by identity.

New tests: `spoofedSenderPrekeyBundleIsRejected`, `replayedPrekeyBundleWithFreshTimestampIsRejected` (E2E), `prekeyBundleGossipIsKeyedByOwnerNotSenderID` (unit).

## Codex P1 — bump published-bundle freshness on consume
`LocalPrekeyStore.markConsumed` now returns whether it retired a prekey and advances `generatedAt` **strictly monotonically** (`max(nowMillis, generatedAt+1)`, so even same-millisecond consumes produce increasing stamps). Since consuming drops the prekey from `currentBundlePrekeys()`, the republished bundle is now strictly newer, so `PrekeyBundleStore.ingest`'s `existing.generatedAt >= bundle.generatedAt` rule accepts the replacement and peers stop assigning the consumed ID. `openPrekeyPayload` propagates the consumed flag; the courier-open path re-gossips (forced when the batch also topped up, else throttled to coalesce bursts; redeliveries don't re-consume, so they don't re-gossip). New test `consumingAPrekeyRepublishesANewerBundlePeersAccept` covers consume → newer republish → peer accepts → no stale-ID assignment → 48h grace; `persistsAcrossInstances` updated to expect the strict bump.

## Review LOW — double-ciphertext on reassignment — **DEFERRED (intentional)**
On a top-up that rotates out an already-assigned prekey ID, `ingest` drops that message's assignment and a re-deposit picks a fresh ID. I'm leaving this as-is: dropping the stale assignment is the *safer* behavior for deliverability — it always re-seals to a currently-offered, in-grace prekey, whereas making the assignment "sticky" would risk sealing to a rotated-out key that may already be past its grace window. The redundant earlier ciphertext is covered by its own 48h grace, and delivery still succeeds via the reassigned prekey, so the only cost is one wasted prekey. Net: minor, and the current behavior is preferable. Happy to revisit if you'd rather trade deliverability for the one-message-one-prekey invariant.

## Review HIGH — announce signing-key binding — **NOT fixed here (out of scope, inherited)**
The last-writer-wins `noiseKey → signingKey` binding in `SecureIdentityStateManager.upsertCryptographicIdentity` + announces never proving possession of the Noise static key is a **pre-existing architectural weakness in the announce trust model** that predates this PR and extends well beyond prekeys (it equally affects any announce-bound attribution). It's not introduced by this PR, and fixing it properly needs its own design pass, so I'm explicitly *not* silently ignoring it but deferring it to a dedicated follow-up. Recommended direction: pin the `noiseKey → signingKey` binding on first verified sighting and reject later conflicting rebindings from unauthenticated announces (first-seen-key pinning), or require proof-of-possession of the noise key before a binding is trusted for bundle attribution. Filing/tracking this separately is advisable before relying on prekey confidentiality against an active on-path attacker.

## Extra hardening
Guarded the panic/clear detached Application Support tree-deletes (`ChatViewModel.panicClearAllData`, `ChatPublicConversationCoordinator.clearCurrentPublicTimeline`) behind `TestEnvironment.isRunningTests` — the SPM test process shares that tree, so those detached wipes could land mid-test under parallel scheduling and flake unrelated file-dependent tests; this PR's added panic-wipe coverage shifts the parallel schedule enough to make that latent race matter.

PR #1382, opened 2026-07-06
NIP-13 proof-of-work for geohash channels: mine on send, relax rate limits for PoW senders

jackjackbits:
## What

NIP-13 proof-of-work for geohash channels (Nostr kind-20000 ephemeral events): outgoing messages mine a `["nonce", "<value>", "<target>"]` tag before publishing, and inbound events with validated PoW relax the per-sender intake rate limit. Motivation: public-timeline spam (flagged in the #1364 review) — PoW makes bulk spam expensive while staying fully permissionless. Scope is kind-20000 channel messages only; presence heartbeats (kind 20001), kind-1 location notes, and DMs are untouched, and there is no UI beyond a `pow=` field in an existing sampled debug log.

## Mining policy (send side)

- **Target: 8 leading zero bits** (`NostrPoW.targetBits`) — ~256 hash attempts, well under 100 ms on any supported device (the end-to-end test mines + signs in ~1 ms on an M-series Mac). All tuning constants sit together at the top of `bitchat/Nostr/NostrPoW.swift`.
- **Off the main actor**: `NostrProtocol.createMinedEphemeralGeohashEvent` is a nonisolated async variant of the existing creator; `created_at` is fixed up front and the event ID is recomputed every attempt, per NIP-13.
- **Allocation-light hot loop**: the canonical NIP-01 serialization is built once per attempt *series* (not per attempt) with a fixed-width 16-hex-char placeholder nonce; the loop rewrites only those 16 bytes in place and rehashes. The placeholder's byte range is located by serializing twice with two same-length placeholders and diffing — immune to JSON escaping and even to message content that contains the placeholder text.
- **Never blocks sending**: hard cap of 2 s (`NostrPoW.miningTimeCap`). On cap or cancellation the committed target steps down (halving to 0, which any hash satisfies) with a small budget per step, so the event still ships promptly with an *honest* commitment at the difficulty actually reached. Serialization failure ⇒ ships unmined, exactly like today.
- **Cancellable**: sending another message or leaving the channel cancels in-flight mining — cancellation only *expedites* it (the pending message still sends at the difficulty reached; nothing is ever dropped). This also keeps sends ordered and prevents mining delays from stacking.
- Mining covers all three kind-20000 send paths: normal sends (`ChatOutgoingCoordinator`), raw/emote sends (`sendPublicRaw`), and the geohash screenshot notice.

## Receive-side scoring + rate limiter integration

- `NostrPoW.validatedDifficulty(idHex:tags:)` implements NIP-13 commitment semantics: the committed target in the nonce tag is what counts — the ID's actual leading zero bits must meet it (otherwise the claim scores 0), and work beyond the commitment earns no extra credit, which stops low-target spammers from cashing in lucky hashes. No/malformed commitment scores 0.
- `NostrInboundPipeline` scores each inbound kind-20000 event (one hex decode + tag scan; negligible next to the Schnorr verify already on that path) and threads `powBits` into the existing `MessageRateLimiter`.
- In the limiter, PoW ≥ 8 bits (`NostrPoW.rateLimitBypassBits`) **skips the per-sender token bucket** (and doesn't drain it); the **per-content flood bucket still applies**, so replaying identical content with PoW is still throttled.

## Why old clients keep working

Non-PoW events are **never rejected** — they simply keep today's strict per-sender token-bucket limits, while PoW events buy themselves out of that bucket. No network partition, no behavior change for normal-rate senders; only floods are affected.

## Prior art: `feature/pow-geohash-mining-ui`

Evaluated and **reimplemented rather than cherry-picked**. The branch predates the conversation-store/coordinator refactor (its 311-line `ChatViewModel` diff no longer applies — `BitchatMessage` moved into the BitFoundation package, sends moved into coordinators), and its mechanism conflicts with this PR's requirements: unbounded mining loop (no time cap or cancellation), **hard receive-side filtering** that would partition old clients, per-attempt re-serialization via `JSONSerialization` in the hot loop, mining UI (hidden timestamps + animation, explicitly out of scope here), XCTest instead of Swift Testing, and `try!`/force unwraps. Kept from it: the overall shape (mine a nonce tag over the canonical serialization, count leading zero bits, low-target test strategy) and the 8-bit-class default difficulty philosophy of "cheap for one message, expensive for a flood".

## Tests

New Swift Testing suites, all passing (`bitchat (macOS)` scheme, derived data outside the repo):

- `NostrPoWTests` (11): leading-zero-bit vectors incl. exact byte boundaries and the NIP-13 spec example (36-bit ID); commitment semantics (claimed < actual credits the claim, actual < claimed voids it, missing/malformed tags score 0); mining produces a nonce tag whose committed difficulty an independently recomputed event ID meets, incl. content needing JSON escaping; end-to-end mined kind-20000 event signs, validates, and scores exactly 8; cancelled mining still returns an honest lower commitment.
- `MessageRateLimiterTests` (5): PoW event passes when the sender bucket is exhausted, low/zero PoW doesn't, PoW doesn't bypass the content bucket, PoW doesn't drain sender tokens.

Full macOS unit-test suite: **1069 tests in 151 suites, all passing** (includes the updated context-protocol mocks and the geohash send test, which now awaits the mining task).

### DIFF
--- a/bitchat/Nostr/NostrPoW.swift
+++ b/bitchat/Nostr/NostrPoW.swift
@@ -1,0 +1,215 @@
+import BitFoundation
+import CryptoKit
+import Foundation
+
+/// NIP-13 proof-of-work for Nostr events.
+///
+/// Outgoing kind-20000 geohash messages mine a `["nonce", "<value>", "<target>"]`
+/// tag so the event ID carries at least `target` leading zero bits. Inbound
+/// events are scored (never hard-rejected — the network has clients that do
+/// not mine): validated PoW at or above `rateLimitBypassBits` relaxes the
+/// per-sender public rate limit, everything else keeps the strict limits.
+enum NostrPoW {
+
+    // MARK: - Tuning
+
+    /// Difficulty (leading zero bits of the event ID) mined onto outgoing
+    /// geohash messages. 8 bits is ~256 hash attempts — typically well under
+    /// 100 ms on any supported device.
+    static let targetBits = 8
+
+    /// Inbound events whose validated NIP-13 difficulty is at least this many
+    /// bits skip the per-sender rate-limit bucket (the content-flood bucket
+    /// still applies). See `MessageRateLimiter.allow`.
+    static let rateLimitBypassBits = 8
+
+    /// Hard cap on mining wall-clock time. When it hits, the committed target
+    /// steps down until a difficulty reachable in a small extra budget is
+    /// found and the message is sent anyway — mining never blocks sending.
+    static let miningTimeCap: TimeInterval = 2.0
+
+    /// Budget for each stepped-down attempt after the main cap (or a task
+    /// cancellation) hits.
+    private static let fallbackTimeCap: TimeInterval = 0.15
+
+    /// The hot loop checks the deadline and task cancellation every this many
+    /// hash attempts.
+    private static let checkInterval: UInt64 = 1024
+
+    /// The nonce value is a fixed-width hex counter so the serialized event
+    /// template can be mutated in place without reallocation.
+    private static let nonceLength = 16
+
+    // MARK: - Scoring
+
+    /// Number of leading zero bits in a byte sequence (NIP-13 difficulty of
+    /// an event-ID hash).
+    static func leadingZeroBits<Bytes: Sequence<UInt8>>(_ bytes: Bytes) -> Int {
+        var total = 0
+        for byte in bytes {
+            if byte == 0 {
+                total += 8
+            } else {
+                total += byte.leadingZeroBitCount
+                break
+            }
+        }
+        return total
+    }
+
+    /// Validated NIP-13 difficulty of an inbound event.
+    ///
+    /// The committed target in the nonce tag is what counts: the actual
+    /// leading zero bits of the ID must meet it (otherwise the claim is void
+    /// and the event scores 0), and work beyond the commitment earns no extra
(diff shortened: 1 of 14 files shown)


jackjackbits:
**Codex P2 (serialize geohash sends during mining) — fixed** in 9ae440be.

Sending two location-channel messages back-to-back only cancelled the previous mining task and started a new one. NIP-13 cancellation merely *expedites* mining (the target is polled and steps down; it never aborts the send), so the cancelled task still appended + relayed once `mineNonceTag` returned. Both tasks ran concurrently and the second (faster-to-mine) could finish first, reordering messages in the timeline and on relays.

Fix: serialize the geohash send tasks. Each send now captures the previous send's `Task`, cancels it (to expedite so mining delays never stack), and `await`s its completion before it echoes and relays:

```swift
let previousSend = geohashMiningTask
previousSend?.cancel()
geohashMiningTask = Task { @MainActor [weak context = self.context] in
    await previousSend?.value
    // …mine, append echo, relay…
}
```

User-visible order now always equals send order. The "never block sending for >2s mining cap" guarantee is preserved: cancelling the previous task expedites the awaited mining, so a send is never blocked beyond `NostrPoW.miningTimeCap`.

Test added: two rapid sends where the first carries much larger content (so its mining runs longer) still land in send order for both the local echo and the relayed events. All `ChatOutgoingCoordinatorContextTests` pass on the macOS scheme; SwiftLint clean on touched files.

PR #1383, opened 2026-07-06
Private groups: creator-managed encrypted group chat over the mesh

jackjackbits:
Small encrypted crews — a hiking party, an event ops team — sitting between public broadcast and 1:1 DMs. Based on #1375 (capability bits): this build advertises the `.groups` capability bit in its announce.

## Crypto

- **Group key**: 32-byte ChaCha20-Poly1305 symmetric key per group, generated by the creator, stored only in the keychain. Group metadata (16-byte random group ID, name, epoch, roster, creator fingerprint) persists as file-protected JSON in Application Support. Both are dropped by `panicClearAllData()`.
- **Epochs**: a `uint32` that bumps on every key rotation. The epoch is bound into the AEAD as additional data together with the group ID, so ciphertext cannot be replayed across groups or epochs; receivers drop envelopes whose epoch doesn't match their current key.
- **Sender signatures**: the encrypted body carries `messageID, senderSigningKey, senderNickname, timestamp, content` plus an Ed25519 signature over `"bitchat-group-msg-v1"|groupID|messageID|timestamp|content`. Receivers verify the signature **and** require the signing key to be pinned in the creator-signed roster — key possession alone is not authorship, so an ex-member holding a stale key (or anyone who exfiltrates it) can't forge messages as someone else.

## Invite / rotation flow (creator-only in v1)

- `groupInvite` (0x06) and `groupKeyUpdate` (0x07) are Noise payloads carrying creator-signed group state: group ID, name, key, epoch, roster snapshot, and an Ed25519 signature over `"bitchat-group-v1"|groupID|epoch|key-hash|roster-hash`.
- State is only accepted when (a) the *authenticated Noise session peer* is the creator named in the state, (b) the signature verifies against the creator's signing key pinned in the roster, and (c) the receiver appears in the roster. A creator-signed roster that no longer includes you is a removal: the group and its key are dropped locally.
- Invites add a member without rotating (same key/epoch); `/group remove` rotates: new key, epoch+1, new roster, sent 1:1 to every remaining connected member. Epochs never regress.
- Hard cap of 16 members, enforced at the store, on the wire (roster blob), and at accept time.
- v1 consistency caveat: roster/key updates reach members with a live Noise path at that moment; offline members catch up on the next state push from the creator. No transitive state gossip in v1.

## Relay + sync behavior

- `MessageType.groupMessage = 0x25` broadcasts ride the standard broadcast relay class in `RelayController` (TTL clamps, jitter, duplicate suppression). The group ID travels in cleartext so relays dedup/filter without membership; non-members relay but cannot read.
- Group messages join gossip sync as their own store (bit 10 of the REQUEST_SYNC types bitfield) on the public-message cadence and freshness window, so members get the standard backfill after a partition. The bitfield was already a variable-length little-endian encoding inside a length-prefixed TLV: bit 10 widens it from 1 to 2 bytes, old clients decode 1–8 bytes and map unknown bits to nothing (covered by `SyncTypeFlagsGroupTests`).
- The outer 0x25 packet is intentionally unsigned: authenticity comes from the Ed25519 signature *inside* the ciphertext, which still verifies for gossip-backfilled copies long after the sender's announce expired.
- **Old-client caveat**: clients that predate 0x25 don't recognize the type and won't relay or store it — group messages degrade gracefully to the sub-mesh of upgraded nodes (the `.groups` capability bit makes that visible per-peer).

## Delivery semantics

Fire-and-flood like public chat: local echo goes straight to `.sent`, there are **no per-member acks in v1**. Messages sent while a member is offline arrive via gossip backfill only.

## UI surface

- Groups open as chat windows through the existing private-chat sheet, keyed by virtual `group_<32hex>` peer IDs (same pattern geo DMs use), with a `person.3` header showing `#name (memberCount)`, a group-specific encryption caption, and no favorite/fingerprint affordances (there is no single peer behind the header).
- `/group create <name>`, `/group invite @name`, `/group remove @name` (creator), `/group leave`, `/group list` — with command suggestions and geohash-context gating like `/fav`.
- A "groups" section in the people sheet (crown for creator, unread envelope), invitees get a system message + local notification and the group appears in their people sheet.
- Both themes (all chrome goes through the themed palette), VoiceOver labels/hints on every new element, all user-facing strings in `Localizable.xcstrings`. No header layout structure was changed beyond an icon branch inside the existing HStack.

## Tests

- `GroupProtocolTests` (10): state round-trip + creator-signature gates (member/outsider forgeries, epoch/key/roster tampering, creator-missing-from-roster), roster cap on the wire, message seal/open round-trip, wrong key, epoch relabeling, spoofed sender signature, tampered ciphertext, malformed payloads.
- `GroupStoreTests` (8): create/read, roster cap, creator-retention, rotation (epoch+1, key replaced), persistence across instances, key-loss handling, panic wipe, per-group key deletion.
- `SyncTypeFlagsGroupTests` (4): bit-10 allocation, 1→2 byte widening, unknown-bit tolerance, REQUEST_SYNC round-trip.
- Full macOS suite: **1078 tests in 152 suites, all passing**; BitFoundation package: 113 tests passing.

## Deferred to v2

- **Courier delivery** for group messages (offline catch-up is gossip-backfill only in v1).
- **Member-initiated invites** (creator-only in v1).
- **MLS-grade rotation**: v1 rotation is creator-fanout of a fresh symmetric key; no forward-secret tree, no post-compromise security beyond the rotation itself, and no protection against a malicious *creator*.
- Group media, mentions/autocomplete inside groups, and a peer-row "invite to group" action (needs a group picker; not low-cost).

## Android parity

Android needs: `groupMessage = 0x25`, Noise payload types 0x06/0x07, SyncTypeFlags bit 10 with the same trailing-zero-trimmed LE bitfield, the two TLV wire formats (state payload + message envelope, both forward-compatible: unknown TLVs ignored), the AAD binding (`groupID || epoch` big-endian), and the two signing domains (`bitchat-group-v1`, `bitchat-group-msg-v1`). Capability bit `.groups` = bit 3.

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -193,7 +193,9 @@ final class ConversationUIModel: ObservableObject {
 
     private func refreshComputedState() {
         if let selectedPeerID = privateConversationModel.selectedPeerID {
-            canSendMediaInCurrentContext = !(selectedPeerID.isGeoDM || selectedPeerID.isGeoChat)
+            // Media transfer is not wired for groups in v1; keep it off so the
+            // composer can't strand a media placeholder that never sends.
+            canSendMediaInCurrentContext = !(selectedPeerID.isGeoDM || selectedPeerID.isGeoChat || selectedPeerID.isGroup)
             return
         }
 
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -26,11 +26,22 @@ struct GeohashPersonRow: Identifiable, Equatable {
     let isBlocked: Bool
 }
 
+struct GroupChatRow: Identifiable, Equatable {
+    let peerID: PeerID
+    let name: String
+    let memberCount: Int
+    let isCreator: Bool
+    let hasUnread: Bool
+
+    var id: String { peerID.id }
+}
+
 @MainActor
 final class PeerListModel: ObservableObject {
     @Published private(set) var allPeers: [BitchatPeer] = []
     @Published private(set) var meshRows: [MeshPeerRow] = []
     @Published private(set) var geohashPeople: [GeohashPersonRow] = []
+    @Published private(set) var groupRows: [GroupChatRow] = []
     @Published private(set) var reachableMeshPeerCount = 0
     @Published private(set) var connectedMeshPeerCount = 0
     @Published private(set) var visibleGeohashPeerCount = 0
@@ -129,6 +140,13 @@ final class PeerListModel: ObservableObject {
             }
             .store(in: &cancellables)
 
+        chatViewModel.groupStore.$groups
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.refresh()
+            }
+            .store(in: &cancellables)
+
         peerIdentityStore.$encryptionStatuses
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in
@@ -217,20 +235,38 @@ final class PeerListModel: ObservableObject {
         }
 
         let geohashPeople = buildGeohashPeople()
+        let groupRows = buildGroupRows()
 
         self.meshRows = meshRows
         reachableMeshPeerCount = meshCounts.reachable
         connectedMeshPeerCount = meshCounts.connected
         self.geohashPeople = geohashPeople
         visibleGeohashPeerCount = geohashPeople.count
+        self.groupRows = groupRows
         renderID = (
             meshRows.map {
                 "\($0.id)-\($0.isConnected)-\($0.isReachable)-\($0.hasUnread)-\($0.isFavorite)-\($0.isBlocked)"
             } +
             geohashPeople.map {
                 "geo:\($0.id)-\($0.isTeleported)-\($0.isBlocked)-\($0.displayName)"
+            } +
+            groupRows.map {
+                "group:\($0.id)-\($0.name)-\($0.memberCount)-\($0.hasUnread)"
             }
         ).joined(separator: "|")
+    }
+
+    private func buildGroupRows() -> [GroupChatRow] {
+        let myFingerprint = chatViewModel.meshService.noiseIdentityFingerprint()
+        return chatViewModel.groupStore.groups.map { group in
+            GroupChatRow(
+                peerID: group.peerID,
+                name: group.name,
+                memberCount: group.members.count,
+                isCreator: group.creatorFingerprint == myFingerprint,
+                hasUnread: chatViewModel.hasUnreadMessages(for: group.peerID)
+            )
+        }
     }
 
     private func buildGeohashPeople() -> [GeohashPersonRow] {
--- a/bitchat/App/PrivateConversationModels.swift
+++ b/bitchat/App/PrivateConversationModels.swift
@@ -108,7 +108,13 @@ struct PrivateConversationHeaderState: Equatable {
     let encryptionStatus: EncryptionStatus?
 
     var supportsFavoriteToggle: Bool {
-        !conversationPeerID.isGeoDM
+        !conversationPeerID.isGeoDM && !conversationPeerID.isGroup
+    }
+
+    /// Group chats have no single peer identity behind the header: no
+    /// fingerprint screen, no per-peer encryption badge.
+    var isGroupConversation: Bool {
+        conversationPeerID.isGroup
     }
 }
 
@@ -206,6 +212,13 @@ final class PrivateConversationModel: ObservableObject {
             }
             .store(in: &cancellables)
 
+        chatViewModel.groupStore.$groups
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.refreshSelectedConversation()
+            }
+            .store(in: &cancellables)
+
         NotificationCenter.default.publisher(for: Notification.Name("peerStatusUpdated"))
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in
@@ -229,6 +242,26 @@ final class PrivateConversationModel: ObservableObject {
     }
 
     private func makeHeaderState(for conversationPeerID: PeerID) -> PrivateConversationHeaderState {
+        // Group chats: the "peer" is the whole crew. Name + member count in
+        // the header; availability reads as mesh since group traffic floods
+        // the local mesh, and the per-peer encryption badge does not apply.
+        if conversationPeerID.isGroup {
+            let displayName: String
+            if let group = chatViewModel.groupStore.group(for: conversationPeerID) {
+                displayName = "#\(group.name) (\(group.members.count))"
+            } else {
+                displayName = String(localized: "common.unknown", comment: "Fallback label for unknown peer")
+            }
+            return PrivateConversationHeaderState(
+                conversationPeerID: conversationPeerID,
+                headerPeerID: conversationPeerID,
+                displayName: displayName,
+                availability: .meshReachable,
+                isFavorite: false,
+                encryptionStatus: nil
+            )
+        }
+
         let headerPeerID = chatViewModel.getShortIDForNoiseKey(conversationPeerID)
         let peer = chatViewModel.getPeer(byID: headerPeerID)
         let displayName = resolveDisplayName(for: conversationPeerID, headerPeerID: headerPeerID, peer: peer)
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -9348,6 +9348,17 @@
         }
       }
     },
+    "content.accessibility.group_chat" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Group chat"
+          }
+        }
+      }
+    },
     "content.accessibility.jump_to_latest" : {
       "comment" : "Accessibility label for the jump to latest messages button",
       "extractionState" : "manual",
@@ -15055,6 +15066,17 @@
         }
       }
     },
+    "content.commands.group" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "create or manage private groups"
+          }
+        }
+      }
+    },
     "content.commands.help" : {
       "comment" : "Description of the /help command in the suggestions panel",
       "extractionState" : "manual",
@@ -18230,6 +18252,17 @@
         }
       }
     },
+    "content.input.group_placeholder" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "create|invite|leave|list"
+          }
+        }
+      }
+    },
     "content.input.placeholder.location" : {
       "comment" : "Composer placeholder for a public geohash channel, naming it",
       "extractionState" : "manual",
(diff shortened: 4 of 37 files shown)


jackjackbits (commented):
Deep adversarial review of the group crypto, membership state machine, virtual `group_` PeerID plumbing, and sync/relay surface. Overall the design is solid: the AEAD binds `groupID||epoch` as AAD, the inner Ed25519 sender signature plus roster-pinning defeats key-possession forgery, epoch regression is guarded, and TLV/roster parsing is fully bounds-checked. Nonce randomness, self-echo drop, and the peerID↔fingerprint derivation all check out. A handful of real behavioral gaps and two defense-in-depth signing gaps below, severity-ordered. Every item is backed by a concrete scenario; I list what I ruled out at the end.

## Confirmed

**1. Equal-epoch roster downgrade silently drops a just-added member (medium).**
`updateRoster` does not bump the epoch (`GroupStore.swift:95-103`), so multiple distinct creator-signed rosters coexist at the same epoch, and `applyGroupState` only rejects a *strictly older* epoch — equal epoch is last-writer-wins with no tie-breaker (`ChatGroupCoordinator.swift:537`, `upsert` at :543).
Scenario: group `{C,A}`@N. Creator invites B → signs `{C,A,B}`@N, sends a keyUpdate to A. Creator invites D → signs `{C,A,B,D}`@N, sends a keyUpdate to A. These are separate `sendGroupKeyUpdate` sends (`:241/:489`) and can arrive out of order (mesh retransmit / gossip reorder). If A applies the D-state then the B-state, the second upsert overwrites A's roster back to `{C,A,B}`, silently dropping D. Because invites keep the same key/epoch, D can still decrypt, but A now has no roster entry for D and drops every message D sends (`member(withSigningKey:)` miss at `:399`) until A happens to receive a newer state that re-lists D. Recommend a per-epoch roster version/counter, or bump the epoch on invite.

**2. Removed member becomes a silent zombie with no removal notice (medium).**
`removeMember` rotates the key (epoch+1, roster excluding the removee) locally, then `distributeState(excluding: [], ...)` iterates `rotated.members` (`ChatGroupCoordinator.swift:483-486`) — which no longer contains the removed member. Confidentiality is correct (they never get the new key), but:
- The removee is never notified: the removal branch at `:518` fires only on *receiving* a creator-signed roster-without-them, and no such state is ever sent to them.
- Their store still holds the group at old epoch N with the old key, so they believe they are still a member. New traffic sealed at N+1 is dropped by their `envelope.epoch == group.epoch` check (`:383`); their own sends (sealed at N) are dropped by everyone else's epoch check. Net: the group goes silent with no error or removal indication. Consider sending the removee an explicit "you were removed" signal (a roster-without-them state, or a distinct message type).

**3. `/block` is not enforced inside groups (low-medium).**
`handleGroupMessagePayload` (`ChatGroupCoordinator.swift:380-440`) has no block check. A peer the user has blocked, if they share a group, still has their group messages appended to the timeline and still triggers a local notification (`:432`). Blocking is applied on the 1:1 DM path and at `startPrivateChat`, but the group receive path bypasses it entirely. A blocked member should probably be filtered from group display/notifications too.

## Plausible / harden before v2

**4. Creator signature does not cover the group name (low; blocks a v2 hazard).**
`signingContent` = `domain|groupID|epoch|SHA256(key)|SHA256(roster)` (`GroupProtocol.swift:197-204`); `name` (TLV 0x02) is decoded and persisted (`asGroup`, `:299-306`) but not signed. It is not forgeable today because `applyGroupState` requires the Noise peer to *be* the creator (`:504-505`) and state travels 1:1 over authenticated Noise. But any future path that caches/relays/replays a signed state blob — e.g. the store-and-forward expansion in #1372 — lets a relay swap the display name while keeping a valid creator signature. Fold `Data(name.utf8)` (or its hash) into `signingContent` now, before state ever leaves the authenticated Noise channel.

**5. Sender signature does not cover the epoch (low).**
`messageSigningContent` = `domain|groupID|messageID|timestamp|content` (`GroupProtocol.swift:396-403`); the epoch lives only in the AEAD AAD. A current member who captured another member's decrypted inner TLV can re-seal those identical bytes under a later epoch key; `openMessage` re-verifies the original signature (valid — no epoch coverage) and the roster check still passes, so it is accepted and attributed to the original author at the wrong epoch. Per-conversation messageID dedup defeats the common case, but a message that was trimmed out of the recipient's timeline can be resurfaced, correctly attributed but out of context. Post-removal replay is already blocked by the roster check. Adding epoch to the signed content closes it for defense-in-depth.

## Ruled out (probed, not defects)

- **Nonce uniqueness**: random 96-bit ChaChaPoly nonce; per-group key rotating on removal, 16-member cap, chat volumes → collision probability negligible (well under 2^-64).
- **Cheap Ed25519 DoS via garbage 0x25**: the expensive signature verify is only reached after a successful AEAD decrypt, which requires the group key — an outsider cannot force it. Unknown-group packets drop immediately at the store lookup (`:382`).
- **0x25 gossip flood/eviction**: same threat class as existing unsigned public-message spam; separate 200-cap store, standard broadcast relay (TTL/dedup/jitter).
- **Read receipts / delivered acks to `group_` peers**: no-op. `PrivateChatManager.markAsRead` only emits receipts for messages whose `senderPeerID == peerID`, which is never true for a group conversation (sender is a real member's 16-hex ID, conversation key is the `group_` ID). No receipt is ever sent to a `group_` ID or to members.
- **Favorites / fingerprint on group headers**: correctly gated off (`PrivateConversationHeaderState.supportsFavoriteToggle`, `isGroupConversation`; the info button early-returns; `startPrivateChat` group branch skips block/handshake/favorite).
- **peerID↔fingerprint derivation** used by `connectedPeerID(forFingerprint:)` and inbound `senderPeerID`: verified correct — `PeerID(publicKey:)` is `sha256Fingerprint().prefix(16)`, so the first-16-hex-of-fingerprint assumption holds.
- **senderNickname spoof**: display prefers the creator-signed roster nickname; the attacker-settable fallback is never reached (roster nicknames are always non-empty) and is inside the ciphertext anyway.
- **Removed-member / old-key resurrection**: epoch bump on rotation + the `state.epoch < existing.epoch` regression guard prevent it.
- **TLV / roster parsing**: fully bounds-checked (`distance(...) >= n` guards, trailing `offset == endIndex`, count-cap, empty-roster reject); no OOB or malformed-accept path.
- **Roster fingerprint↔signingKey binding not re-verified on receive**: trusted because the creator signed it — within the documented "no protection against a malicious creator" v1 trust model.
- **SyncTypeFlags bit-10 widening**: LE trailing-zero-trim + `type(forBit:)`→nil for unknown bits keeps old clients compatible.

jackjackbits (commented):
Addendum — one more Confirmed defect from a deeper DM-pipeline trace, plus a note reinforcing finding #3.

**6. Media send inside a group chat silently dies, stuck on "sending" (medium).**
`canSendMediaInCurrentContext` only excludes geo contexts, so it returns `true` for group chats (`ChatViewModel.swift:103-110`, mirrored in `ConversationUIModel.swift:196`) — the mic/photo composer buttons are enabled in a group window. `sendVoiceNote`/`sendImage` append a placeholder message as `.sending` and then call `sendFilePrivate(packet, to: group_…)` (`ChatMediaTransferCoordinator.swift:96-113`). In `BLEService.sendFilePrivate` the recipient guard `Data(hexString: targetID.id)` fails for a `group_…` ID and the function logs "Invalid recipient peer ID for file transfer" and returns (`BLEService.swift:781-782`). Nothing is broadcast to the group, no failure is surfaced, and the placeholder is left stuck on "sending" forever. Fix: exclude `isGroup` in `canSendMediaInCurrentContext` at both sites (media isn't wired for groups in v1), or fail the message the way the too-large path does.

**Reinforcing #3 (block not enforced in groups):** confirmed the concrete entry point — long-press a group message → Block blocks the member's fingerprint, but `handleGroupMessagePayload` still decrypts, displays, marks unread, and fires a local notification for that member's later messages, unlike every other inbound path (`ChatTransportEventCoordinator.didReceiveMessage` guards `!isMessageBlocked`; the Noise path drops blocked senders).

Minor: the read-receipt / delivered-ack safety for `group_` conversations is real but *emergent* — it holds only because the receipt loops gate on `message.senderPeerID == <conversation peerID>`, which is never true for a group (sender is a member's 16-hex ID). A defensive `isGroup` guard/comment in `markPrivateMessagesAsRead` would keep that from regressing silently.

jackjackbits:
Pushed `ccccb2dd` to `feat/private-groups` addressing the Codex review and the adversarial review. Full macOS suite green (1082 tests) plus the group suites (27); SwiftLint clean on touched files.

## Codex findings
- **TLV oversize truncation (GroupProtocol.swift):** `GroupTLV.put` now throws `GroupTLVError.valueTooLong` instead of `UInt16(clamping:)` + prefix. `GroupMessageEnvelope.encode()` and the inner-message puts propagate it, so an oversize message fails to seal and `sendGroupMessage` surfaces `send_failed` — no silent truncation. `GroupStatePayload.encode()` returns nil on the same condition. Test: `oversizeMessageContentFailsToSealInsteadOfTruncating`.
- **Roster nickname byte-prefix split:** `GroupRosterCoding.encode` now truncates on a `Character` boundary (`truncatedNicknameBytes`), so the ≤64-byte prefix is always valid UTF-8 and the signed roster round-trips. Test: `multiByteNicknameTruncatesOnScalarBoundary` (40× `€`).

## Confirmed findings
1. **Equal-epoch roster downgrade — FIXED.** `inviteMember` now calls `rotateKey` (epoch+1, new key) instead of `updateRoster`, so every roster change is monotonic. Two out-of-order invite states can no longer share an epoch, and `applyGroupState`'s strict-older-epoch reject now gives a total order — a just-added member can't be last-writer-wins'd back out.
2. **Removed member = silent zombie — FIXED.** `removeMember` now sends the removee a creator-signed roster-without-them (`notifyRemovedMember`) so their existing removal branch fires and surfaces "removed from group". It carries a **throwaway all-zero key, never the rotated key**, so it grants no post-removal decryption; it's sent 1:1 over authenticated Noise so no remaining member sees it. Offline removees still miss it (same v1 limitation as any missed key update) — documented in the method.
3. **/block not enforced in groups — FIXED.** `handleGroupMessagePayload` now drops display + notification for a blocked member's fingerprint (`isFingerprintBlocked`, backed by `identityManager.isBlocked(fingerprint:)`), consistent with the other inbound paths.

## Plausible / harden-before-v2
4. **Creator signature omits name — FIXED now.** `signingContent` folds in `SHA256(name.utf8)`. Test: `creatorSignatureCoversName`.
5. **Sender signature omits epoch — FIXED now.** `messageSigningContent` folds in the epoch. Test: `messageSignatureCoversEpoch`.

Both #4/#5 are **wire-format-affecting** and must land before anyone ships — flagging for **Android parity**.

## Addendum
6. **Media send in group silently dies — FIXED.** `canSendMediaInCurrentContext` (ChatViewModel) and `refreshComputedState` (ConversationUIModel) now exclude `isGroup`, so the composer media buttons are disabled in groups and no placeholder gets stranded. Media-in-groups is deferred to **v2**.
- **Read-receipt safety (emergent → explicit):** added a `guard !peerID.isGroup` in `markPrivateMessagesAsRead` so the receipt loops can never emit into a `group_` conversation even if their `senderPeerID == peerID` matching is refactored.

## Deferred to v2 (with reasons)
- **Media transfer in groups** — needs `sendFilePrivate` (and fragment/relay) plumbing for the virtual `group_` recipient; out of scope for v1. Disabled + no stuck placeholder in the meantime.
- **Offline delivery of state/removal notices** — v1 is fire-and-flood with no per-member acks or store-and-forward for group state; offline members (including a removee) catch up only on the next creator-initiated send. Unchanged from the original design.

jackjackbits:
PASS on-device. Create/invite/message/remove all work incl. member-removal key rotation. Note: group messages show a single "sent" check and never advance — this is by design (v1 fire-and-flood, no per-member acks); consider hiding the glyph or adding per-member delivery/read in v2.

PR #1384, opened 2026-07-06
Gateway mode: opt-in mesh↔Nostr uplink for geohash channels

jackjackbits:
Based on #1375 (capability bits); targets `feat/capability-bits`.

One connected phone can now share its internet with the mesh: an opt-in **internet gateway** toggle uplinks the local geohash channel for mesh-only peers and rebroadcasts inbound channel traffic back onto the mesh. This is the disaster/protest path — a crowd with one online device gets a working `#geohash` channel.

## How it works

**Wire.** New `MessageType.nostrCarrier = 0x28` carrying a TLV payload (`NostrCarrierPacket`): direction flag (`toGateway`/`fromGateway`), geohash, and a complete **signed** kind-20000 Nostr event as JSON, capped at 16 KiB. Uplink deposits are DIRECTED packets (recipientID = gateway, relayed multi-hop like courier envelopes); downlink rides broadcast with default TTL. 2-byte TLV lengths, tolerant decoder, unknown TLVs skipped.

**Uplink (mesh peer → internet).** When a mesh-only user sends in a geohash channel with no working relay connection (`NostrRelayManager.isConnected == false`) and a reachable peer advertises the `.gateway` capability, the event — already signed locally with the user's per-geohash ephemeral identity — is wrapped `toGateway` and sent directed to one gateway. The gateway verifies (Schnorr signature, kind == 20000, `#g` tag matches the carrier geohash, sane geohash, ≤16 KiB, ±15 min clock skew) and publishes to the geo relays through the existing relay manager (which already respects Tor). A once-per-channel system notice "sent via mesh gateway" plus a log line keep the sender informed; **v1 is fire-and-forget — no gateway ack** (deferred, see below), and the event also stays in the sender's own relay send queue, so relays dedup by event ID if internet returns.

**Downlink (internet → mesh).** While the toggle is on and the gateway is viewing a geohash channel, inbound kind-20000 events from its subscription are wrapped `fromGateway` and broadcast on the mesh. Receivers verify the signature, then inject through the same `NostrInboundPipeline` entry point as relay-received events, so dedup (Nostr event-ID cache), blocking, rate limits, and rendering behave identically. Receivers need no UI at all.

## Threat model

- **Keys never leave the device.** Senders sign locally; only the finished signed event travels.
- **The gateway cannot forge or alter events.** Every carried event is Schnorr-verified by the gateway before publishing/rebroadcasting, and independently by relays and every receiver; a forged copy sharing a real event's ID is rejected before it can poison any cache (unit-tested).
- **Contents are already public.** Geohash chat is plaintext on Nostr; the carrier adds no confidentiality loss and no extra encryption.
- **Depositor accountability.** Uplink packets are signed with the sender's announced signing key and verified by the gateway (courier-deposit pattern), so quotas can't be dodged by spoofing sender IDs.

## Quotas & loop prevention

CourierStore-style bounds in `GatewayService.Limits`:
- Uplink: 10 events/min/depositor rate limit; while relays are down, a queue of 20 total / 5 per depositor (drop-oldest eviction), flushed on reconnect.
- Downlink: **30 rebroadcasts/min bandwidth guard** (BLE airtime is precious) with a bounded drop-oldest backlog that drains as the window frees.

Loop-prevention rules, stated in code comments and each unit-tested:
1. An event learned from a `fromGateway` broadcast is never re-published, never re-uplinked, and never rebroadcast — a second gateway on the same mesh cannot echo mesh-carried traffic back out.
2. Deposits publish at most once; relay echoes of our own uplinks rebroadcast at most once.
3. Uplink is only attempted for freshly composed events at the send site; received carrier events never re-enter the uplink path.
Mesh-level storms are additionally bounded by `MessageDeduplicator` + TTL.

## Capability advertising

`.gateway` (bit 2) is advertised **only while the toggle is on**. `BLEService.setLocalCapability(_:enabled:)` holds runtime bits (queue-guarded), ORs them into `PeerCapabilities.localSupported` at announce time, and forces a re-announce on change — the one-line "add build-time bit" pattern from #1375 is untouched for the other in-flight PRs.

## UX

- "internet gateway" toggle in the location channels sheet next to the Tor toggle, with a one-line explanation; IRC-style toggle, both themes, VoiceOver via the existing row semantics.
- Subtle globe indicator in the header while active (same pattern as the courier `figure.walk` indicator), with accessibility label and tooltip.
- New strings in `Localizable.xcstrings` (en).

## Tests

`NostrCarrierPacketTests` (5) + `GatewayServiceTests` (17), Swift Testing with closure-injected fakes: TLV round-trip/oversized/truncated/unknown-TLV, forged-signature rejection, kind/geohash/staleness gates, rate limit + queue quotas with drop-oldest eviction and flush-on-reconnect, all three loop-prevention rules, forged-broadcast poison resistance, downlink budget, receiver dedup and channel scoping, sender uplink conditions, toggle persistence.

Verified locally: `bitchat (macOS)` and `bitchat (iOS)` build; new suites pass (22/22) plus adjacent suites (Packets, BLEReceivePipeline, BLEPeerRegistry, GossipSyncManager, ProtocolContract, LocationChannels, ChatViewModel init/sending, CourierStore, BitFoundation package). Real relay publishing is not exercised — the policy layer is tested with injected fakes.

## Deferred (v2)

- Gateway acks / delivery confirmation for uplinked events (v1 fire-and-forget; delivery status stays honest — no fake "sent to relays" state).
- DM gift-wrap uplink (kind-1059) — needs its own privacy review.
- Persistence of the offline uplink queue across launches.
- **Android parity**: `nostrCarrier = 0x28`, the TLV layout, capability bit 2, and the quota/loop rules above are the cross-platform contract; mirroring tracked for the Android client.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -18,6 +18,9 @@ final class AppChromeModel: ObservableObject {
     private let chatViewModel: ChatViewModel
     private var cancellables = Set<AnyCancellable>()
 
+    /// Bulletin-board coordinator, created on first use of the board sheet.
+    private(set) lazy var boardManager = BoardManager(transport: chatViewModel.meshService)
+
     init(chatViewModel: ChatViewModel, privateInboxModel: PrivateInboxModel) {
         self.chatViewModel = chatViewModel
         self.nickname = chatViewModel.nickname
--- a/bitchat/App/LocationChannelsModel.swift
+++ b/bitchat/App/LocationChannelsModel.swift
@@ -12,20 +12,26 @@ final class LocationChannelsModel: ObservableObject {
     @Published private(set) var bookmarkNames: [String: String]
     @Published private(set) var locationNames: [GeohashChannelLevel: String]
     @Published private(set) var userTorEnabled: Bool
+    @Published private(set) var gatewayEnabled: Bool
 
     private let manager: LocationChannelManager
     private let network: NetworkActivationService
+    private let gateway: GatewayService
     private var cancellables = Set<AnyCancellable>()
 
     init(
         manager: LocationChannelManager? = nil,
-        network: NetworkActivationService? = nil
+        network: NetworkActivationService? = nil,
+        gateway: GatewayService? = nil
     ) {
         let manager = manager ?? .shared
         let network = network ?? .shared
+        let gateway = gateway ?? .shared
 
         self.manager = manager
         self.network = network
+        self.gateway = gateway
+        self.gatewayEnabled = gateway.isEnabled
         self.permissionState = manager.permissionState
         self.availableChannels = manager.availableChannels
         self.selectedChannel = manager.selectedChannel
@@ -96,6 +102,10 @@ final class LocationChannelsModel: ObservableObject {
         network.setUserTorEnabled(enabled)
     }
 
+    func setGatewayEnabled(_ enabled: Bool) {
+        gateway.setEnabled(enabled)
+    }
+
     func refreshMeshChannelsIfNeeded() {
         guard case .mesh = selectedChannel,
               permissionState == .authorized,
@@ -160,6 +170,10 @@ final class LocationChannelsModel: ObservableObject {
         network.$userTorEnabled
             .receive(on: DispatchQueue.main)
             .assign(to: &$userTorEnabled)
+
+        gateway.$isEnabled
+            .receive(on: DispatchQueue.main)
+            .assign(to: &$gatewayEnabled)
     }
 
     private func level(forLength length: Int) -> GeohashChannelLevel {
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -14,6 +14,9 @@ struct MeshPeerRow: Identifiable, Equatable {
     let isMutualFavorite: Bool
     let encryptionStatus: EncryptionStatus
     let showsVerifiedBadgeWhenOffline: Bool
+    /// Vouched-for by someone I verified, without an explicit verification of
+    /// mine — rendered as the unfilled seal (verified gets the filled one).
+    let showsVouchedBadge: Bool
 
     var id: String { peerID.id }
 }
@@ -183,13 +186,12 @@ final class PeerListModel: ObservableObject {
         let myPeerID = chatViewModel.meshService.myPeerID
         let meshRows = allPeers.map { peer in
             let isMe = peer.peerID == myPeerID
-            let verifiedBadge: Bool
-            if !isMe && !peer.isConnected,
-               let fingerprint = chatViewModel.getFingerprint(for: peer.peerID) {
-                verifiedBadge = peerIdentityStore.isVerified(fingerprint)
-            } else {
-                verifiedBadge = false
-            }
+            let fingerprint = isMe ? nil : chatViewModel.getFingerprint(for: peer.peerID)
+            let isVerifiedFingerprint = fingerprint.map { peerIdentityStore.isVerified($0) } ?? false
+            let verifiedBadge = !peer.isConnected && isVerifiedFingerprint
+            // Vouched is subordinate to verified: never show both seals.
+            let vouchedBadge = !isVerifiedFingerprint
+                && (fingerprint.map { chatViewModel.isVouchedFingerprint($0) } ?? false)
 
             return MeshPeerRow(
                 peerID: peer.peerID,
@@ -202,7 +204,8 @@ final class PeerListModel: ObservableObject {
                 isReachable: peer.isReachable,
                 isMutualFavorite: peer.isMutualFavorite,
                 encryptionStatus: chatViewModel.getEncryptionStatus(for: peer.peerID),
-                showsVerifiedBadgeWhenOffline: verifiedBadge
+                showsVerifiedBadgeWhenOffline: verifiedBadge,
+                showsVouchedBadge: vouchedBadge
             )
         }
 
--- a/bitchat/App/VerificationModel.swift
+++ b/bitchat/App/VerificationModel.swift
@@ -9,6 +9,14 @@ struct FingerprintPresentationState: Equatable {
     let theirFingerprint: String?
     let myFingerprint: String
     let isVerified: Bool
+    /// Number of currently-valid vouches from peers the user verified
+    /// (0 when the peer is explicitly verified — the stronger badge wins).
+    let voucherCount: Int
+    /// Display names of the (verified) vouchers, where known.
+    let voucherNames: [String]
+
+    /// Vouched for by ≥1 peer the user verified (and not explicitly verified).
+    var isVouched: Bool { voucherCount > 0 }
 
     var canToggleVerification: Bool {
         encryptionStatus == .noiseSecured || encryptionStatus == .noiseVerified
@@ -82,6 +90,24 @@ final class VerificationModel: ObservableObject {
         let encryptionStatus = chatViewModel.getEncryptionStatus(for: statusPeerID)
         let theirFingerprint = chatViewModel.getFingerprint(for: statusPeerID)
         let peerNickname = resolveDisplayName(for: peerID, statusPeerID: statusPeerID)
+        let isVerified = theirFingerprint.map { peerIdentityStore.isVerified($0) } ?? false
+
+        // Vouch state is recomputed on read: only vouchers still in the
+        // verified set count, so removing a verification silently retires the
+        // vouches that peer gave.
+        let vouchers: [VouchRecord]
+        if !isVerified, let theirFingerprint {
+            vouchers = chatViewModel.identityManager.validVouchers(for: theirFingerprint)
+        } else {
+            vouchers = []
+        }
+        let voucherNames = vouchers.compactMap { record -> String? in
+            guard let social = chatViewModel.identityManager.getSocialIdentity(for: record.voucherFingerprint) else {
+                return nil
+            }
+            if let petname = social.localPetname, !petname.isEmpty { return petname }
+            return social.claimedNickname.isEmpty ? nil : social.claimedNickname
+        }
 
         return FingerprintPresentationState(
             statusPeerID: statusPeerID,
@@ -89,7 +115,9 @@ final class VerificationModel: ObservableObject {
             encryptionStatus: encryptionStatus,
             theirFingerprint: theirFingerprint,
             myFingerprint: chatViewModel.getMyFingerprint(),
-            isVerified: theirFingerprint.map { peerIdentityStore.isVerified($0) } ?? false
+            isVerified: isVerified,
+            voucherCount: vouchers.count,
+            voucherNames: voucherNames
         )
     }
 
@@ -117,6 +145,17 @@ final class VerificationModel: ObservableObject {
             .store(in: &cancellables)
 
         chatViewModel.$allPeers
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.objectWillChange.send()
+            }
+            .store(in: &cancellables)
+
+        // Vouch state changes (ChatVouchCoordinator.notifyPeerTrustChanged)
+        // are signalled via this notification rather than a published
+        // property, so an open fingerprint sheet refreshes its vouched badge
+        // live when a vouch batch is accepted.
+        NotificationCenter.default.publisher(for: Notification.Name("peerStatusUpdated"))
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in
                 self?.objectWillChange.send()
--- a/bitchat/Identity/IdentityModels.swift
+++ b/bitchat/Identity/IdentityModels.swift
@@ -126,11 +126,35 @@ struct SocialIdentity: Codable {
     var notes: String?
 }
 
+/// Trust ladder: unknown → casual → vouched → trusted → verified.
+///
+/// Persistence compatibility: `TrustLevel` is stored by its *String* raw
+/// value ("unknown", "casual", …), not by ordinal position, so inserting
+/// `vouched` mid-ladder cannot corrupt previously persisted values — every
+/// pre-existing case keeps the exact raw value it was written with. The
+/// `vouched` tier is additionally never persisted into `SocialIdentity`
+/// (it's recomputed on read from stored vouches), so downgraded builds never
+/// encounter the unfamiliar raw value.
 enum TrustLevel: String, Codable {
     case unknown
     case casual
+    /// Transitively trusted: vouched for by at least one peer *I* verified.
+    /// Derived at read time — never written to persistent storage.
+    case vouched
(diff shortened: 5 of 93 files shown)


jackjackbits (commented):
Adversarial review of the gateway-mode policy layer, focused on the loop-prevention rules, uplink trust, downlink budgeting, event validation, capability bits, and privacy. I traced multi-gateway topologies on paper and read every hunk in context. **The crown-jewel loop rules hold** (details at the bottom); the confirmed defects are in the downlink airtime path and a couple of accounting/consistency gaps. Severity-ordered.

## Findings

### 1. Downlink rebroadcast has no age gate — every channel switch floods the 30/min budget with stale backfill that receivers provably drop (CONFIRMED)
`GatewayService.rebroadcastRelayEvent` (`bitchat/Services/Gateway/GatewayService.swift:258-277`) checks kind, dedup, and signature but never `created_at`. Receivers reject anything older than `maxEventAgeSeconds` = 15 min (`GatewayService.swift:349`). The feeding subscriptions use a **1-hour lookback, limit 200** (`GeohashSubscriptionManager.swift:148-154` / `241-247`, `TransportConfig.swift:144-145`).

Scenario: gateway user opens/switches to a geohash channel → up to 200 historical events pour into `rebroadcastRelayEvent`, saturating the 30/min BLE budget for minutes with 15–60-min-old events that **every** receiver's `validatedEvent` discards at line 349, while genuinely live events queue behind the backfill. This defeats the feature's whole purpose (fresh chat to mesh-only peers) on every resubscribe, no attacker required. Fix: mirror the `maxEventAgeSeconds` check inside `rebroadcastRelayEvent` (and in the drain loop).

### 2. Uplink per-depositor rate limit runs *after* Schnorr verify + dedup, so it can never bound the expensive work (CONFIRMED mechanism)
`handleUplinkDeposit` order (`GatewayService.swift:177-191`): `validatedEvent` (JSON parse + full Schnorr verify, line 177) → dedup guards (183-186) → `allowUplinkDeposit` (189). An authenticated mesh peer can replay one valid signed deposit with a fresh **packet** timestamp each time (defeating the BLE `messageID` dedup, which includes the timestamp). Each replay early-returns at the event-ID dedup guard (line 185) *before* the quota check, so the 10/min limiter never records a hit — yet the gateway pays one Ed25519 packet verify plus one JSON parse + Schnorr verify **per packet, on the MainActor**. The limiter as placed only throttles distinct *new* valid events. Reorder: cheap checks + dedup-by-carried-ID before `isValidSignature()`.

### 3. Deposits dropped by the queue quota are still rendered on the gateway's own timeline (CONFIRMED behavioral divergence)
When relays are down, `enqueueUplink` (`GatewayService.swift:224-234`) can return without storing the item (per-depositor cap at line 226, or drop-oldest overflow at 231), but control falls through to lines 202-204 which `injectInbound` the event unconditionally. Scenario: relays unreachable, a depositor already has 5 queued, sends a 6th while the gateway user views that geohash → the gateway user *sees* the message, but it was never queued, never published, and (deposits are directed, not broadcast) no other peer ever sees it. The gateway's timeline diverges permanently from what actually reached the channel — an honesty-of-delivery bug for a feature that markets "no fake sent state." Only inject on the path that actually accepted the event.

### 4. Mark-before-send + drain-only-on-next-event strands or permanently loses downlink events (PLAUSIBLE)
`rebroadcastEventIDs.insert` happens at `GatewayService.swift:269`, before the event is actually sent. Consequences: (a) an event evicted by the overflow drop at line 274 can never be rebroadcast even if the relay redelivers it — its ID is already marked; (b) `drainPendingDownlinks` is only invoked from `rebroadcastRelayEvent` (line 276) with no timer, so after a burst >30/min on a channel that then goes quiet, up to 30 queued events sit until the next relay event — and combined with finding #1, when they finally drain they may already be past `maxEventAgeSeconds`, i.e. pure wasted airtime. The `Limits` comment "drain as the window frees" (line 62-64) is not what the code does. The `downlinkBudget` test (`GatewayServiceTests.swift:335-338`) masks this: it injects a fresh event to trigger the drain, proving drain-on-next-event, not drain-when-window-frees.

### 5. `publishToRelays` fallback is inconsistent with the sender path and lands on relays no geo subscriber reads (PLAUSIBLE)
`configureGateway`'s `publishToRelays` (`ChatViewModelBootstrapper.swift:263-267`) falls back to `NostrRelayManager.shared.sendEvent(event)` (default relays) when `GeoRelayDirectory` returns no geo relays — while the local send path in the identical situation refuses to send at all (`GeohashSubscriptionManager.swift:286-288`). With an empty geo directory, a carried mesh event is published to default relays where no geo subscriber (which subscribe via closest-geo-relay) will ever find it: silent dead traffic rather than delivery.

### 6. Rejected queue deposits still consume a rate-limit slot (minor, self-correcting)
`allowUplinkDeposit` appends `now()` (`GatewayService.swift:244`) before `enqueueUplink` decides whether to keep the item (line 226). A depositor at its 5-item queue cap (relays down) burns a 10/min rate slot on each rejected attempt, so a later legitimate deposit can be throttled even though nothing was accepted. Bounded, clears in 60s.

### 7. Doc overclaims test coverage of loop rule 3 (minor / comment)
The header (`GatewayService.swift:33`) says each loop rule is "enforced here and unit-tested." Rule 3 (uplink only from the compose site) is a call-site convention no test enforces; only the `meshBroadcastEventIDs` backstop is tested. The `publishedEventIDs` backstop in `uplinkViaMesh` (`GatewayService.swift:323`) is exercised by no test. Either soften the comment or add the test.

## What I probed and ruled out (loop prevention holds)
- **Two gateways, same mesh, both with internet:** each rebroadcasts an event at most once (`rebroadcastEventIDs`); each other's `fromGateway` broadcast lands in `handleDownlinkBroadcast`, which **never** rebroadcasts — so the no-echo property is *structural*, not merely dependent on the bounded set. Bounded (event appears on mesh once per internet-connected gateway). No loop.
- **Bounded-set flush attack (10k events):** `handleDownlinkBroadcast` never republishes regardless of set state, so evicting an ID from `meshBroadcastEventIDs` cannot reopen a loop; reopening a *republish* additionally requires an active malicious depositor and relay redelivery (relays don't redeliver ephemeral events to a live sub). No practical loop.
- **Uplink trust:** directed deposits require a valid packet signature verified against the depositor's *announced* signing key (`BLEService.swift` `handleNostrCarrier`); an unannounced/spoofed sender is rejected. Quota keyed to the authenticated `senderID`. Tampering breaks the packet sig; replay of the same deposit dedups by event ID.
- **Forge/poison:** `NostrEvent.isValidSignature()` recomputes the event ID from content, and `handleDownlinkBroadcast` marks `meshBroadcastEventIDs` only *after* verification (`GatewayService.swift:301`) — a forged copy sharing a real ID cannot poison the never-republish set. Well-tested.
- **created_at ±15 min:** geohash ephemeral events use real `Date()` (`NostrProtocol.swift:182`), not the ±15-min `randomizedTimestamp()` (which is gift-wrap only), so legitimate events are not falsely rejected.
- **Privacy / Tor:** only the finished signed event (per-geohash ephemeral pubkey) is published; the depositor's mesh signing key stays on the packet layer and is stripped before publish. Publishing goes through `NostrRelayManager`, which centralizes Tor enforcement — no bypass introduced.
- **Capability bit:** `setLocalCapability` uses a `collectionsQueue` barrier and re-announces; the announce reads `localSupported ∪ runtimeCapabilities` inside the same `collectionsQueue.sync` as the routing snapshot — no race. Deposits arriving in the toggle-off advertising window are dropped by the `isEnabled` guard (silent, but fire-and-forget by design). One design note: a gateway keeps advertising `.gateway` even when *its own* relays are down, and senders pick `availableGatewayPeers().first` blindly — a mesh-only sender can deposit into a gateway whose queue may never flush while a working gateway is ignored.

jackjackbits:
Pushed `d4594b9` addressing the Codex + adversarial-review findings on the downlink/uplink paths.

## Fixed

**#1 — Downlink flood on stale backfill (Codex P2 + review #1).** `rebroadcastRelayEvent` now applies the *same* freshness window receivers enforce (`isFresh`, `maxEventAgeSeconds`) **and** validates the event's `#g` tag against the carrier geohash, both **before** touching `pendingDownlinks`/budget. A 1h/200-event resubscribe backfill now consumes zero downlink budget; the drain loop re-checks freshness too, so an event that ages out while queued is skipped. Test: `does not spend downlink budget on stale backfill or geohash mismatch`.

**#2 — Rate limit ran after Schnorr verify (review #2).** Split `validatedEvent` into a cheap `structurallyValidEvent` (parse/size/geohash/kind/`#g`/age, no crypto) + the signature check. `handleUplinkDeposit` now does structural check → carried-ID dedup → rate-token consume → **then** `isValidSignature()`. A replay of one valid deposit short-circuits at the ID dedup before any verify; a flood of distinct forged deposits is bounded by the rate token, not by main-actor Schnorr work. Covered by existing rate-limit/forgery tests.

**#3 — Quota-dropped deposits still injected (review #3).** `enqueueUplink` now returns whether the item was actually stored; `injectInbound` fires only on the accepted (published or queued) path. A per-depositor-cap or overflow drop is no longer rendered locally. Test: `a quota-dropped deposit is not rendered on the local timeline`.

**#4 — Queued downlinks stranded when channel goes quiet (Codex P2 + review #4).** Two changes: (a) **mark-after-send** — `rebroadcastEventIDs.insert` now happens in `drainPendingDownlinks` *after* the broadcast goes out, so an overflow-dropped event stays retryable on relay redelivery (a redelivery that's still queued is deduped via a `pendingDownlinks` membership check); (b) a **drain timer** (`scheduleDrainTimer`, closure-injected for deterministic tests, nil arms a real `Task`) is armed when the budget is exhausted with events still queued, so a burst-then-silence channel drains on the next window instead of waiting for unrelated inbound traffic. Test: `drains a downlink backlog on the scheduled timer when the channel goes quiet`.

**#5 — Asymmetric publish fallback (review #5).** The gateway `publishToRelays` closure now refuses (logs + returns) when `GeoRelayDirectory` yields no geo relays, matching `GeohashSubscriptionManager.sendGeohash`, instead of publishing dead traffic to default relays no geo subscriber reads.

**#7 — Doc overclaim + missing tests (review #7).** Softened the header: rule 3 is now described as a call-site convention with unit-tested `meshBroadcastEventIDs`/`publishedEventIDs` backstops (rules 1–2 remain "enforced here and unit-tested"). Added `uplinkViaMesh backstop refuses an event this gateway already published`, which exercises the previously-untested `publishedEventIDs` backstop.

## Deferred (with reasons)

**#6 — Rejected queue deposit burns a rate slot.** Now inherent to the #2 fix (rate token is consumed before the enqueue decision, as required to bound crypto work). Bounded and self-corrects in 60s; not worth reintroducing a pre-check that would reopen the verify-ordering issue.

**Design note — gateway advertises `.gateway` with its own relays down; senders pick `.first` blindly.** Deferred as a follow-up. Tying the runtime capability bit to `NostrRelayManager` connectivity is the correct fix but needs debounced re-announce to avoid announce churn on relay flap, and round-robin alone (without connectivity-gating) still sprays deposits into dead gateways — so a half-measure adds complexity without fully solving it. Recommend a dedicated pass: connectivity-gated `.gateway` bit + flap debounce, optionally with sender round-robin.

Full macOS app suite green (1081 + XCTest, exit 0), 21/21 gateway tests pass, SwiftLint clean on touched files.

jackjackbits:
## Device-testing finding (3-node mesh, iOS 26.5): gateway self-echoes uplinked events back onto the mesh

Field-tested gateway mode on real hardware (one mesh-only iPhone + one internet-connected gateway iPhone + a Mac, geohash `#sp`). **Uplink and downlink both work end to end** — the mesh-only device successfully read inbound internet geohash traffic and posted outbound via the gateway.

**Confirmed bug (reproduced every uplink):** the gateway re-broadcasts onto the mesh every event it just uplinked, when its own relay subscription echoes the event back. Representative log (`GatewayService.swift`):

```
[GW] uplink accept→publish 00ce9faa… to relays for #sp      (GatewayService.swift:250)
[GW] downlink rebroadcast 00ce9faa… to mesh for #sp         (GatewayService.swift:345)   ~0.15s later
[GW] uplink accept→publish 006798f0… to relays for #sp
[GW] downlink rebroadcast 006798f0… to mesh for #sp         ~0.29s later
[GW] uplink accept→publish 001bb690… to relays for #sp
[GW] downlink rebroadcast 001bb690… to mesh for #sp         ~0.15s later
```

Each mesh-originated event (`00ce9faa`, `006798f0`, `001bb690`) is published to relays, comes back via the gateway's own subscription, and is then downlink-rebroadcast onto the mesh it originated from.

**Impact:** not an infinite loop (mesh dedup catches it at the origin), but it roughly doubles the BLE airtime cost of every uplinked message and, on a busy channel, these echoes consume the 30/min downlink budget and starve genuine inbound internet events. This is the same `publishedEventIDs` self-suppression the review flagged as untested — the uplinked event id isn't being recorded/checked on the downlink path.

**Fix:** record every event id the gateway publishes (uplinks) and skip it in `drainPendingDownlinks()` so the gateway never downlinks an event it just uplinked. Add a test that an uplinked event arriving back via subscription is not rebroadcast to the mesh.

jackjackbits:
PASS on-device (bidirectional uplink+downlink confirmed between a mesh-only phone and an internet-connected gateway phone). One efficiency bug found and already filed as a separate comment (self-echo of uplinked events back onto the mesh) — see that comment.

jackjackbits:
The device-reported gateway self-echo is now fixed in a1a582cb.

**Root cause.** Every event the gateway uplinks (`publish()`) comes back through its own geohash subscription ~0.15s later. `rebroadcastRelayEvent` deduped against `meshBroadcastEventIDs`, `rebroadcastEventIDs`, and `pendingDownlinks` — but not `publishedEventIDs` — so an event the gateway just uplinked was downlink-rebroadcast onto the very mesh it originated from, doubling BLE airtime per uplinked message and able to starve the 30/min downlink budget on a busy channel. This matched the reported `[GW] uplink accept→publish` immediately followed by `[GW] downlink rebroadcast` for the same id.

**Fix.** `rebroadcastRelayEvent` now also skips ids present in `publishedEventIDs`. That set is the existing loop-rule-2 uplink cache — bounded (drop-oldest, capacity `maxTrackedEventIDs`) and populated only by `publish()` — so an uplinked event is never rebroadcast, while genuine inbound-from-internet events (never published here) still downlink-rebroadcast exactly as before. No new state; it reconciles with the existing dedup sets rather than adding a parallel one.

**Test.** Added `GatewayServiceTests.uplinkedEventNotSelfEchoed`: deposit → uplink an event, feed it back via `rebroadcastRelayEvent`, assert no mesh broadcast; a subsequently-received genuine inbound event still rebroadcasts once.

Validated: `bitchat (macOS)` + `bitchat (iOS)` sim build; `GatewayServiceTests` (22) + `NostrCarrierPacketTests` (5) = 27/27 pass.

Issue #907, opened 2025-12-02
Wifi Aware Support


allebone:
Support for Wi-Fi Aware Protocol - Might Increase Distance capabilites and simplfy any connected needs for P2P. 

https://developer.apple.com/documentation/WiFiAware

erlend-sh:
Further justification and implementation guidance for Aware from @grjte:

- https://hackmd.io/@grjte/bitchat-wifi-aware
- https://hackmd.io/@grjte/cross-platform-wifi-aware

anasouardini:
In my experience, Bluetooth is pointless!
Wifi and low frequencies (through SDR dongles) are a must.
IDK if phones support SDR, but today's phones support RJ-45, keyboards, mice, external microphones and cameras, so SDR would be probably supported.

jackjackbits:
Status: Wi-Fi transport work is in flight — PR #1385 implements an AWDL bulk data plane for large media with BLE fallback and is deliberately held on a prerequisite (recipient-aware transfer prep). Keeping this open as the tracking issue for Wi-Fi Aware proper (the cross-platform discovery/rendezvous layer, per grjte's design docs linked above), which #1385's transport would slot under.

PR #1385, opened 2026-07-06
Wi-Fi bulk transport: AWDL data plane for large media with BLE fallback

jackjackbits:
Based on #1375 (capability bits). BLE mesh moves media at roughly 15 KB/s (469-byte fragments spaced 30 ms). This PR keeps BLE as the control plane and hops large media onto peer-to-peer Wi-Fi (AWDL) for the data plane — three orders of magnitude more throughput — with BLE fragmentation as the always-available fallback.

## Negotiation flow (control plane stays on BLE/Noise)

1. `sendFilePrivate` finds a queued file **> 64 KiB** for a **directly connected (1-hop)** peer that advertises the **`.wifiBulk` capability** and has an **established Noise session**.
2. Sender sends `bulkTransferOffer` (`NoisePayloadType 0x04`) inside the Noise session: transferID (16 B), file size, SHA-256 of the payload bytes that will cross the channel, a fresh 32-byte session token, and a **random per-transfer Bonjour instance name** (never the nickname or peer ID). The listener starts at offer time so Bonjour registration overlaps the round-trip.
3. Receiver validates the offer against **its own** size cap and replies `bulkTransferResponse` (`0x05`) — accept (with its own 32-byte token half) or decline.
4. Receiver browses (`NWBrowser`, `includePeerToPeer`) for that exact instance of `_bitchat-bulk._tcp`, connects, authenticates with the first frame, and the sealed chunk stream flows. The listener lives only for the transfer window (≤ 60 s) and is torn down as soon as one client authenticates.

Both negotiation payloads are consumed inside the transport (`BLEService`); they never reach the UI layer, and the Nostr inbound paths explicitly ignore them.

## Channel security (why not TLS-PSK)

The TCP stream is encrypted and authenticated **independently of TLS**:

- Key: `HKDF-SHA256(ikm = senderToken ‖ receiverToken, salt = transferID, info = "bitchat-bulk-v1")` → ChaChaPoly-256. Both token halves traveled inside the established Noise session, so only the two endpoints can derive the key.
- Stream: length-prefixed frames, each a ChaChaPoly sealed box over a ≤ 64 KiB plaintext chunk. Nonces are structured, never random: `[direction byte][3 zero bytes][8-byte BE counter]`, and the reader requires the *exact* expected nonce per frame — no replay, reorder, truncation-splice, or cross-direction reflection.
- First frame on the wire is receiver → sender: a sealed copy of the transferID. A Bonjour-level gatecrasher that connects to the advertised listener cannot produce one valid MAC and is disconnected (covered by an integration test); the listener stays up so the real peer still succeeds.
- Final frame is a sealed receipt over the payload hash; the receiver has already verified the full SHA-256 against the offer before sending it. Sender reports completion only after the verified receipt.

Why not `NWProtocolTLS` with a PSK: TLS-PSK on Network.framework drags in cipher-suite negotiation, session resumption and versioning surface we don't need and can't easily audit, and its PSK identity travels in cleartext in the ClientHello. A boring AEAD framing over the Noise-exchanged secret is a few dozen lines, fully deterministic, and unit-testable byte-for-byte (HKDF vectors + tamper/replay tests included).

## Fallback matrix (BLE fragmentation is the common case and must stay bulletproof)

| Condition | Result |
|---|---|
| Feature disabled (`TransportConfig.wifiBulkEnabled`) | BLE, no offer |
| File ≤ 64 KiB | BLE, no offer |
| Peer lacks `.wifiBulk` capability | BLE, no offer |
| Peer not directly connected (multi-hop) | BLE, no offer |
| No established Noise session | BLE, no offer |
| Offer declined | BLE fallback (once) |
| Offer unanswered 10 s | BLE fallback (once) |
| Listener/socket failure, connection drop, MAC/hash mismatch, 60 s window expiry | BLE fallback (once) |
| Fallback also fails | Normal BLE failure path |

The fallback closure is guarded to fire **at most once** per transfer (tested with duplicate declines). Progress rides the existing `TransferProgressManager` (chunks report as fragments; the final tick is gated on the verified receipt), so the UI is completely unchanged; a fallback silently resets progress state before the BLE path re-registers it.

## Size caps

- New `FileTransferLimits.maxWifiBulkPayloadBytes = 8 MiB` for Wi-Fi-negotiated transfers; the **receiver enforces the cap from the accepted offer** (assembler refuses construction beyond the local ceiling and aborts on byte overflow past the offered size).
- BLE path keeps its existing caps (1 MiB frame, 512 KiB images/voice). `BitchatFilePacket.encode/decode` and the incoming-file validator take an explicit `limit` that defaults to the BLE cap, so every existing call site is unchanged.
- A >1 MiB payload whose Wi-Fi channel fails cannot ride BLE; it surfaces the normal failure path instead of silently stalling.

## Privacy notes

- AWDL announces presence at the Wi-Fi layer while a listener is up. Instance names are random per transfer (16 random bytes, hex), only resolvable during the ≤ 60 s transfer window, and torn down immediately after authentication or completion.
- No identifiers cross in cleartext: tokens, hash, and the instance name all traveled inside the Noise session; the TCP stream is AEAD-sealed end to end.
- Cellular/wired/loopback interfaces are prohibited on the production parameters; infrastructure Wi-Fi remains permitted alongside AWDL (same-LAN peers benefit, and the payload is independently encrypted either way).

## Testing

- **Unit**: offer/response TLV round-trip + malformed/forward-compat cases; HKDF derivation against fixed vectors (computed independently) + role/order sensitivity; chunk framing encrypt/decrypt, tamper rejection, replay/reflection rejection, hash-mismatch and overflow rejection; fallback policy matrix; size-cap enforcement at every layer.
- **Service-level**: real `WifiBulkTransferService` instances with loopback config — no-session/decline/timeout/wrong-peer/cancel/stop paths, fallback-once guarantees, deterministic teardown assertions.
- **Loopback integration (ran, passing)**: two services negotiate through an in-process pipe and move the payload over **real Network.framework TCP sockets on 127.0.0.1** (peer-to-peer/Bonjour disabled in tests — test hosts have no AWDL/mDNS), including a gatecrasher test where a keyless client connects first, gets disconnected on its first bogus frame, and the real peer still completes. Passing on both macOS and iOS Simulator.
- Builds: `bitchat (macOS)` and `bitchat (iOS)` (simulator, arm64) both green; regression suites for the touched areas (BLEService, fragmentation, media coordinator, protocol contract, BitFoundation package) all pass.

## Follow-ups

- **iOS 26 Wi-Fi Aware** is explicitly out of scope; it's the sanctioned successor for cross-ecosystem peer-to-peer and a natural second data plane behind the same offer/response negotiation.
- **Android parity**: Android needs a Wi-Fi Direct (or Wi-Fi Aware) counterpart speaking the same offer/response TLVs and frame format; until then Android peers simply never advertise the capability bit and stay on BLE.
- The offer already carries everything needed for resume support (transferID + hash); deposit/resume across reconnects is deferred.

### DIFF
--- a/bitchat/Features/media/ImageUtils.swift
+++ b/bitchat/Features/media/ImageUtils.swift
@@ -1,3 +1,4 @@
+import BitFoundation
 import Foundation
 import ImageIO
 import UniformTypeIdentifiers
@@ -13,11 +14,28 @@ enum ImageUtilsError: Error {
 }
 
 enum ImageUtils {
-    private static let compressionQuality: CGFloat = 0.82
-    private static let targetImageBytes: Int = 45_000
+    private static let compressionQuality: CGFloat = 0.85
+    // Upper bound for the compressed JPEG. This is only a ceiling: the encoder
+    // keeps whatever a photo naturally weighs at `defaultMaxDimension` and
+    // `compressionQuality`, and only steps quality down when a payload would
+    // exceed this budget. It stays well under `FileTransferLimits.maxImageBytes`
+    // (512 KiB) so the BLE path never overruns its cap.
+    //
+    // Wi-Fi bulk relevance: the old 45 KB / 448 px budget crushed every photo
+    // to ~40 KB — below `TransportConfig.wifiBulkMinPayloadBytes` (64 KiB) — so
+    // `WifiBulkPolicy.shouldOffer` never fired and the AWDL data plane was dead
+    // in production. A genuinely detailed photo at `defaultMaxDimension` now
+    // weighs well over 64 KiB, so it becomes Wi-Fi-bulk eligible to a capable
+    // direct peer while still riding BLE fragmentation for everyone else.
+    private static let targetImageBytes: Int = 200_000
     private static let maxSourceImageBytes: Int = 10 * 1024 * 1024
+    // Longest-side ceiling for shared photos. 448 px was thumbnail-tier and
+    // (together with the tiny byte budget) forced every photo below the Wi-Fi
+    // bulk threshold. 1024 px keeps a shared photo legible and lets detailed
+    // images clear 64 KiB, without approaching the 512 KiB hard cap.
+    static let defaultMaxDimension: CGFloat = 1024
 
-    static func processImage(at url: URL, maxDimension: CGFloat = 448, outputDirectory: URL? = nil) throws -> URL {
+    static func processImage(at url: URL, maxDimension: CGFloat = defaultMaxDimension, outputDirectory: URL? = nil) throws -> URL {
         try validateImageSource(at: url)
 
         let data = try Data(contentsOf: url)
@@ -47,34 +65,33 @@ enum ImageUtils {
     }
 
     #if os(iOS)
-    static func processImage(_ image: UIImage, maxDimension: CGFloat = 448, outputDirectory: URL? = nil) throws -> URL {
+    static func processImage(_ image: UIImage, maxDimension: CGFloat = defaultMaxDimension, outputDirectory: URL? = nil) throws -> URL {
         return try autoreleasepool {
-            // Scale the image first
-            let scaled = scaledImage(image, maxDimension: maxDimension)
-
-            // Get CGImage from UIImage - this is the key to stripping metadata
-            guard let cgImage = scaled.cgImage else {
-                throw ImageUtilsError.encodingFailed
-            }
-
-            // Use CGImageDestination to encode without metadata (same as macOS)
-            var quality = compressionQuality
-            guard var jpegData = encodeJPEG(from: cgImage, quality: quality) else {
-                throw ImageUtilsError.encodingFailed
-            }
-
-            // Compress to target size
-            while jpegData.count > targetImageBytes && quality > 0.3 {
-                quality -= 0.1
-                autoreleasepool {
-                    if let next = encodeJPEG(from: cgImage, quality: quality) {
-                        jpegData = next
-                    }
+            var dimension = maxDimension
+            var jpegData: Data?
+            // Downscale-and-compress until the payload fits the hard image cap.
+            // A normal photo converges on the first pass; this loop only kicks
+            // in for near-incompressible inputs (e.g. full-frame noise) that
+            // would otherwise overrun `maxImageBytes` at the raised dimension.
+            while true {
+                let scaled = scaledImage(image, maxDimension: dimension)
+                // Get CGImage from UIImage - this is the key to stripping metadata
+                guard let cgImage = scaled.cgImage else {
+                    throw ImageUtilsError.encodingFailed
                 }
+                guard let data = compressToBudget(cgImage) else {
+                    throw ImageUtilsError.encodingFailed
+                }
+                jpegData = data
+                if data.count <= FileTransferLimits.maxImageBytes || dimension <= minRetryDimension {
+                    break
+                }
+                dimension = (dimension * dimensionRetryFactor).rounded(.down)
             }
+            guard let finalData = jpegData else { throw ImageUtilsError.encodingFailed }
 
             let outputURL = try makeOutputURL(outputDirectory: outputDirectory)
-            try jpegData.write(to: outputURL, options: .atomic)
+            try finalData.write(to: outputURL, options: .atomic)
             return outputURL
         }
     }
@@ -93,66 +110,49 @@ enum ImageUtils {
         UIGraphicsEndImageContext()
         return rendered ?? image
     }
-
-    // Shared EXIF-stripping JPEG encoder for both iOS and macOS
-    private static func encodeJPEG(from cgImage: CGImage, quality: CGFloat) -> Data? {
-        guard let data = CFDataCreateMutable(nil, 0) else {
-            return nil
-        }
-        guard let destination = CGImageDestinationCreateWithData(data, UTType.jpeg.identifier as CFString, 1, nil) else {
-            return nil
-        }
-        // Security: Strip ALL metadata (EXIF, GPS, TIFF, IPTC, XMP)
-        // By only specifying compression quality and no metadata keys,
-        // we ensure a clean JPEG with no privacy-leaking information
-        let options: [CFString: Any] = [
-            kCGImageDestinationLossyCompressionQuality: quality
-        ]
-        CGImageDestinationAddImage(destination, cgImage, options as CFDictionary)
-        guard CGImageDestinationFinalize(destination) else {
-            return nil
-        }
-        return data as Data
-    }
     #else
-    static func processImage(_ image: NSImage, maxDimension: CGFloat = 448, outputDirectory: URL? = nil) throws -> URL {
+    static func processImage(_ image: NSImage, maxDimension: CGFloat = defaultMaxDimension, outputDirectory: URL? = nil) throws -> URL {
         return try autoreleasepool {
-            let scaled = scaledImage(image, maxDimension: maxDimension)
-            guard let inputCG = scaled.cgImage(forProposedRect: nil, context: nil, hints: nil) else {
-                throw ImageUtilsError.encodingFailed
-            }
-            let width = inputCG.width
-            let height = inputCG.height
-            let colorSpace = CGColorSpace(name: CGColorSpace.sRGB) ?? CGColorSpaceCreateDeviceRGB()
-            guard let context = CGContext(
-                data: nil,
-                width: width,
-                height: height,
-                bitsPerComponent: 8,
-                bytesPerRow: 0,
-                space: colorSpace,
-                bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
-            ) else {
-                throw ImageUtilsError.encodingFailed
-            }
-            context.draw(inputCG, in: CGRect(x: 0, y: 0, width: width, height: height))
-            guard let cgImage = context.makeImage() else {
-                throw ImageUtilsError.encodingFailed
-            }
-            var quality = compressionQuality
-            guard var jpegData = encodeJPEG(from: cgImage, quality: quality) else {
-                throw ImageUtilsError.encodingFailed
-            }
-            while jpegData.count > targetImageBytes && quality > 0.3 {
-                quality -= 0.1
-                autoreleasepool {
-                    if let next = encodeJPEG(from: cgImage, quality: quality) {
-                        jpegData = next
-                    }
+            var dimension = maxDimension
+            var jpegData: Data?
+            // See the iOS path: normal photos converge immediately; the loop
+            // only shrinks further for near-incompressible inputs so the
+            // output never overruns `maxImageBytes`.
+            while true {
+                let scaled = scaledImage(image, maxDimension: dimension)
+                guard let inputCG = scaled.cgImage(forProposedRect: nil, context: nil, hints: nil) else {
+                    throw ImageUtilsError.encodingFailed
                 }
+                let width = inputCG.width
+                let height = inputCG.height
+                let colorSpace = CGColorSpace(name: CGColorSpace.sRGB) ?? CGColorSpaceCreateDeviceRGB()
+                guard let context = CGContext(
+                    data: nil,
+                    width: width,
+                    height: height,
+                    bitsPerComponent: 8,
+                    bytesPerRow: 0,
+                    space: colorSpace,
+                    bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
+                ) else {
+                    throw ImageUtilsError.encodingFailed
+                }
+                context.draw(inputCG, in: CGRect(x: 0, y: 0, width: width, height: height))
+                guard let cgImage = context.makeImage() else {
+                    throw ImageUtilsError.encodingFailed
+                }
+                guard let data = compressToBudget(cgImage) else {
+                    throw ImageUtilsError.encodingFailed
+                }
+                jpegData = data
+                if data.count <= FileTransferLimits.maxImageBytes || dimension <= minRetryDimension {
+                    break
+                }
+                dimension = (dimension * dimensionRetryFactor).rounded(.down)
             }
+            guard let finalData = jpegData else { throw ImageUtilsError.encodingFailed }
             let outputURL = try makeOutputURL(outputDirectory: outputDirectory)
-            try jpegData.write(to: outputURL, options: .atomic)
+            try finalData.write(to: outputURL, options: .atomic)
             return outputURL
         }
     }
@@ -172,6 +172,31 @@ enum ImageUtils {
         scaledImage.unlockFocus()
         return scaledImage
     }
+    #endif
+
+    // When even the quality floor can't get an image under the byte budget,
+    // shrink the longest side by this factor and re-encode. Bounded below so
+    // the retry loop always terminates.
+    private static let dimensionRetryFactor: CGFloat = 0.75
+    private static let minRetryDimension: CGFloat = 256
+
+    /// Encodes `cgImage` to JPEG, stepping quality down toward
+    /// `targetImageBytes`. Shared by both platforms.
+    private static func compressToBudget(_ cgImage: CGImage) -> Data? {
+        var quality = compressionQuality
+        guard var jpegData = encodeJPEG(from: cgImage, quality: quality) else {
+            return nil
+        }
+        while jpegData.count > targetImageBytes && quality > 0.3 {
+            quality -= 0.1
+            autoreleasepool {
+                if let next = encodeJPEG(from: cgImage, quality: quality) {
+                    jpegData = next
+                }
+            }
+        }
+        return jpegData
+    }
 
     // Shared EXIF-stripping JPEG encoder for both iOS and macOS
     private static func encodeJPEG(from cgImage: CGImage, quality: CGFloat) -> Data? {
@@ -193,7 +218,6 @@ enum ImageUtils {
         }
         return data as Data
     }
-    #endif
 
     private static func makeOutputURL(outputDirectory: URL? = nil) throws -> URL {
         let formatter = DateFormatter()
--- a/bitchat/Info.plist
+++ b/bitchat/Info.plist
@@ -33,12 +33,18 @@
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>LSMinimumSystemVersion</key>
 	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+	<key>NSBonjourServices</key>
+	<array>
+		<string>_bitchat-bulk._tcp</string>
+	</array>
 	<key>NSBluetoothAlwaysUsageDescription</key>
 	<string>bitchat uses Bluetooth to create a secure mesh network for chatting with nearby users.</string>
 	<key>NSBluetoothPeripheralUsageDescription</key>
 	<string>bitchat uses Bluetooth to discover and connect with other bitchat users nearby.</string>
 	<key>NSCameraUsageDescription</key>
 	<string>bitchat uses the camera to scan QR codes to verify peers.</string>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>bitchat uses peer-to-peer Wi-Fi to transfer large photos and voice notes directly between nearby devices.</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>bitchat uses your approximate location to compute local geohash channels for optional public chats. Exact GPS is never shared.</string>
 	<key>NSMicrophoneUsageDescription</key>
--- a/bitchat/Protocols/PeerCapabilities+Local.swift
+++ b/bitchat/Protocols/PeerCapabilities+Local.swift
@@ -1,0 +1,7 @@
+import BitFoundation
+
+extension PeerCapabilities {
+    /// Capabilities this build advertises in its announce packets.
+    /// Each feature adds its bit here when it ships.
+    static let localSupported: PeerCapabilities = []
+}
--- a/bitchat/Protocols/PeerCapabilities+Local.swift
+++ b/bitchat/Protocols/PeerCapabilities+Local.swift
@@ -3,5 +3,5 @@ import BitFoundation
 extension PeerCapabilities {
     /// Capabilities this build advertises in its announce packets.
     /// Each feature adds its bit here when it ships.
-    static let localSupported: PeerCapabilities = []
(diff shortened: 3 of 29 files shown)


jackjackbits:
**Codex P2 (normalize peer IDs before Wi-Fi eligibility) — fixed** in 2b7fd200.

When `peerID` was a full 64-hex Noise key (stable/verified private chats address the peer that way), the Noise session is keyed by the short routing ID, so `noiseService.hasEstablishedSession(with:)` returned false, `shouldOffer` never selected Wi-Fi, and a payload over the BLE cap cancelled as "Wi-Fi unavailable" even when the direct peer advertised `.wifiBulk`. (The registry lookups `isPeerConnected`/`peerCapabilities` already normalize; the session check was the gap.)

Fix: normalize with `peerID.toShort()` once before the session/capability checks — extracted into a single `wifiBulkSendCandidate(payloadBytes:to:)` helper — and use the normalized `routingID` for the Wi-Fi negotiation packet (`wifiBulkService.sendFile`) and the BLE fallback too.

Test added (`BLEServiceCoreTests.wifiBulkEligibility_resolvesWhenAddressedBy64HexNoiseKey`): establishes a real Noise session keyed by the short routing ID, asserts the raw 64-hex lookup misses it (`hasEstablishedSession == false`), then registers the peer as a direct `.wifiBulk` neighbor and verifies the send path — addressed by the 64-hex key — resolves an eligible candidate and `shouldOffer` returns true. Added small `_test_` seams (`_test_noiseService`, `_test_registerConnectedPeer`, `_test_wifiBulkSendCandidate`) to drive it.

All BLEServiceCoreTests + WifiBulkPolicyTests pass on the macOS scheme (21 tests). SwiftLint clean on touched files.

jackjackbits:
## Fix: image prep never crossed the 64 KiB Wi-Fi bulk threshold

On-device testing showed Wi-Fi bulk (AWDL) works, but it never fired in production. Root cause is in the image-prep path, not the transport.

### Root cause
`ImageUtils.processImage` downscaled every photo to **448 px** longest side and force-compressed it to a **45 KB** byte budget (`targetImageBytes`, quality 0.82). A typical camera photo came out **~40 KB** — below `TransportConfig.wifiBulkMinPayloadBytes` (64 KiB). Since `WifiBulkPolicy.shouldOffer` requires `payloadBytes > 64 KiB`, the offer was never made and every real photo fell back to BLE fragmentation. The AWDL path was dead in practice.

### Fix (raise the prep budget)
Prep runs before the recipient's capabilities are known in any usable way, and doesn't apply to public broadcasts at all, so a recipient-aware two-tier prep would need capability plumbing through the pipeline for limited benefit. The clean, safe fix is to raise the budget so genuinely detailed photos naturally clear 64 KiB while staying under the `FileTransferLimits.maxImageBytes` (512 KiB) hard cap:

- `defaultMaxDimension` 448 → **1024 px**
- `compressionQuality` 0.82 → **0.85**
- `targetImageBytes` 45 KB → **200 KB** (a ceiling, not a target to hit — the encoder keeps whatever a photo naturally weighs)

`wifiBulkMinPayloadBytes` is **unchanged** (correctly 64 KiB).

### Before / after
Measured on a representative photo-like image (gradients + medium-frequency texture at ~12 MP source), replaying both pipelines:

| | prepared size | over 64 KiB threshold? |
|---|---|---|
| **old** (448 px, q0.82, 45 KB budget) | 40,954 B (~40 KB) | no |
| **new** (1024 px, q0.85, 200 KB budget) | 195,168 B (~190 KB) | **yes** |

~190 KB is ~2.9x the offer threshold and ~2.6x under the 512 KiB image cap, so detailed photos are now Wi-Fi-bulk eligible to a capable direct peer while still riding BLE fragmentation for everyone else.

### Robustness
The raised dimension means near-incompressible inputs (e.g. full-frame noise) could exceed `maxImageBytes` at the quality floor — impossible at 448 px, newly possible at 1024 px. Prep now **downscales-and-retries** until the payload fits the hard cap, so a send can never fail with `imageTooLarge` on pathological input. Also de-duped the previously copy-pasted per-platform `encodeJPEG` into one shared helper.

### Tests
Added `prepareImagePacket_detailedImageExceedsWifiBulkThreshold` (random-noise image — an incompressible stand-in for a detailed photo) asserting the prepared payload is `> wifiBulkMinPayloadBytes` **and** `<= maxImageBytes`. Passes on macOS and iOS simulator. Both schemes build clean (DerivedData outside repo, fresh mtime verified).

### Residual limitation
Prep is not recipient-capability-aware, so BLE-only peers and public broadcasts now also carry ~190 KB images (still within the existing 512 KiB cap; previously ~40 KB). Per-recipient fidelity selection is deferred — it needs capability plumbing into the prep pipeline and doesn't help public broadcasts.

jackjackbits:
#907

Chessing234 (commented):

jozanek (commented):
Thanks for the unusually complete write-up — the fallback matrix, the independently computed HKDF vectors, and the loopback integration test that moves bytes over real `Network.framework` sockets all made this reviewable in a way most transport PRs aren't.

I read this as: BLE stays the control plane, large private media hops to AWDL for the data plane, and BLE fragmentation stays the always-available fallback. If I've misread the goal, most of what follows is aimed at the wrong thing.

**The main thing I found isn't in the diff — it's underneath it.** Two merged PRs have moved three seams this branch attaches to, so what's left reads more like a re-derivation than a rebase:

- `BLEFileTransferHandler` now requires `packet.signature != nil` (`:478`, #1434), so the synthesized packet in `deliverReceivedFile` wouldn't deliver.
- `sendFilePrivate` now routes through `privateMediaSendPolicy(to:)`, and rebased as-is the Wi-Fi check would sit ahead of that gate. To be clear, that gate landed three weeks after you cut this branch — nothing here routed around anything.
- `PeerCapabilities+Local.swift` is an add/add conflict. I'd initially flagged this as silently un-shipping private-media encryption; `reservedReplacementBitIsNotAdvertised` already pins that pair, so CI catches it. `.vouch` / `.prekeys` / `.groups` are the unpinned ones.

**One I missed on the first pass, and it may be the sharpest.** In `WifiBulkChannel`, a candidate that connects and sends nothing looks like it's never evicted — there's no per-candidate auth deadline that I can find. Four idle sockets from anywhere on the subnet would refuse the real peer at the `maxCandidateConnections` guard and push every transfer to the 60 s window. `gatecrasherIsRejected` covers the noisy case; the silent one is the expensive one. Left inline.

**On the image-prep change.** I only found it in `9b8256bf` and your follow-up comment, not in the description — and I think it deserves to be its own PR. Because prep is recipient-unaware, the raise also applies to `sendFileBroadcast`, which can never use Wi-Fi bulk: roughly 88 → 417 fragments and ~2.6 s → ~12.5 s of airtime per broadcast photo. You name the same limitation yourself in that comment. It also merges clean today, which the rest of the branch doesn't.

**On the seam, since I expected to argue the opposite:** I went in thinking Wi-Fi selection belonged behind `Transport`/`MessageRouter` per CLAUDE.md, and the code changed my mind. `MessageRouter.swift` owns "the Bluetooth-first / Nostr-fallback / queue transport-selection logic" for DM text and has no media surface at all, and `MeshFileTransferring` is already the capability port that owns `sendFilePrivate` and `privateMediaSendPolicy`. Hosting this on `BLEService` looks right to me; if anything the open question is whether it should register through that port rather than as a stored property.

**Questions I'm holding:**

1. On #907 you wrote this is "deliberately held on a prerequisite (recipient-aware transfer prep)". I traced that plumbing and it looks smaller than it might feel: the recipient is already in scope at `ChatMediaTransferCoordinator.swift:701`, eleven lines above the prep dispatch, and `ChatVouchCoordinator.swift:76` is a one-line precedent for reaching `peerCapabilities`. I sketched it at roughly 105 lines across five files. Does that match your read, or is there a layer I'm not seeing? The retry/retain path looks like the genuinely awkward part.

2. How would we know AWDL is working in the field? Every failure degrades to a working BLE transfer, so a dead data plane and a healthy one look identical from outside — which is how the 64 KiB threshold problem stayed hidden until device testing. A counter behind the existing `MeshDiagnosing` surface looks like ~30 lines.

3. Was reusing `NoiseCipherState` considered? It's already counter-nonce ChaChaPoly with a replay window, carries the Cure53 remediations, and zeroizes keys on deinit. I think rejecting TLS-PSK is right, though I'd put the weight on "untestable in CI" rather than the ClientHello PSK identity — that identity is smaller than the Bonjour instance name already broadcast in clear.

4. Smaller, and flagging it here rather than in its own thread: `surfaceUndeliverableTransfer` (`BLEService.swift:884`) emits `.cancelled`, which removes the outgoing placeholder, so the photo vanishes with nothing to say what happened. I couldn't find a way to reach it today, since no producer exceeds 512 KiB — it goes live when the caps move. `main` added `rejectBeforeStart(id:reason:)` for exactly that case.

**What I couldn't verify:** anything requiring real radios — I ran no builds or tests, so the AWDL behavior, the permission flow, and the 60 s stall are read from the code rather than observed. I also couldn't find any issue or PR describing the "recipient-aware transfer prep" prerequisite, so I may be reconstructing the wrong thing. And I searched all 407 issues for reports of slow or over-compressed media and found none, which shapes how I read the urgency — if there's demand signal I didn't look in the right place for, that changes my read.

**What would move me to approve:** the three integration points resolved against current `main`; a per-candidate auth deadline; the `ImageUtils` change split out; a diagnostics counter; and either `wifiBulkEnabled` off on merge or a user-facing off switch. If you'd rather keep it held as you said on #907, rebasing to off would at least stop ~2,300 lines of `WifiBulk*` and its tests rotting — and none of the conflicts are in those files; they're all in the ~250 lines of glue.

The single question that would most change my verdict: **is recipient-aware prep the blocker you meant, or is there a different prerequisite I haven't found?** If it's the former, sequencing that first dissolves most of this review.

bitchat/Protocols/PeerCapabilities+Local.swift
```diff
@@ -0,0 +1,7 @@
+import BitFoundation
+
+extension PeerCapabilities {
+    /// Capabilities this build advertises in its announce packets.
+    /// Each feature adds its bit here when it ships.
+    static let localSupported: PeerCapabilities = TransportConfig.wifiBulkEnabled ? [.wifiBulk] : []
```
jozanek, line 6: `main` lists five bits here — `.vouch`, `.prekeys`, `.groups`, `.privateMedia`, `.privateMediaReceipts` — so this is an add/add conflict rather than a clean add.

I started out worried a branch-favouring resolution would silently un-ship private-media encryption, but `reservedReplacementBitIsNotAdvertised` (`bitchatTests/Services/BLENoiseReconnectPolicyTests.swift:106`) already pins that pair, so CI would catch it. I couldn't find anything asserting `.vouch`, `.prekeys` or `.groups` — that's where a quiet regression would fit.

The shape question is the one I'd rather ask about: `main` reaches runtime-toggled bits through `BLELocalIdentityStateStore.swift:16` (`localSupported.union(runtimeCapabilities)`), which is how the gateway bit works. Would routing `.wifiBulk` through `runtimeCapabilities` instead of a compile-time literal fit? It would also give the user-facing off switch nearly for free, which is the ask in my `Info.plist` comment.

bitchat/Services/BLE/BLEService.swift
```diff
@@ -1182,6 +1271,57 @@ final class BLEService: NSObject {
...
+            deliverReceivedFile: { [weak self] payload, peerID, payloadLimit in
+                self?.messageQueue.async { [weak self] in
+                    guard let self = self else { return }
+                    let packet = BitchatPacket(
+                        type: MessageType.fileTransfer.rawValue,
+                        senderID: Data(hexString: peerID.toShort().id) ?? Data(),
+                        recipientID: self.myPeerIDData,
+                        timestamp: UInt64(Date().timeIntervalSince1970 * 1000),
+                        payload: payload,
+                        signature: nil,
```
jozanek, line 1303: **Blocking on merge, as I read it** — though the fix looks small.

`BLEFileTransferHandler.authenticatedRawSenderNickname` opens with `guard packet.signature != nil else { return nil }` (`BLEFileTransferHandler.swift:478`, from #1434). A synthesized packet with `signature: nil` returns nil there, so a file delivered over Wi-Fi wouldn't arrive once this is rebased. This path also bypasses the stable-ID dedup and `commitPrivateMediaFile` / `acknowledgePrivateMedia` that private media now runs.

Would `handlePrivatePayload(_:from:timestamp:)` (`:179`) work as the ingress instead? It's the Noise-authenticated entry point, which matches a channel keyed off Noise-exchanged tokens — but you know that ingress far better than I do, so I may be missing why the synthesized packet was the easier target.

bitchat/Services/BLE/BLEService.swift
```diff
@@ -765,10 +785,72 @@ final class BLEService: NSObject {
...
+    /// Builds the Wi-Fi bulk eligibility candidate for a private send.
+    ///
+    /// Normalizes `peerID` to its short routing ID first: stable/verified
+    /// private chats address the peer by its full 64-hex Noise key, but the
+    /// Noise session, advertised capabilities, and connection state are all
+    /// keyed by the short (SHA256-derived 16-hex) ID. Resolving them with the
+    /// 64-hex key would miss the session (`hasEstablishedSession` returns
+    /// false), so Wi-Fi would never be offered even to a directly-connected
+    /// `.wifiBulk` peer.
+    private func wifiBulkSendCandidate(payloadBytes: Int, to peerID: PeerID) -> WifiBulkPolicy.SendCandidate {
```
jozanek, line 831: **Blocking on merge, as I read it** — and it's a sequencing decision rather than a defect in what's written here.

`main` routes private media through `privateMediaSendPolicy(to:)` (`BLEService.swift:1174`), whose outcomes include `.blockedDowngrade`. That gate didn't exist when this branch was cut (#1434 landed three weeks later), so nothing here routed around anything — but rebased as-is, this eligibility check would sit ahead of it, and a Wi-Fi transfer would proceed where the policy says stop.

The `discovery only; it can start negotiation but cannot pin capability or select encrypted media` framing from #1434 reads to me as permitting the offer but not the path selection — and `.wifiBulk` reads the same unauthenticated announce field, a different bit in it. Where do you see Wi-Fi selection sitting relative to that gate, and does the `0x21` proof fit here?

bitchat/Features/media/ImageUtils.swift
```diff
@@ -13,11 +14,28 @@ enum ImageUtilsError: Error {
...
+    // in production. A genuinely detailed photo at `defaultMaxDimension` now
+    // weighs well over 64 KiB, so it becomes Wi-Fi-bulk eligible to a capable
+    // direct peer while still riding BLE fragmentation for everyone else.
+    private static let targetImageBytes: Int = 200_000
     private static let maxSourceImageBytes: Int = 10 * 1024 * 1024
+    // Longest-side ceiling for shared photos. 448 px was thumbnail-tier and
+    // (together with the tiny byte budget) forced every photo below the Wi-Fi
+    // bulk threshold. 1024 px keeps a shared photo legible and lets detailed
+    // images clear 64 KiB, without approaching the 512 KiB hard cap.
+    static let defaultMaxDimension: CGFloat = 1024
```
jozanek, line 36: Raising these reverses one of the five knobs #845 set together — and because prep is recipient-unaware, it raises the budget for `sendFileBroadcast` too, which flood-relays, feeds gossip-sync replay, and can never use Wi-Fi bulk.

By your own measurements that's 40,954 B → 195,168 B, so roughly 88 → 417 fragments and ~2.6 s → ~12.5 s of airtime per broadcast photo, for peers who get nothing back from the Wi-Fi path.

Was the broadcast path in view when you picked 200 KB, or did it ride along? Either way, would you be open to splitting this into its own PR? It merges clean today, unlike the rest of the branch, and the case for a larger budget reads better argued on its own terms.

bitchat/Services/TransportConfig.swift
```diff
@@ -281,6 +281,21 @@ enum TransportConfig {
...
+    // only the sealed chunk stream moves to TCP over AWDL.
+    static let wifiBulkEnabled: Bool = true
+    // Below this size BLE fragmentation is fast enough that negotiation
+    // overhead isn't worth it.
+    static let wifiBulkMinPayloadBytes: Int = 64 * 1024
+    static let wifiBulkChunkBytes: Int = 64 * 1024
+    // Offer unanswered for this long → fall back to BLE fragmentation.
+    static let wifiBulkOfferTimeoutSeconds: TimeInterval = 10.0
+    // Hard ceiling on how long the Bonjour listener/connection may live.
+    static let wifiBulkTransferWindowSeconds: TimeInterval = 60.0
```
jozanek, line 295: I can't test this without real radios, so treat it as a question rather than a finding: when the receiver's Local Network permission is denied, does anything decline the offer?

As far as I can trace the code, the receiver browses only after accepting, `tearDownIncoming` sends nothing back, and the sender cancels its 10 s `offerTimeout` the moment the accept arrives — so both sides would wait out this 60 s window. That would turn a ~2.2 s photo into ~60 s of silence plus ~10.4 s of BLE, and I couldn't find anything that remembers the failure, so it would repeat every time. The same shape would apply to AWDL unavailable, mDNS not crossing the LAN, or a macOS firewall.

Could the receiver decline when its browser can't start, and the sender keep a short per-peer negative cache?

localPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift
```diff
@@ -2,6 +2,10 @@
 public enum FileTransferLimits {
     /// Absolute ceiling enforced for any file payload (voice, image, other).
     public static let maxPayloadBytes: Int = 1 * 1024 * 1024 // 1 MiB
+    /// Ceiling for transfers negotiated onto the peer-to-peer Wi-Fi bulk
+    /// channel. The receiver enforces it against the size in the accepted
+    /// offer; the Bluetooth path keeps `maxPayloadBytes`.
+    public static let maxWifiBulkPayloadBytes: Int = 8 * 1024 * 1024 // 8 MiB
```
jozanek, line 8: I went looking for a producer that can exceed 512 KiB and couldn't find one — `prepareImagePacket` and `prepareVoiceNotePacket` both guard at `maxImageBytes` / `maxVoiceNoteBytes`, and there's no arbitrary file-attach path. So `fitsBLECaps` looks always-true and the send-side `limit:` plumbing unreachable.

The receive side isn't unreachable, though: `shouldAccept` takes any offer up to 8 MiB, 4 concurrent, without consent, and `BLEIncomingFileValidator` has no per-mime cap — so a modified direct peer could land an 8 MiB "image/jpeg" where today's inbound ceiling is 1 MiB.

Would setting this to `FileTransferLimits.maxPayloadBytes` for now work, until a producer needs the headroom?

bitchat/Info.plist
```diff
@@ -33,12 +33,18 @@
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>LSMinimumSystemVersion</key>
 	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+	<key>NSBonjourServices</key>
+	<array>
+		<string>_bitchat-bulk._tcp</string>
```
jozanek, line 38: `docs/privacy-assessment.md:34` already names the fixed BLE service UUID as part of what lets a passive scanner "detect that the app is in use at all". This adds a second constant identifier on a longer-range radio, reachable by anyone on the subnet via mDNS since infrastructure Wi-Fi is permitted — and random per-transfer instance names hide who, not what.

The part I'd most like your read on: `beginOutgoing` publishes the listener before the offer is answered, so the record appears even on transfers that end up falling back to BLE. Would publishing only after an accept cost much? It would also shrink the window the silent-gatecrasher problem needs.

Separately, and stated as a requirement rather than a request: `PRIVACY_POLICY.md` and `privacy-assessment.md` currently describe no Wi-Fi or local-network use, and the release checklist treats announce-surface changes as privacy-surface changes — so those two docs need updating before this ships, independent of what you decide above.

bitchat/Services/WifiBulk/WifiBulkChannel.swift
```diff
@@ -0,0 +1,463 @@
...
+
+    func cancel() {
+        finished = true
+        cancelNetworkResources()
+    }
+
+    // MARK: - Connection handling
+
+    private func acceptCandidate(_ connection: NWConnection) {
+        guard !finished, authenticated == nil, candidates.count < maxCandidateConnections else {
```
jozanek, line 183: A candidate that connects and then sends nothing looks like it's never evicted: candidates are dropped on a failed frame or a socket error, and I couldn't find a per-candidate authentication deadline — the only time bound I see is the 60 s window.

If that's right, any device on the subnet can browse `_bitchat-bulk._tcp` and hold four idle sockets open, so the real peer is cancelled at this guard and the sender waits out the full window on every photo. That reads as a gap against the PR description, which says a gatecrasher "cannot produce one valid MAC and is disconnected … the listener stays up so the real peer still succeeds" — `gatecrasherIsRejected` covers the noisy case (`Data(repeating: 0xAA, count: 60)`), and the silent one is the one that costs the transfer.

Would a per-candidate auth deadline work here, so an unauthenticated socket can't hold a slot?

bitchat/Services/WifiBulk/WifiBulkMessages.swift
```diff
@@ -0,0 +1,191 @@
...
+    /// Exact byte count of the payload that will cross the channel.
+    let fileSize: UInt64
+    /// SHA-256 over the payload bytes as they cross the channel, verified by
+    /// the receiver after reassembly.
+    let payloadHash: Data
+    /// Sender's random half of the channel secret.
+    let token: Data
+    /// Random Bonjour instance name the sender publishes for this transfer.
+    /// Never derived from nickname or peer ID.
+    let serviceName: String
```
jozanek, line 67: `serviceName` is required (decode returns nil without it) and capped at 63 bytes, the Bonjour instance-name limit. With the rendezvous role fixed in the MAC (`makeClientAuthFrameBody` is `.receiverToSender` counter 0) and chunk size unnegotiated, an Android Wi-Fi Direct or iOS 26 Wi-Fi Aware peer would have to fake a Bonjour name to emit a valid offer — and neither can reliably dictate who dials.

Since the PR asks Android to speak "the same offer/response TLVs and frame format", that makes this the contract rather than an internal detail.

Would a repeatable rendezvous TLV of `[mechanism][role][locator]` plus a negotiated frame ceiling be worth it while it's still free? `parseTLVs` already skips unknowns, so today's sender emits one entry and nothing else changes. A version byte would be the other half — though I'll note that cuts against our own convention, since `BitchatFilePacket` is unversioned and `0x21` is the exception rather than the rule.

PR #1396, opened 2026-07-07
BLE background presence: pending-connect wake-on-proximity + wake-window maintenance

jackjackbits:
Re-opened for on-device testing (was #1395, merged prematurely and reverted in 0c313628).

## What

Keeps bitchat responsive to nearby bitchatters while backgrounded, at ~zero battery cost.

iOS never expires a pending `CBCentralManager.connect()`: the Bluetooth controller completes it whenever the target comes back into range, and relaunches the app via the existing state-restoration path. Previously the app-level 8s timeout cancelled every connect, throwing that wake-on-proximity capability away.

- **`BLERecentPeripheralCache`** — retains handles to recently seen/dropped peripherals (LRU 16, 15 min max age to respect BLE address rotation)
- **On backgrounding** — arm indefinite pending connects to cached peripherals, budgeted so 2 of the 6 central slots stay free for live background discovery; armed entries carry `lastConnectionAttempt == nil` so a quick background→foreground bounce can't strand them as connecting
- **Connect timeout defers while backgrounded** — discovery-driven background connects also stay pending
- **Foreground return** — cancels stale pendings (including `connecting` entries rebuilt by state restoration after a relaunch) and hands control back to the scanner/scheduler
- **Link dropped while backgrounded** — re-arms after the 3s disconnect-settle window so a peer walking away and returning wakes us again
- **Wake-window maintenance** — packet ingress while backgrounded triggers one catch-up `performMaintenance` pass (announces/spool flush/drains), since the maintenance timer is suspended with the app; rate-limited to the normal 5s cadence

Battery: pending connects live in the controller's allowlist (no scanning, no app CPU, no new timers); the catch-up pass only runs inside wake windows the radio already granted.

Known physics unchanged: two backgrounded iPhones can't *discover* each other cold (overflow-area limit) — but reconnection to recently-seen peers now works from suspension, and any foreground device or Mac still bridges everyone.

## Testing

- 7 new unit tests for the cache (eviction/expiry/budget/exclusion)
- Full suite: 1355 tests in 179 suites pass
- macOS + iOS device builds green
- Field-test logs at info level: `🌙` armed/deferred/background-wake, `🌅` foreground cleanup
- [ ] On-device walk-away/walk-back test (Mac + iPhone), pending

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1398, opened 2026-07-07
Reconnect hygiene: paced acks, persistent gift-wrap dedup, self-fragment sync fix

jackjackbits:
The last three findings from the July 7 locked-phone test sessions, all rooted in reconnect/relaunch behavior. Follows #1396/#1397.

## 1. All Nostr acks flow through the paced queue

Direct READ receipts were already throttled (`nostrReadAckInterval`, ~3/s), but DELIVERED acks and both geohash ack kinds fired immediately — reconnect redelivery produced 8 DELIVERED acks in under a second and damus rejected several with "rate-limited: you are noting too much". The read queue is now a general ack queue (`QueuedAck` enum, four cases) with unchanged pacing and unchanged per-case send code.

## 2. Processed gift-wrap event IDs persist across launches

Root cause of both the re-ack bursts *and* the "delivered ack for unknown mid" warnings: NIP-59 randomizes gift-wrap timestamps, so a since-cursor can't work and the DM subscriptions look back 24h — relays therefore redeliver the same events on **every launch**, and the in-memory dedup started empty each time. Each relaunch reprocessed old PMs (→ re-sent DELIVERED bursts) and old acks (→ phantom "unknown mid" warnings on the sender).

- New `NostrProcessedEventStore` (GossipMessageArchive pattern): JSON file under Application Support, debounced writes off the main actor, `UntilFirstUserAuthentication` file protection so it loads during locked-background restoration relaunches.
- Wired through `MessageDeduplicationService` as an optional store — nil default keeps tests hermetic; the bootstrapper opts in.
- Wiped on the existing `clearAll`/`clearNostrCaches` (panic) paths.
- The unknown-mid log drops to debug: a genuinely stale ack is expected occasionally and not actionable.

## 3. Self-fragments handed back by sync replay re-enter the sync store

The relayed self-echo (`type 32 from <own peerID>`) wasn't a missing guard — it's the deliberate RSR ttl=0 restore path. The actual bug: `BLEFragmentHandler` dropped own fragments *before* `trackPacketSeen`, and the fragment store isn't archived, so after a relaunch our sync filter never covered our own fragments and peers re-offered them **every 30s fragment round indefinitely**. Own fragments are now recorded as seen (next round's filter covers them, redelivery stops) while assembly stays skipped. Existing test updated to assert the new contract.

## Testing

- Full suite: 1355 tests in 179 suites pass; macOS + iOS device builds green.
- On-device check: after one relaunch cycle, the launch log should show no `GeoDM: recv PM` replays for old mids, at most one round of self-fragment redelivery, and DELIVERED acks spaced ~0.35s apart with no damus rate-limit rejections.

### DIFF
--- a/bitchat/Services/BLE/BLEFragmentHandler.swift
+++ b/bitchat/Services/BLE/BLEFragmentHandler.swift
@@ -33,12 +33,20 @@ final class BLEFragmentHandler {
 
     func handle(_ packet: BitchatPacket, from peerID: PeerID) {
         let env = environment
-        // Don't process our own fragments
+        guard let header = BLEFragmentHeader(packet: packet) else { return }
+
+        // Sync replay legitimately hands us our own fragments back (the RSR
+        // ttl=0 restore path): after a relaunch the fragment store starts
+        // empty, so our sync filter doesn't cover them and peers re-offer
+        // them. Record them as seen — the next round's filter then covers
+        // them and the redelivery stops — but skip assembly: we authored
+        // the original, there is nothing to reassemble.
         if peerID == env.localPeerID() {
+            if header.isBroadcastFragment {
+                env.trackPacketSeen(packet)
+            }
             return
         }
-
-        guard let header = BLEFragmentHeader(packet: packet) else { return }
 
         if header.isBroadcastFragment {
             env.trackPacketSeen(packet)
--- a/bitchat/Services/MessageDeduplicationService.swift
+++ b/bitchat/Services/MessageDeduplicationService.swift
@@ -64,6 +64,11 @@ final class LRUDeduplicationCache<Value> {
         head = 0
     }
 
+    /// Active keys in insertion order (oldest first), for persistence.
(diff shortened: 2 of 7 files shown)


jackjackbits:
Both Codex findings addressed in ba2bcb96:

**Shared ack pacer** — geohash acks ride throwaway `NostrTransport` instances (`makeGeohashNostrTransport()` per ack), so the per-instance queue never paced a burst. `Dependencies.live` now wires a process-wide `sharedAckPacer`; the default `Dependencies` init builds an isolated pacer from the same injected `scheduleAfter`, so tests keep manual control of the throttle and stay isolated from each other.

**Transient clears keep the disk record** — `clearNostrCaches()` (runs on every geohash channel switch) no longer wipes `processed-events.json`; only `clearAll()` (panic path) does. Persistence also switched from snapshot-overwrite to append-merge, serialized on the store's IO queue, so an in-memory clear between debounced flushes can't shrink the on-disk record either.

1355 tests pass, macOS + iOS builds green.

PR #1403, opened 2026-07-07
Live push-to-talk voice for DMs (streams while you talk, voice note as fallback)

jackjackbits:
## What

Holding the mic in a DM now streams the audio **live** to the peer while you talk, walkie-talkie style (~0.5 s mouth-to-ear at one hop), and still delivers a normal voice note on release. Design doc included at `docs/PUSH-TO-TALK-DESIGN.md`.

The "smart" part is that PTT is a **delivery strategy, not a mode**: the same hold-to-record gesture picks the backend automatically.

| Context | Behavior |
|---|---|
| DM peer mesh-reachable + Noise session established + toggle on | live stream **+** finalized voice note (absorbed silently into the live bubble) |
| Anything else (Nostr-only peer, no session, geohash, toggle off) | exactly today's voice note flow |

## How

**Wire** — new `NoisePayloadType.voiceFrame = 0x08` carrying `VoiceBurstPacket` (`burstID(8) + seq(u16) + flags + length-prefixed AAC frames`; START/data/END/CANCELED). A greedy packetizer enforces a 210-byte content budget so the Noise ciphertext stays inside `MessagePadding`'s 256-byte bucket → every voice packet is a single 288-byte BLE frame and **never touches the fragment scheduler** (which would let voice starve file transfers at its 2-transfer cap). Frames are fire-and-forget: no session → dropped, never queued (late audio is worthless).

**Sender** — `PTTCaptureEngine` taps `AVAudioEngine` and dual-encodes: live AAC-LC frames (16 kHz mono 16 kbps, ~130 B / 64 ms, ~15.6 pkt/s ≈ 5.3 KB/s against the mesh's ~15 KB/s per-link budget) plus an `.m4a` written with the exact `VoiceRecorder` settings. On release the m4a rides the existing `sendVoiceNote` path; its file name `voice_<burstID>.m4a` links note↔burst with zero `BitchatFilePacket`/Android changes. `VoiceRecordingViewModel` now drives a pluggable `VoiceCaptureSession` (classic recorder or live), and the composer HUD shows a pulsing red **LIVE** treatment so the sender knows audio is leaving in real time (slide-to-cancel still works but can't unplay what was heard — CANCELED tells receivers to discard).

**Receiver** — `ChatLiveVoiceCoordinator` assembles bursts: jitter-ordered delivery with a 0.5 s gap-skip, 3 s idle end, flood/size caps (8 concurrent, 384 KB, ~3× realtime rate guard). Frames persist progressively as ADTS `.aac`, so even a partial burst is a replayable voice-note bubble (with a pulsing LIVE badge while streaming). Live autoplay only when that conversation is on screen, the app is active, and the new app-info **"live voice messages"** toggle is on; otherwise it's just a bubble. Playback exclusivity goes through the existing coordinator via a small `ExclusivePlayback` protocol. When the finalized note arrives, it's **absorbed** into the live bubble (same message ID via `upsertByID`, partial file deleted) — no duplicate row, no second notification.

## Notes

- 6 new localization keys added across all 29 locales (coverage test passes).
- Traffic-analysis note: the steady ~15 pkt/s cadence during a burst is a timing fingerprint even under Noise (observers learn "someone is speaking", not content) — accepted for v1, documented in the design doc.
- Public-mesh PTT (`MessageType 0x29`, signed broadcast, relay policy) lands as a follow-up PR stacked on this.
- Foreground-only; no `audio` background mode.

## Testing

- `swift test`: full suite green (1379 tests), including new suites: burst framing round-trips/validation, packetizer budget (asserts padded ciphertext ≤ 256), ADTS header bits + Core Audio readability of the streamed file, real AAC encode→decode round-trip, and the assembly lifecycle (out-of-order delivery, absorb, cancel, empty-burst withdrawal, caps, blocked peers).
- Needs on-device verification: two-phone live DM latency, talk-over, cancel semantics, mic-permission-denied path.

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -150,6 +150,17 @@ final class ConversationUIModel: ObservableObject {
         chatViewModel.sendVoiceNote(at: url)
     }
 
+    /// Capture backend for the mic gesture: live PTT when the current DM
+    /// peer can hear it now, classic voice note otherwise.
+    func makeVoiceCaptureSession() -> VoiceCaptureSession {
+        chatViewModel.makeVoiceCaptureSession()
+    }
+
+    /// Whether this message is a live voice burst still streaming in.
+    func isLiveVoiceMessage(_ message: BitchatMessage) -> Bool {
+        chatViewModel.liveVoiceCoordinator.isLiveVoiceMessage(message)
+    }
+
     func cancelMediaSend(messageID: String) {
         chatViewModel.cancelMediaSend(messageID: messageID)
     }
--- a/bitchat/Features/voice/PTTAudioCodec.swift
+++ b/bitchat/Features/voice/PTTAudioCodec.swift
@@ -1,0 +1,175 @@
+//
+// PTTAudioCodec.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import AVFoundation
+import BitLogger
+import Foundation
+
+/// Streaming PCM -> AAC-LC encoder for live voice. Stateful (the AAC encoder
+/// carries a bit reservoir across frames); one instance per burst.
+/// Not thread-safe — confine to one queue.
+final class PTTFrameEncoder {
+    private let converter: AVAudioConverter
+    private var pendingInput: [AVAudioPCMBuffer] = []
+
+    init?() {
+        guard let pcm = PTTAudioFormat.pcmFormat,
+              let aac = PTTAudioFormat.aacFormat,
+              let converter = AVAudioConverter(from: pcm, to: aac)
+        else { return nil }
+        converter.bitRate = PTTAudioFormat.bitRate
+        self.converter = converter
+    }
+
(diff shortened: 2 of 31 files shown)


jackjackbits:
Field-test status (Mac + 2 iPhones, July 8): **DM live voice verified working end-to-end on devices** — live streaming while holding, LIVE HUD on the sender, live bubble + absorption of the finalized note on the receiver, no duplicates.

Follow-up commits on this branch since open, all from device findings:
- Codex review: peer-ID normalization via `toShort()` in capture-backend selection; the live-voice toggle now gates inbound bursts too (off = classic notes in both directions, with a test)
- iPhone capture fix: fresh `AVAudioEngine` per capture bound to the now-active audio session (input unit previously went dead with AURemoteIO -10851 / 0 Hz), one retry after route settle, and a fallback to the classic recorder within the same hold so a route glitch never costs the voice note
- Diagnostics on every previously-silent drop path (undecodable frames, toggle drops, no-session sends, quick-release holds)

⚠️ Depends on **#1405** for DMs in practice: the private-chat sheet's swipe gesture cancels the composer's mic hold within milliseconds (a standing main bug, PTT or not). Without it, DM voice recording is unreachable by gesture.

PR #1404, opened 2026-07-07
Public-mesh push-to-talk: signed live voice bursts in the mesh channel

jackjackbits:
**Stacked on #1403** (DM live PTT) — merge that first; this PR's base is `feat/ptt-dm-live`.

## What

Holding the mic in the **public mesh channel** now streams the burst live to everyone in range, walkie-talkie style, on top of the voice note that still ships on release. The design doc's phase 2.

One deliberate deviation from the original design (doc updated in this PR): the spec said public bursts would be live-only to avoid doubling bandwidth, but dropping the note broadcast would regress mixed-version meshes — old clients can't decode live bursts and would stop receiving public voice entirely, and late joiners/out-of-range peers would get nothing. So the finalized note still broadcasts as today; live receivers absorb it silently into the live bubble via the same burstID-in-filename dedup as DMs.

## How

**Wire** — `MessageType.voiceFrame = 0x29`: signed like public messages, ephemeral (mapped to no `SyncTypeFlags` bit so it never rides gossip sync), and explicitly unpadded in `BLEOutboundPacketPolicy` (padding to the 512 block would push every ~490-byte signed packet over the MTU into the fragment scheduler).

**Relay** — `RelayController` treats `voiceFrame` with the media-fragment policy: dense-graph TTL clamp (5) contains the sustained ~15 pkt/s per-talker stream, sparse graphs get full depth, and the tight 8–25 ms jitter keeps per-hop latency well inside the receiver's 350 ms jitter buffer.

**Inbound gate** — mirrors the public-message identity gate: broadcast-recipient only, 30 s freshness cap (relay stragglers/replays are not audio anyone should start hearing), and the packet signature must verify against the claimed sender's announce (registry key first, persisted-identity fallback) before anything reaches the UI. Delivered via a new `TransportEvent.publicVoiceFrameReceived` carrying the verified nickname.

**App** — `ChatLiveVoiceCoordinator` gains burst scopes (`directMessage` / `publicMesh`): public bubbles land in the mesh timeline through the normal public pipeline, autoplay only while that timeline is on screen (same app-active + toggle gates as DMs), and note absorption is scope-bound — a public note can't replace a DM burst or vice versa.

**Floor courtesy** — while someone talks live in the public channel, the composer mic tints red and pulses, with an accessibility value naming the talker (new `"%@ is speaking"` key, localized in all 29 locales). Holding still works: a decentralized mesh has no floor arbiter; the tint just discourages talk-over, exactly as designed.

## Testing

- Full suite green: **1382 tests**.
- New: relay-policy coverage for voiceFrame (sparse fragment cap + dense clamp + jitter bounds), public burst lifecycle (mesh bubble with verified nickname, talker indicator set → cleared, republish on end), broadcast-note absorption into the mesh store, and scope-binding rejection.
- Needs on-device verification: three-phone public channel (talker + direct listener + relayed listener), talk-over behavior, old-client interop (should see only the voice note).

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -12,6 +12,9 @@ final class ConversationUIModel: ObservableObject {
     @Published private(set) var currentNickname: String
     @Published private(set) var isBatchingPublic = false
     @Published private(set) var canSendMediaInCurrentContext = true
+    /// Who is talking live in the public mesh channel right now (floor
+    /// courtesy: the composer mic tints "busy" while someone holds the floor).
+    @Published private(set) var activeLiveVoiceTalker: String?
 
     private let chatViewModel: ChatViewModel
     private let privateConversationModel: PrivateConversationModel
@@ -185,6 +188,10 @@ final class ConversationUIModel: ObservableObject {
         chatViewModel.$isBatchingPublic
             .receive(on: DispatchQueue.main)
             .assign(to: &$isBatchingPublic)
+
+        chatViewModel.$activePublicVoiceTalker
+            .receive(on: DispatchQueue.main)
+            .assign(to: &$activeLiveVoiceTalker)
 
         conversations.$activeChannel
             .receive(on: DispatchQueue.main)
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -164,6 +164,11 @@ final class ConversationUIModel: ObservableObject {
         chatViewModel.liveVoiceCoordinator.isLiveVoiceMessage(message)
     }
 
+    /// Whether holding the mic right now would stream live (blue mic).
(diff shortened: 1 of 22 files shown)


jackjackbits:
Field-test status (Mac + 2 iPhones, July 8): **public-mesh live voice verified working end-to-end on devices** — signed broadcast bursts assemble, play live, and absorb the finalized note (`burst started → finalized (29 frames) → absorbed finalized note` in device logs), floor-courtesy mic tint works.

Follow-up commits since open: both Codex findings addressed (rejected voice frames no longer relay; public live bubbles append directly to the store instead of the batched pipeline), plus the merged PR-1 follow-ups.

Note: this branch also carries the DM-sheet gesture fix that made mic holds work in private chats; it's split out as **#1405** (a standing main bug worth merging first) — the identical hunk here will dedup on merge.

PR #1434, opened 2026-07-10
Encrypt private media before BLE fragmentation

jackjackbits:
## Summary

- encrypt the complete private `BitchatFilePacket` before BLE fragmentation and send the canonical cross-platform Noise inner type `0x20`
- accept prerelease iOS inner type `0x09` on receive only
- add permanent authenticated peer-state inner type `0x21`, carrying versioned canonical capability TLVs plus the peer Ed25519 key inside the established Noise channel
- treat the signed public capability bit as discovery only; it can start negotiation but cannot pin capability or select encrypted media
- bind peer state, policy, and private-file encryption to the exact manager-owned Noise session generation
- preserve an explicit one-send, signed directed raw `0x22` fallback only for unpinned released clients that cannot prove support
- retain the 256-fragment route cap, bounded admission/cancellation state, voice preparation ordering, and visible failure handling

## Rolling-upgrade behavior

| Peer state | Send behavior |
| --- | --- |
| current Noise session proves private-media support in `0x21` | encrypted Noise `0x20` |
| current session proves no support and identity was previously pinned capable | blocked as a visible downgrade |
| current session sends no proof before the bounded watchdog expires and identity is unpinned | one-send warning and explicit consent for signed directed raw `0x22` |
| no stable authenticated Noise identity | blocked before wire transmission |

The fallback is never automatic, unsigned, persisted as blanket consent, or sent alongside ciphertext. Released peers that do not understand `0x21` decrypt and ignore the unknown inner type, then follow the watchdog/consent path. New clients continue accepting the temporary prerelease iOS `0x09` receive alias while always transmitting canonical `0x20`.

## Security and lifecycle notes

- a copied public Noise key in an attacker-signed announce cannot bind the attacker's Ed25519 key or capability; the real peer's encrypted `0x21` establishes both
- malformed, duplicate, non-canonical, and unknown-version authenticated peer-state packets are rejected; unknown TLVs are skipped for forward compatibility
- one bounded authenticated echo recovers message-3/proof cross-link reordering
- automatic rekey broadcasts its exact XX message 1 instead of stranding a partial initiator session
- session insertion, promotion, removal, decrypt, capability commit, policy, and private-file encrypt share an exact generation invariant; stale callbacks/proofs fail closed
- pending private payloads wait for proof, and cancellation remains effective before approval, during preparation, in the Noise queue, and at final scheduler admission
- Nostr paths explicitly ignore mesh peer-state control messages

## Validation

## Dependencies and rollout

This PR is intentionally stacked on #1432. Land #1432 first, then retarget this PR to `main`.

Coordinate the cross-platform release in this order:

1. iOS #1432 and Android [#730](https://github.com/permissionlesstech/bitchat-android/pull/730) establish authenticated peer/session identity binding.
2. This PR and Android [#728](https://github.com/permissionlesstech/bitchat-android/pull/728) add matching authenticated `0x21` negotiation and migration policy.
3. Keep the warned fallback until minimum-supported iOS and Android clients emit authenticated bit-8 `0x21` state and the legacy population has aged out; removal is a later coordinated release.

#1437 is wire-independent and can land separately; the exact combined stack is validated above.

## Current stack

This is layer 3 of 14 and targets `codex/fix-noise-peer-binding`.

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -12,6 +12,7 @@ final class ConversationUIModel: ObservableObject {
     @Published private(set) var currentNickname: String
     @Published private(set) var isBatchingPublic = false
     @Published private(set) var canSendMediaInCurrentContext = true
+    @Published private(set) var legacyPrivateMediaConsentRequest: LegacyPrivateMediaConsentRequest?
     /// Who is talking live in the public mesh channel right now (floor
     /// courtesy: the composer mic tints "busy" while someone holds the floor).
     @Published private(set) var activeLiveVoiceTalker: String?
@@ -153,6 +154,13 @@ final class ConversationUIModel: ObservableObject {
         chatViewModel.sendVoiceNote(at: url)
     }
 
+    func resolveLegacyPrivateMediaConsent(requestID: UUID, approved: Bool) {
+        chatViewModel.resolveLegacyPrivateMediaConsent(
+            requestID: requestID,
+            approved: approved
+        )
+    }
+
     /// Capture backend for the mic gesture: live PTT when the current DM
     /// peer can hear it now, classic voice note otherwise.
     func makeVoiceCaptureSession() -> VoiceCaptureSession {
@@ -192,6 +200,10 @@ final class ConversationUIModel: ObservableObject {
         chatViewModel.$activePublicVoiceTalker
             .receive(on: DispatchQueue.main)
             .assign(to: &$activeLiveVoiceTalker)
+
+        chatViewModel.$legacyPrivateMediaConsentRequest
+            .receive(on: DispatchQueue.main)
+            .assign(to: &$legacyPrivateMediaConsentRequest)
 
         conversations.$activeChannel
             .receive(on: DispatchQueue.main)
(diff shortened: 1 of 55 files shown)


jackjackbits:
**Rebased onto main + fragment-cap regression fixed. Holding merge pending Android wire-constant confirmation.**

- Rebased onto current `main` (after #1431/#1432 landed): conflict-free, net delta is only this PR's own 9 commits, #1431/#1432 files untouched.
- **Fragment-cap fix** (`14749e94`): the blanket 256-fragment preflight regressed iOS→iOS private photos in the ~120–512 KB range. Now the 256-cap applies **only to the directed `fileTransfer` migration fallback** (current Android); encrypted `noiseEncrypted` media to capable peers uses the full receiver ceiling. `TODO(#1434)` left for negotiating an explicit per-peer fragment limit if a future Android client adopts 0x20 with a smaller reassembler.

**Blocking question before merge:** the wire constants — inner type `0x20`, capability-proof TLV `0x21`, capability bit 8, and the deliberate >64 KB Noise-message deviation — need confirmation that Android has actually deployed matching values (PR body references Android #728/#730). If they differ, cross-platform private media breaks. Once confirmed, this is ready to merge.

BitFoundation builds + tests pass; the app target needs a device/sim build to fully verify.

PR #1437, opened 2026-07-10
[HOLD: needs Android 1402] Migrate private Nostr envelopes to kind 1402 with dual-publish

jackjackbits:
## Summary

- label BitChat's proprietary relay-DM construction accurately instead of claiming NIP-17/44/59 compliance
- publish provisional kind 1402 as the primary envelope and always publish an atomic legacy kind-1059 twin during the coordinated iOS/Android migration
- accept released-iOS empty legacy inner tags and current-Android's exact recipient `p` tag only; keep kind 1402 strict
- deduplicate authenticated twins before message/ack side effects
- preserve a 24-hour delivery window across Android's 48-hour timestamp randomization plus 15 minutes of skew (72h15m total)
- recover each wire kind through an independent 500-event relay filter
- queue each 1402+1059 pair as one protected unit; retry the whole pair after either write fails, background reconnects, or sockets change
- prevent a dead relay from pinning otherwise-delivered pairs and exhausting the protected queue
- retain each relay attempt until both exact event IDs receive matching NIP-01 OK responses; a fast primary rejection still writes the legacy compatibility twin
- keep Tor fail-closed at the actual queue-flush boundary
- surface terminal/admission failure for user messages and retry control envelopes in one bounded process-wide queue
- reject malformed, misbound, and oversized nested layers before expensive decoding
- include a cryptographically valid fixture generated by Android production `NostrProtocol.createPrivateMessage` at `b7f0b33d3a267c770d3d5a65ee2d8c7e755450db`, plus the reproducible generator patch, exact command, keys, and hashes

The [official NIPs registry](https://github.com/nostr-protocol/nips) does not reserve kind 1402. This PR therefore calls the format BitChat-specific and provisional.

## Rolling-upgrade compatibility

| Sender | Recipient | Result |
| --- | --- | --- |
| new iOS | new iOS | kind 1402 is primary; the 1059 twin is collapsed before side effects |
| new iOS | old iOS/current Android | kind 1059 remains available; no calendar cutoff |
| old iOS/current Android | new iOS | legacy 1059/13/14 remains readable, including Android's exact inner recipient tag |
| duplicate/retry | new iOS | cross-format twins deduplicate; same-format retries can re-drive acknowledgements |

Legacy publication can be removed only after a coordinated supported-client decision across iOS and Android, not by local date.

## Security and privacy notes

- the outer `p` tag still exposes the recipient public key to relays
- the seal sender, inner sender, kinds, tag shape, and recipient are authenticated
- each public wrapper uses a fresh ephemeral key, but the construction does not provide forward secrecy
- kind 1402 remains provisional and may require another migration if assigned upstream
- user-authored messages are never evicted into the control retry queue; bounded control-pair eviction is whole-pair and explicitly logged

## Validation

## Current stack

This is layer 10 of 14 and targets `codex/private-media-safe-delete`.


jackjackbits:
## Review: request changes (strategic, not code quality)

**The protocol engineering here is sound**, and the honesty premise is correct: bitchat's Nostr DM construction genuinely is not NIP-17/44/59 (XChaCha20-Poly1305, 24-byte nonce, base64url, `v2:` prefix, non-NIP-44 HKDF), so relabeling it is right. The migration matrix verified correct: new→old rides an exact legacy 1059 twin that old clients' `kinds:[1059]` filters consume unchanged; old→new is accepted via the format-isolated legacy decoder including Android's exact-recipient inner `p` tag shape; atomicity covers early rejection racing writes, mid-write disconnects, stale-socket callbacks, `duplicate:` OKs, and Tor fail-closed at the flush boundary. The frozen cross-platform fixtures (released iOS 733098bb + Android) are the best part of the PR, and the doc claims match the implementation line-by-line.

### MAJOR — unilateral wire migration with no linked Android counterpart
The honesty goal (docs relabeling) is **fully achievable without any wire change** — the legacy format is already bitchat-only and authenticated. Moving to kind 1402 additionally buys:
- permanent 2× DM relay events until a "coordinated iOS/Android release" that is asserted but not linked anywhere in-repo;
- another migration if 1402 is ever assigned upstream (acknowledged in the PR body);
- ~600 new lines of atomic-batch state machine in `NostrRelayManager`, on the send path for **all** Nostr traffic (geo chat, boards, presence), not just DMs.

If Android never adopts 1402, dual-publish runs forever. **Recommend splitting:** land the docs/relabel + legacy hardening (size caps, tag validation, fixtures) now as a PR against main; gate the 1402 dual-publish on a linked, reviewed Android implementation.

### MINOR — kind 1402 makes bitchat DM recipients enumerable by a passive kind filter
Old 1059 events blended into the wider NIP-59 gift-wrap anonymity set at the filter level (content inspection could already distinguish the `v2:` payload, but kind filtering is free). During the indefinite migration window both events publish, so relays see the union of both schemes' metadata plus trivially linkable twins (same `p` tag, adjacent timestamps). The docs disclose the recipient-tag exposure but not this kind-level fingerprint / anonymity-set reduction.

### MINOR — 32 KiB envelope cap vs the 60,000-char composer
`maximumPrivateEnvelopePlaintextBytes = 32*1024` while `InputValidator.Limits.maxMessageLength = 60_000`. A ~40 KB DM that previously would be encrypted and attempted over the Nostr fallback now deterministically fails. `privateEnvelopePublicationBatchRejectsMaximumComposerPayload` enshrines the mismatch rather than aligning the two limits.

### MINOR — whole-pair OK requirement can mark delivered messages failed
`resolvePrivateEnvelopeAcknowledgement`: one rejected half retires the relay without success. If every target relay rejects the unknown kind 1402 (kind-whitelist policies) while accepting the 1059 twin, the message *reaches the recipient* but the sender sees terminal failure and the outbox re-sends. Not triggered by today's five defaults, but it couples delivery UX to ecosystem acceptance of an unregistered kind.

### MINOR / NITs
- `shouldProcessPrivatePayload`: once both format bits are recorded, a later outbox retry pair passes the gate for both copies → duplicate DELIVERED/READ acks per retry (deduped downstream; wasted traffic).
- `NostrPrivateEnvelopeRetryQueue.retry` resets `attempt` to 0 on every terminal-failure re-enqueue, so backoff never compounds (~12s loop instead of growing to 60s).
- `uiProcessedNostrEventsCap = 2000` is unchanged while the mailbox window tripled (24h→72h15m) and per-DM events doubled.

Also note: rollback after release is hard — once new clients emit 1402, reverting iOS recreates the mixed-format world.

jackjackbits:
**Split per maintainer decision.** Android has confirmed the BLE private-media constants (0x20/0x21/bit 8 — #1434 merged), but NOT the Nostr kind-1402 migration. So:

- The safe half of this PR — the docs honesty relabel + legacy kind-1059 envelope hardening + frozen cross-platform fixtures — is extracted as **#1480** against main (with doc text corrected to describe the *current* wire format, and the pieces #1451 already covered dropped).
- **This PR now holds only the wire migration** (kinds 1402/1403/1404, `bitchat-pm-v1:`, dual-publish batch machinery, dual-kind subscriptions, retry queue). It stays open, blocked on a linked Android counterpart implementation.

When Android's 1402 implementation exists, this needs a rebase over #1480 (same files) plus the review findings above (32 KiB cap vs 60k composer, whole-pair OK semantics, backoff reset, dedup cap).

jackjackbits:
**Branch trimmed to its true delta.** The +24k diff was an artifact: this branch was a frozen snapshot of the July-25 stack lineage while its base was rebased out from under it. Now rebased onto current main with only the wire-migration work: **+3,637/−716 across 19 files** (4 commits).

- Everything #1480 landed (docs relabel, legacy 1059 hardening, fixtures) is deduplicated — main's versions taken verbatim.
- The 1402 machinery is re-integrated with main's current architecture: the dual-publish batch and `acceptedPrivateEnvelopeKinds` filtering sit inside #1352's off-main verify pipeline; the cross-format twin dedup is grafted at all three delivery points of the detached-decrypt structure; the README/WHITEPAPER 1402 sections now build on the honest 1059 docs.
- Full suite green (1,909 + 215 tests).

Still **HOLD**: ships only with a linked Android kind-1402 counterpart. Known revision items for that day: 32 KiB plaintext cap vs 60k composer (test documents the conflict), retry-queue backoff reset, dedup-cap sizing vs the 72h15m window.

Chessing234 (commented):
dual-publish of 1059 and 1402 doubles the relay fingerprint of every dm until 1059 is dropped. if this is hold-for-android anyway, gate 1059 behind a compatibility flag instead of always sending both.

PR #1444, opened 2026-07-22
Location channels: locale-derived one-tap quick join

RodsF1:
## What

A "quick join" row in the location channels sheet: one tap into the region-level geohash channel around the device region's main population center. No location access — the suggestion derives from the device locale, and the tap reuses the same `LocationChannelsModel.teleport(to:)` path as typing a geohash.

## Why

First-time users of location channels face a cold start: granting location access or knowing a geohash. One tap into the region channel where people from their country tend to gather removes that hurdle.

## Review follow-up (reframed since the first cut)

The first version shipped a curated "heavily censored countries" roster. Per review, that's an editorial list that goes stale and is real App Store / geopolitical exposure. Changes:

- **No roster.** The suggestion is derived from `Locale.current.region` over one neutral rule applied to every country: ISO region → the 2-char cell of the main population center (largest metro, not always the capital: US → New York, TR → Istanbul, MM → Yangon). 219 regions covered; unmapped region codes just hide the row.
- **The 12 review-verified cells reproduce unchanged.** The table is generated by geohashing each center's coordinates; the twelve entries the roster shipped (BY, CN, CU, ER, IR, KP, MM, RU, SY, TM, VE, VN) came out identical to the hand-verified values, and near-boundary entries were checked by hand. (Cells are 11.25°×5.625°, so distinct countries can share one — Seoul and Pyongyang are both `wy`.)
- **Honest copy.** The caption states plainly: the cell is the main population center's — not "your country's channel" and not your location; the channel is public and well-known, so assume it's watched; quick join is discovery, not circumvention — it hides nothing and bypasses no blocks. All 30 locales carry the warning unsoftened.
- Rebased on current main (the string-catalog conflict and the pre-fa-rollout `LocalizationCoverageTests` failure are gone; the resolution also drops a stale `location_channels.tor.subtitle` key the old base carried, which #1486 had replaced).


jackjackbits (changes requested):
The code is in good shape — I independently verified all 12 two-character geohashes against the named population centers (all correct), and reusing `LocationChannelsModel.teleport(to:)` with no location permission is the right mechanism. The blockers are framing, not implementation:

1. **The hardcoded "censored countries" list needs reframing before this can merge.** A static Freedom House snapshot goes stale, invites disputes over inclusion/exclusion (AF, ET, SA, EG, PK are absent), and an explicit in-app roster of censored countries is real App Store / geopolitical exposure. I'd rather ship this as a neutral "popular region channels" list, or derive the suggestion from device locale — same one-tap benefit, no editorial roster to defend.

2. **Safety framing needs one honest sentence.** Funneling at-risk people into 12 well-known, enumerable rendezvous cells makes those `#g` filters the obvious thing to watch, and joining still depends on reaching relays at all (that's the Tor layer's job, not this feature's). "Quick join" is channel discovery, not circumvention — the copy shouldn't imply otherwise.

3. **Copy correctness:** "your country's region channel" is really the capital/main population center's 2-char cell — someone in Shanghai lands in Beijing's cell. The code comments say this; the UI copy should too.

4. Needs a rebase (xcstrings conflict).

Happy to iterate on the framing — the mechanism itself is solid.

RodsF1:
Reworked per your review — went with the locale-derived option. The suggestion now derives from Locale.current.region under one rule for every country: ISO region → the 2-char cell of the main population center (largest metro, not always the capital: US → New York, TR → Istanbul). 219 regions covered; unmapped codes just hide the row, so there's no roster left to defend. Your 12 verified cells reproduce unchanged from the generator, and I hand-checked the near-boundary entries (fun one: Seoul and Pyongyang both land in wy). The caption now says it plainly — the cell is the main population center's, not "your country's channel" and not your location; it's public and well-known, so assume it's watched; quick join saves typing; it doesn't hide you or bypass blocks. Same wording unsoftened in all 30 locales. Rebased on main and green. 

Lmk 
-RB

vincenzopalazzo (commented):
**Production Readiness: GO** — no blockers; one small cleanup.

Nice, conservative design: deriving the suggestion from the locale rather than location access, and reusing `teleport(to:)` (which is infallible for an internally generated geohash, so dismissing the sheet immediately is fine). The blunt "public and watched" caption is the right call.

### Issues & Suggestions

- **Cleanup:** `bitchat/Views/LocationChannelsSheet.swift` (~L255 and ~L346) — `QuickJoinSuggestion.current()` is evaluated once to gate the section and again inside `quickJoinSection` on every body re-render. Hoist it into a single `let` in the body (or a computed property) so the gate and the content can't drift if `current()` ever becomes non-deterministic.

### Summary

Mergeable once a human approves; the cleanup is optional.

---

jackjackbits (approved):
re-reviewed after the rework rounds — all asks verified resolved across the triage passes; CI green on current main. merging.

PR #1445, opened 2026-07-22
Header: add a gear button as a discoverable settings entry point

RodsF1:
## What

A gearshape button at the right end of the header icon cluster (next to the people count) that opens the App Info / settings sheet — directly on its settings pane.

## Why

Settings are currently reachable only by tapping the `bitchat/` wordmark, which reads as static text — there's no affordance telling a new person it's tappable. (The accessibility hint helps VoiceOver, but sighted people get nothing.) A gear is the universally understood entry point.

## Notes

- The sheet's pane memory is sticky (`@AppStorage`) and defaults to info, so a control labeled "settings" would land first-time openers on the info tab. `presentAppInfo(pane:)` overwrites the remembered pane before presenting; the gear passes `.settings`. The logo tap still works as a shortcut and keeps the sticky last-used pane.
- The deliberately undiscoverable triple-tap panic wipe is unchanged.
- Styling matches the neighboring header icons (size 12, secondary palette color, `headerTapTarget()`).
- The label/tooltip is a new `content.header.settings` key seeded with copies of the catalog's existing "settings" translations — localized in all 30 supported languages from the start.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -80,7 +80,13 @@ final class AppChromeModel: ObservableObject {
         showingFingerprintFor = nil
     }
 
-    func presentAppInfo() {
+    /// Presents the App Info sheet. Pass a pane when the launching control
+    /// promises one (the header gear is labeled "settings"); nil keeps the
+    /// sheet's own sticky last-used pane.
+    func presentAppInfo(pane: AppInfoView.Pane? = nil) {
+        if let pane {
+            AppInfoView.setSelectedPane(pane)
+        }
         isAppInfoPresented = true
     }
 
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -32410,7 +32410,7 @@
         "fil" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "settings"
+            "value" : "mga setting"
           }
         },
         "fr" : {
@@ -56149,7 +56149,7 @@
         "en" : { "stringUnit" : { "state" : "translated", "value" : "settings" } },
         "es" : { "stringUnit" : { "state" : "translated", "value" : "ajustes" } },
         "fa" : { "stringUnit" : { "state" : "translated", "value" : "تنظیمات" } },
-        "fil" : { "stringUnit" : { "state" : "translated", "value" : "settings" } },
+        "fil" : { "stringUnit" : { "state" : "translated", "value" : "mga setting" } },
         "fr" : { "stringUnit" : { "state" : "translated", "value" : "réglages" } },
         "he" : { "stringUnit" : { "state" : "translated", "value" : "הגדרות" } },
         "hi" : { "stringUnit" : { "state" : "translated", "value" : "सेटिंग्स" } },
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -56139,6 +56139,42 @@
         }
       }
     },
+    "content.header.settings" : {
+      "comment" : "Accessibility label and tooltip for the header gear button that opens the settings/app info sheet",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : { "stringUnit" : { "state" : "translated", "value" : "الإعدادات" } },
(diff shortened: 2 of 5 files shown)


jackjackbits:
Still wanted after the Settings|Info consolidation — on main the sheet is reachable only via the logo tap plus the conditional gateway globe, so a discoverable entry point is a real gap, and the styling matches the header icon cluster conventions. Three asks, then this merges:

1. **Open on the Settings pane.** `presentAppInfo()` only sets `isAppInfoPresented`, and `AppInfoView.selectedPane` is sticky `@AppStorage` defaulting to `.info` — so a control labeled "settings" opens the Info tab on first use (the Codex finding is correct). A `presentAppInfo(pane:)` parameter is all it needs.
2. **Rebase** — only `Localizable.xcstrings` conflicts; the header change itself is clean.
3. Nit: the description says it reuses existing settings translations, but the diff adds a new `content.header.settings` key with copied values — that's fine, just update the description to match.

RodsF1:
All three addressed: presentAppInfo(pane:) added and the gear passes .settings — it overwrites the sticky pane just before presenting, so the sheet constructs directly on settings with no flash, and the logo tap keeps the sticky behavior. Rebased (only the xcstrings conflicted, as you said). Description fixed — it's a new content.header.settings key seeded with copies of the existing settings translations; the commit message had the same wrong claim, so it's corrected there too. All checks green.

CI note, since you re-ran #1446: that app-test failure is LocalizationCoverageTests, not the flake class — the branches predate fa landing across the catalog, so a PR carrying fa keys makes every base key fail as "missing locales: fa." Rebasing onto current main cures it (#1445 proves it); #1446 will clear the same way when I rebase it after the guide-scope discussion.

Thx
RB

Chessing234 (commented):
fil is english "settings" marked translated — please don't mark untranslated locales as done. also setSelectedPane writes userdefaults behind @AppStorage; if the sheet is already in memory the pane can stay on info.

RodsF1:
Rebased onto current main (the branch had fallen 48 commits behind) and added one follow-up commit; both heads went fully green on CI, including the newer blocking SwiftLint and Periphery checks, and the PR is conflict-free again.

**Rebase (9c01cd4)** — only the two expected files conflicted:
- `Localizable.xcstrings`: rebuilt additively on top of main's newer localization work — the result is main's catalog plus the one new `content.header.settings` entry, verified by semantic JSON diff (0 other keys added/removed/changed).
- `ContentHeaderView.swift`: main's panic-wipe confirmation and newer sharing controls are kept byte-identical; only the stale "only entry point to App Info" comment was reworded, since the gear changes that.

**Follow-ups (9f2dced)**:
- The gateway globe now passes `.settings` too — its accessibility hint already promised "opens settings to turn the gateway on or off", but the action presented the sticky pane.
- @Chessing234 good catch on fil — `content.header.settings` is now "mga setting" (the form the catalog already uses in the gateway hint). Also fixed `app_info.tab.settings`, the label of the very tab these controls land on, so fil VoiceOver users don't hear "mga setting" and arrive at a tab labeled "settings". `content.alert.bluetooth_required.settings` has the same issue but refers to the iOS Settings app, so I left it for a separate pass unless you want it folded in here.
- On the @AppStorage concern: I couldn't reproduce a stale pane. The pane is written before `isAppInfoPresented` flips (both synchronous on the main actor), the sheet closure constructs `AppInfoView` fresh on presentation, and `@AppStorage` observes external defaults writes, so even an already-presented sheet switches live. Added a regression test (`presentAppInfoPaneRouting`) pinning the routing contract: a promised pane overwrites the sticky memory, a plain present never touches it, and first-open leaves the key absent so new users still land on Info.

RodsF1:
@jackjackbits this one's ready whenever you are. All three asks are on the head, plus @Chessing234's fil fix; 0 behind main, CI green including SwiftLint and Periphery.

Recap, since the description predates the last commit: `presentAppInfo(pane:)` routes both the gear and the gateway globe to `.settings`; `content.header.settings` and `app_info.tab.settings` are now "mga setting" in fil; and `presentAppInfoPaneRouting` pins the routing contract — a promised pane overwrites the sticky memory, a plain present doesn't touch it, and first open still lands on Info.

PR #1446, opened 2026-07-22
Add a "guide for sensitive areas": honest in-app OpSec primer

RodsF1:
## What

A new **SENSITIVE AREAS** section in the App Info sheet's info pane (right under "how to use"): a concise, non-alarmist safety guide for people using bitchat under surveillance. Localized into all 30 languages.

## Why

The people bitchat matters most for — activists, journalists, people in authoritarian states — are also the ones most exposed by using it wrong. The app never told them how to use it safely, or, just as importantly, **where its protections stop**. This guide fills that gap while being scrupulously honest: false confidence is the dangerous failure mode.

## The entries

Each is grounded in the actual implementation and states the limit as clearly as the protection:

- **public means watched** — mesh/geohash channel messages are signed but not encrypted; treat every public channel as monitored.
- **a geohash is not an identity** — anyone can teleport into any geohash, so a monitor is indistinguishable from a neighbor; you cannot "detect" one, and while location access is on your device keeps subscribing to your real cells at every level down to block (~150 m) — relays see those subscriptions before you ever post.
- **pick a coarse geohash** — posting pins you to the channel's cell (block ~150 m, building ~38 m), and background presence covers every level except building; prefer city or coarser, or keep location access off.
- **verify before you trust** — the lock means encrypted, not identified; "mark as verified" compares nothing on its own.
- **verified on mesh isn't verified online** — the badge is a Bluetooth/Noise concept; internet (Nostr) fallback uses a different key with no fingerprint.
- **tor hides your ip, not your words** — what Tor routing does and doesn't cover.
- **minimize what you leak** — what BLE broadcasts reveal (service UUID, signed announce with nickname + long-term keys, stable peer ID).
- **watch for elicitation** — social-engineering hygiene.
- **block is a local filter, not a shield** — receive-side only, per-identity, no notification.
- **panic wipe is local and irreversible** — that it asks to confirm first, what it erases, the two signals that tell you it worked or didn't, and what it can't recall.
- **get real training** — points to EFF Surveillance Self-Defense and the Access Now Digital Security Helpline.

## How it was built (for reviewer confidence)

Because inaccurate safety claims are worse than none, the content was produced by auditing each feature against the source, then **adversarially fact-checking every claim** to strip out any overstated protection; every entry cites the code path it rests on. Translations went through a second-pass bilingual accuracy review on the languages most central to at-risk people (Persian, Chinese, Russian, Arabic, Korean) to ensure no warning was softened or limit dropped.

## Review follow-up (jackjackbits' first pass)

1. **Location-leak understatement fixed** — the teleport and precision entries claimed automatic presence stops at "city level." Corrected in en and all 30 locales: background presence covers every regional level except building, block (~150 m) included, via `#g` subscriptions relays see before any post. The entries now also give the actionable out (keep location access off).
2. **"user" removed from every locale** — the en source had seeded it into 28 translations; five more had "username" compounds for "known handle." All reworded (person/people or account-name equivalents), verified by a scripted audit across all sensitive keys.
3. **`.vscode/launch.json` dropped** from the PR.
4. **Rebased past #1486** and re-verified the Tor and transport entries against the new behavior (stall detection, visible toggle, manual + `.onion` relays): their claims hold — fail-closed queueing, IP masking scope, and the "leave it on" advice are all still accurate — so their copy is unchanged.
5. **Maintenance contract** — open for discussion on the thread.

Minor: entries now carry stable ids; the ForEach no longer keys on `\.icon`.

Happy to adjust tone, length, or the external-resource choices to fit maintainer preferences.

---

**Follow-up: `066e76d` + `230bda7` — the panic entry had gone stale, twice flagged by @Chessing234.**

Two claims in `app_info.sensitive.panic.body` were false against current main, in the entry where being wrong costs the most. #1588 made the triple-tap raise the same "wipe all data?" dialog the settings pane uses, so "with no confirmation" invited someone under duress to tap, walk away, and have wiped nothing — the catalog already contradicted itself here, since `app_info.privacy.panic.description` says "asks to confirm". And #1431 made media deletion synchronous inside the wipe, so "keep the app open briefly" was obsolete.

The entry now names the two signals that matter under duress — the success system message and the red "wipe incomplete" banner meaning quit and reopen — reusing each locale's real wording for the banner so the guide names what the person will actually see. The title drops "instant" for "irreversible", which is both still true and the more useful warning. All 30 locales.

`230bda7` carries the same correction into the two `defaultValue:` literals in `AppInfoView.swift`, which `066e76d` had missed: the convention here is that every guide entry's default matches its catalog `en` value, 22 of 24 already did, and the two that diverged were the two just changed. Now 24 of 24.


jackjackbits (changes requested):
The content quality here is genuinely high — I checked ~11 entries against the implementation and the guide is honest, doesn't overpromise anonymity, and the fa/zh-Hans/ru spot-checks show the warnings survive translation unsoftened. That's exactly the right instinct. But a safety guide's one job is accuracy, and there are blocking issues:

1. **Understated location leak (the Codex finding is correct and must be fixed).** `app_info.sensitive.teleport.body` and `app_info.sensitive.precision.body` claim automatic presence stops at "city level." On main, `GeoChannelCoordinator.sampledRegionalGeohashes` samples every authorized regional level except `.building` — including block-level (~150 m) cells — and those geohashes reach relays as `#g` subscriptions before a person ever posts. A guide built to prevent false confidence can't ship a false-confidence claim. Both entries need correcting across all 30 locales.

2. **"user" in copy.** `app_info.sensitive.identity.body` has "you're visibly a user" (fil: "isa kang user", possibly mirrored elsewhere). Project rule: we never use "user" in any locale — say person/people.

3. **Drop `.vscode/launch.json`** — unrelated to the PR.

4. **Rebase timing:** this collides with #1486 (touches the same `AppInfoView.swift` + xcstrings, and changes the Tor/relay behavior your Tor entry describes — stall detection, manual + .onion relays). Please rebase after that lands and re-verify the Tor and relay entries against the new behavior.

5. **Maintenance contract to discuss:** 24 keys × 30 locales ≈ 720 hand-maintained, safety-critical translated strings that silently go stale with every behavior change. Before we own that, I'd like your take on trimming the in-app guide to the claims that are structurally stable and linking out (e.g. EFF SSD) for the rest.

Minor: `ForEach(..., id: \.icon)` breaks if two entries ever share an SF Symbol — give entries a stable id.

RodsF1:
On the maintenance contract: the concern is real; two of my eleven entries went stale against main before this even merged. But the entries most prone to going stale (precision, teleport, tor, transport) are also the ones doing the most protective work; trimming them to an EFF link would cut exactly what someone at risk can't get from a generic guide, bitchat-specific behavior.

Counter-proposal: keep the guide, and make staleness loud instead of silent. A small test in the spirit of LocalizationCoverageTests that pins the behavioral invariants the guide cites, background sampling excludes only building; tor defaults on and fails closed; nostr DMs use a separate un-fingerprinted key with each assertion annotated with the guide keys that state it. Change the behavior, and CI fails with "update app_info.sensitive.precision.body in all locales" instead of the guide rotting quietly. The 720 strings only need to be retranslated when behavior actually changes, and the tripwire makes that moment unmissable.

If you'd still rather trim, I'd keep the structural entries (public channels, verify, elicitation, blocking, panic, identity, help) and fold precision+teleport into one generalized entry, but I think the tripwire gets durability without giving up the substance. Happy to add the test to this PR if you like the direction.

RB

Chessing234 (commented):
i re-ran the two mechanical audits behind jack's asks against the current head, and both come back clean: all 24 new keys are present in all 30 locales with `state:"translated"`, no locale carries the english string verbatim, and a "user"-equivalent sweep across all 30 languages returns two hits, both in hebrew (`משתמש ב-tor`, `משתמשות במפתח`) where it's the verb "uses", not the noun. so ask 2 is genuinely done, not done-in-english-only.

the thing i'd raise instead is that one entry has already gone stale, which is the failure mode this pr exists to prevent.

`app_info.sensitive.panic.body` says:

> triple-tap the bitchat/ logo to wipe with no confirmation (settings has a confirmed version)

and the title is "panic wipe is local and instant". #1588 landed on 10 aug and made the triple-tap raise the same confirmation dialog the settings pane uses, plus a "all data wiped — new identity created" system message afterwards. so as written the guide tells someone under duress that the logo gesture fires immediately, when on main it now asks first — and the parenthetical distinguishing it from settings no longer distinguishes anything. that's in all 30 locales, and it's the one entry where being wrong costs the most.

worth re-checking the neighbours in the same window while you're in there: #1595 (10 aug) tightened `getEncryptionStatus` so the DM lock no longer resolves optimistically from a persisted fingerprint, and #1596 removed the public-channel screenshot broadcast. i read the `verify` and `public` entries against both and they still hold, but they're the ones i'd want re-read rather than assumed.

that's really an argument about shape rather than about this entry. the branch is 4464 catalog lines over 30 locales, it's CONFLICTING, and its merge base is 26 jul — 15 days was enough for one claim to go wrong. every rebase re-opens the whole surface because a claim can be invalidated by a commit that touches none of these files. it might land faster and stay accurate longer as two or three slices (location/geohash, transports/verification, panic+block) that each cite a small set of code paths, so a change to one area only reopens one pr. that also gives the maintenance contract question a natural answer: each slice names the files whose change should trigger a re-read.

one small thing on the copy itself, not accuracy: `app_info.sensitive.precision.body` now carries block ~150m, building ~38m, "every level except building", and the actionable out, in one paragraph. it's the densest entry in the set and it's the one a frightened person reads fastest. the correction is right; splitting the "what you leak by posting" sentence from the "what you leak by just having location on" sentence would keep both.

Chessing234 (commented):
teleport/precision look fixed for the block-level presence claim. panic.body is still wrong after #1588: the tip still says triple-tap wipes with no confirmation, but main now raises the same confirm dialog as settings. a safety guide can't ship that.

RodsF1:
@Chessing234 you were right twice on `panic.body` and I should have fixed it the first time. Done in 066e76d, both claims:

- The triple-tap raises the same "wipe all data?" dialog as settings (#1588), so "with no confirmation" was the dangerous direction to be wrong in — someone under duress could tap, walk away, and have wiped nothing. Worth noting the catalog already contradicted itself: `app_info.privacy.panic.description` on main says "asks to confirm".
- Media deletion is synchronous inside the wipe now (#1431), so "keep the app open briefly" is obsolete.

The entry now names the two signals that actually matter under duress: `system.panic.completed` for success and the `content.banner.panic_blocked` banner for retry, reusing each locale's real wording so the guide names what the person will actually see rather than paraphrasing it. Title is "panic wipe is local and irreversible" — "instant" no longer holds, and irreversible is the more useful warning anyway.

@jackjackbits — I re-audited every technical entry against current main while I was in here, and found three more that have drifted. I've deliberately **not** touched them, because fixing them is 90 more translated strings on a PR you've already questioned the maintenance cost of, and that's your call:

1. `precision.body` / `teleport.body` — "every level except building" is false after any nearby-notes action (the notes hint on the empty timeline, the notices geo tab, or `/drop`), and location notes default on. The ~38 m building cell then joins background sampling for the rest of the session. Same false-confidence class as the original block.
2. `teleport.body` — "a watcher looks identical to a neighbor" is right in conclusion, wrong in mechanism. The people list only shows people who've *sent* something, and the presence heartbeat only goes out at city precision or coarser into real cells; a reader never appears at all. The teleport mark rides only on posts and is self-declared by the sender's own app.
3. `transport.body` — this one *overstates* the danger. Inbound Nostr DMs are seal-signature-checked against the npub learned inside the Noise session and the seal still renders, and fallback only happens for favourites carrying that key. The real residual gaps are different: the QR carries the npub but `beginQRVerification` discards it, and a later `[FAVORITED]` arriving over Nostr can rebind it.

Which is exactly your maintenance-contract point, and I think it now answers itself in both directions: four entries drifted in six weeks, and every one was caught by someone reading code rather than by anything failing. That's the argument for the tripwire test I proposed in July — it turns each of these into a red CI run naming the keys to retranslate, rather than a reviewer happening to look.

So: **tripwire or trim?** If tripwire, I'll add it to this PR along with the three fixes above. If trim, I'll fold precision+teleport into one entry and cut to the structurally stable set (public channels, verify, elicitation, blocking, panic, identity, help), and the maintenance surface drops by roughly half. Either way I'd rather do it than leave this sitting.

RodsF1:
Three corrections to my comment above, one of which is a real gap I left.

**1. The fix was only half applied — now finished in 230bda7.** I said "Done in 066e76d, both claims". 066e76d corrected the catalog but not the two `defaultValue:` literals in `AppInfoView.swift`, so the English source of truth still read *"panic wipe is local and instant"* and *"triple-tap the bitchat/ logo to wipe with no confirmation … keep the app open briefly"* — exactly the two claims that commit exists to remove. The catalog wins at runtime whenever the key resolves, so nobody would have seen the old text, but it is the wrong thing to leave in the file a reader treats as authoritative, and it is what ships if the key ever fails to resolve. The convention here is that each entry's default matches its catalog `en`: 22 of 24 already did, and the only two that diverged were the two I touched. Now 24 of 24.

**2. I overstated the wording reuse.** I said the entry reuses "each locale's real wording so the guide names what the person will actually see rather than paraphrasing it". That is true for the failure banner — `content.banner.panic_blocked`'s leading phrase appears verbatim in all 30 locales — but not for the success side. The body says only "a system message means it worked" and never quotes `system.panic.completed` ("all data wiped — new identity created"). So half of that sentence paraphrases, which is the thing the sentence claims not to do. Under duress the success signal is the one someone most needs to match against the screen, so if you want it named too, say the word and I'll quote it per locale the same way.

**3. Drift-note item 2 was wrong in its mechanism.** I told you the people list "only shows people who've *sent* something … a reader never appears at all". `GeoPresenceTracker.subscribeNostrEvent` accepts kind 20001 presence as well as kind 20000 messages and calls `recordGeoParticipant` for both, before the content check — so in a city-or-coarser cell that is genuinely yours, the automatic heartbeat lists you without you posting anything. The conclusion I drew still stands (someone who only reads, or who reads through another Nostr client, never appears, so you cannot enumerate watchers) but "only people who have sent something" is wrong unless you count a heartbeat the person never chose to send. Worth getting right before that rewrite is written, not after.

Items 1 and 3 of the drift list are unaffected. The tripwire-or-trim question stands.

PR #1456, opened 2026-07-25
Guard the NoiseSessionManager incoming-handshake reuse branch instead of force-unwrapping

### LINKED ISSUE #645: Stability: Remove force unwraps and `try!` in critical paths
# Stability: Remove force unwraps and try! in critical paths

## Problem:

Force unwraps and `try!` risk crashes under rare timing or data conditions.

- `existingSession!` in `NoiseSession` can crash on races.
- Forced layer cast in `VerificationViews` can fail.
- `try!` `NSRegularExpression` can crash on invalid patterns.
- `last!` access in helpers risks empty-state crashes.

## Proposed Solution:
Replace with safeguards and fail-soft paths.

1. Swap `!` with `guard let/if let` and `log via SecureLogger`.
2. Convert `try!` to `do/catch` with safe fallbacks.
3. Audit hot paths and add minimal tests for failure modes.
4. Keep debug assertions where helpful; avoid crashing in release.

## Benefits:

Better stability and diagnostics.

- Avoids sporadic crashes.
- Improves resilience to malformed inputs.
- Clearer error logs for troubleshooting.

sankalpsthakur:
Addresses the `NoiseSessionManager` force-unwrap item in #645 (umbrella "Stability: Remove force unwraps and `try!` in critical paths"). This does **not** close #645 — it fixes one item from that list.

**What.** `handleIncomingHandshake(from:message:)` selects an incoming session via two separately-maintained locals (`shouldCreateNew`, `existingSession`) and reuses the existing one with `session = existingSession!`. This swaps the `!` for a `guard let` that logs the impossible state via `SecureLogger` and fails soft (`throw NoiseSessionError.invalidState`) instead of trapping — exactly the issue's proposed remedy ("Swap `!` with `guard let/if let` and log via SecureLogger").

**When can it be nil?** The reuse branch runs only when `shouldCreateNew == false`, which today is set only on the one path that also assigns `existingSession = existing` (an existing non-established session that isn't a 32-byte restart). Every other path (no session / established→removed / 32-byte re-init→removed) sets `shouldCreateNew = true`. So the unwrap is safe *by construction* right now.

**Why it's still unsafe.** The safety rests on an implicit, type-unchecked coupling between two mutable locals updated in separate branches. A future edit — new branch, reordered condition, refactored establishment/restart handling — can break `!shouldCreateNew ⟹ existingSession != nil` and turn a rare state combination into a `fatalError` on a critical, attacker-reachable path (untrusted inbound handshake bytes). That's precisely the "crash under rare timing or data conditions" class #645 targets.

**If the invariant is violated post-patch.** The manager logs `NoiseSessionManager: reuse branch reached with no session …` (`SecureLogger.error`, category `.session`) and throws `NoiseSessionError.invalidState`. The sole caller, `NoiseEncryptionService`, already calls this with `try`, and the function already throws (`.alreadyEstablished`, plus re-thrown handshake failures) — so the new throw travels an existing, handled path. The peer simply retries; no crash, no session-state corruption.

**Behavior preserved.** New handshake w/ no session → responder created & processes. Established session → removed, fresh responder before processing. Unfinished + 32-byte re-init → replaced. Unfinished + continuation → same session reused (the former force-unwrap branch — identical behavior, `existingSession` non-nil). Processing failure → session removed, error propagated, `onSessionFailed` once. No changes to the Noise wire protocol, handshake messages, key derivation, retry semantics, or barrier-queue synchronization.

**Tests.** New `bitchatTests/Noise/NoiseSessionManagerStateTransitionTests.swift` (Swift Testing) drives the selection state machine through the existing `#if DEBUG` `sessionFactory` hook with scripted `NoiseSession` doubles: (1) create; (2) reuse, no 2nd responder; (3) replace established; (4) restart replace; (5) failure cleanup + single `onSessionFailed`. Scope note: the nil case the former `!` would trap on is unreachable through the public API by construction, so these are characterization/regression guards that lock in the create/reuse/replace/fail invariant — a future refactor breaking the coupling now fails a test (and, at runtime, fails a handshake softly) rather than crashing.

**Validation.** macOS no-signing build: `** BUILD SUCCEEDED **`. iOS simulator (iPhone 16 Pro), new suite: `** TEST SUCCEEDED **`, 5/5 passed. `git diff --check` clean.

**AI/LLM disclosure.** This change was drafted with the assistance of an AI coding agent (Claude Code). A human maintainer reviewed the diff, the nil-reachability reasoning, and the fail-soft error choice, and ran the macOS build and iOS test suite locally before submission.

### DIFF
--- a/bitchat/Noise/NoiseSessionManager.swift
+++ b/bitchat/Noise/NoiseSessionManager.swift
@@ -132,7 +132,15 @@ final class NoiseSessionManager {
                 sessions[peerID] = newSession
                 session = newSession
             } else {
-                session = existingSession!
+                // `shouldCreateNew` is only false after the branch above assigned
+                // `existingSession`, so this should always succeed. Guard instead of
+                // force-unwrapping so a future change that breaks that coupling fails
+                // this handshake softly rather than crashing a critical path.
+                guard let reusableSession = existingSession else {
+                    SecureLogger.error("NoiseSessionManager: reuse branch reached with no session for \(peerID); rejecting handshake", category: .session)
+                    throw NoiseSessionError.invalidState
+                }
+                session = reusableSession
             }
             
             // Process the handshake message within the synchronized block
--- a/bitchatTests/Noise/NoiseSessionManagerStateTransitionTests.swift
+++ b/bitchatTests/Noise/NoiseSessionManagerStateTransitionTests.swift
@@ -1,0 +1,276 @@
+//
+// NoiseSessionManagerStateTransitionTests.swift
+// bitchatTests
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import CryptoKit
+import Foundation
+import Testing
+import BitFoundation
+
+@testable import bitchat
+
+/// State-transition coverage for `NoiseSessionManager.handleIncomingHandshake`.
+///
+/// These tests drive the incoming-session *selection* state machine — create,
+/// reuse, replace, fail — through the `#if DEBUG` `sessionFactory` hook using
+/// scripted `NoiseSession` doubles, so each branch can be exercised deterministically
+/// without a real Noise handshake.
+///
+/// They characterize the invariant that used to sit behind the `existingSession!`
+/// force unwrap in the reuse branch: whenever the manager decides to *reuse*, a
+/// session is present; whenever it decides to *create* or *replace*, exactly one new
+/// responder is minted before the message is processed. The nil case that the former
+/// `!` would have trapped on is unreachable through the public API by construction,
+/// so these are regression guards for the state machine, not a reproduction of a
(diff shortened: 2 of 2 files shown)


jackjackbits:
Thanks for the careful writeup — the reasoning and the disclosure in the PR body are appreciated. Closing as superseded: the force-unwrap this guards was removed by #1432, which rewrote `handleIncomingHandshake` — the reuse branch is now a plain `if let existing` chain with generation tracking and quarantine-then-rollback semantics (see `NoiseSessionManager.swift` on current main), which is why this branch shows conflicts.

That also means the state-transition tests here pin the old replace-on-handshake semantics that #1463/#1464 deliberately changed, so they can't be salvaged as-is. But the test *idea* is genuinely valuable: scripted state-machine characterization tests targeting the new create/reuse/quarantine/rollback flow (the `sessionFactory` injection hook still exists for exactly this) would be a very welcome follow-up PR if you're up for it.

PR #1460, opened 2026-07-25
Keep open DMs alive across Bluetooth settings

jackjackbits:
## Summary

- keep the selected private conversation stable while SwiftUI reconciles the sheet during scene/background and Bluetooth-alert transitions
- present the Bluetooth alert from the already-presented people/DM sheet instead of forcing a competing root presentation
- defer alert presentation while the app is inactive or another modal owns presentation
- preserve normal explicit sheet dismissal when the app is active and alert-free

## User-visible fix

Opening Settings to toggle Bluetooth no longer destroys the open DM view or leaves incoming messages and read receipts updating a different replacement window.

This is presentation-only: no transport, wire format, conversation-store, delivery-status, Android, or older-client behavior changes.

## Validation

## Current stack

This is layer 11 of 14 and targets `codex/fix-nostr-envelope-labeling`.

### DIFF
--- a/bitchat/Views/ContentSheetViews.swift
+++ b/bitchat/Views/ContentSheetViews.swift
@@ -6,6 +6,18 @@ import UIKit
 import AppKit
 #endif
 
+struct ContentPeopleSheetModalPresentationState {
+    var isImagePreviewPresented = false
+    var isVerificationSheetPresented = false
+    var isMediaPickerPresented = false
+
+    var hasPresentation: Bool {
+        isImagePreviewPresented
+            || isVerificationSheetPresented
+            || isMediaPickerPresented
+    }
+}
+
 struct ContentPeopleSheetView: View {
     @EnvironmentObject private var appChromeModel: AppChromeModel
     @EnvironmentObject private var privateConversationModel: PrivateConversationModel
@@ -24,6 +36,7 @@ struct ContentPeopleSheetView: View {
     var isTextFieldFocused: FocusState<Bool>.Binding
     @ObservedObject var voiceRecordingVM: VoiceRecordingViewModel
     @Binding var autocompleteDebounceTimer: Timer?
+    @State private var showVerifySheet = false
     @ThemedPalette private var palette
 
     let headerHeight: CGFloat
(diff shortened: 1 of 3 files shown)


jackjackbits:
## Review: merge with nits (1 incomplete-fix finding)

Confirmed not superseded — main still has the bug at `ContentView.swift:107-115` (unconditional `endConversation()` in the sheet setter) and `:222` (alert bound directly to `$appChromeModel.showBluetoothAlert`). The mechanism and the fix both check out, and the in-sheet fingerprint view is correctly excluded from the modal-state struct (it's a `navigationDestination` push, not a modal).

### MAJOR — the same bug is still reachable via the voice-recording alert
`ContentView.swift` ~L317: `.alert("Recording Error", isPresented: $voiceRecordingVM.showAlert)` is still presented unconditionally from root while the people/DM sheet is up — and voice recording happens *inside* the DM sheet (`voiceRecordingVM` is passed into `ContentPeopleSheetView`).

By this PR's own verified premise (a root alert force-dismisses the presented sheet), a recording error — e.g. mic permission denied — while a DM is open dismisses the sheet, and the new guard does **not** protect it: it checks only `showBluetoothAlert`. So the DM is still destroyed via that path. The PR already added `isVoiceAlertPresented` to the root guard, so the awareness is there; it just defers the Bluetooth alert behind it and not the voice one.

Cheap fix: include `voiceRecordingVM.showAlert` in the `endConversation()` guard, or gate/relocate the voice alert the same way the Bluetooth alert is.

### MINOR — `showSidebar = false` still runs unconditionally
New reconciliation path ~L205-216. Consequences: a people-list-only sheet isn't preserved across the same reconciliation this PR fixes; and with a DM open the flag is silently cleared, so a later in-DM "back" lands outside the sheet instead of on the people list.

### MINOR — `!showBluetoothAlert` is a proxy for "dismissal caused by reconciliation"
Any sheet dismissal while that flag is true is swallowed. Mostly unreachable on iOS (the visible alert blocks sheet interaction), but if the alert ever fails to present — deferred behind an inner modal, or a UIKit presentation race — an explicit swipe-down snaps back until the alert resolves.

### MINOR — `showVerifySheet` state lifetime change
Hoisting the `@State` from `ContentPeopleListView` into `ContentPeopleSheetView` (`ContentSheetViews.swift:41`) means the flag now outlives its presenting anchor. If a DM opens programmatically while the verify sheet is up, the anchor leaves the hierarchy with the flag still set; depending on whether SwiftUI writes back false, `hasModalPresentation` could stay true and suppress the sheet-level Bluetooth alert indefinitely.

### NIT
Two identical Bluetooth alert definitions now exist (ContentView ~L327 and ContentSheetViews ~L236-246) — a future copy change must be made twice. Consider a shared modifier.

**Tests:** the four added cases are largely tautological (construct the OR-struct with one flag set, assert `hasPresentation`). Nothing covers the actual fix — the `endConversation()` guard or sheet survival. Understandable given SwiftUI limits, but it means the headline behavior rests on device testing alone.

Copy rule: clean (no new UI strings; reuses existing keys).

PR #1461, opened 2026-07-25
Keep DMs canonical across transport aliases

jackjackbits:
## Summary

- recognize the authenticated full Noise-key account ID and its derived short mesh ID as one conversation
- keep the connected short ID canonical, then hand the open conversation to the stable ID when the peer goes offline
- migrate messages and the selected DM so the already-open sheet updates instead of leaving a stale window behind
- scope delivered/read acknowledgements and durable removal tombstones to authenticated peer aliases, so a colliding message ID for another peer cannot be cleared
- preserve destination-wins semantics when the aliases already contain the same message ID

## User-visible fix

A DM received through the account/Nostr path now updates the already-open mesh DM, including delivery/read checks, instead of creating a second conversation that is visible only after reopening the notification target.

## Compatibility and security

- no wire-format or capability change; Android and older clients are unaffected
- aliases are derived only from the authenticated Noise key, never nickname matching or unrelated-chat scans
- peer-scoped acknowledgement tombstones remain safe across protected-data cold-load recovery

## Validation

## Current stack

This is layer 12 of 14 and targets `codex/cleanup-dm-window`.

### DIFF
--- a/bitchat/Services/Courier/MessageOutboxStore.swift
+++ b/bitchat/Services/Courier/MessageOutboxStore.swift
@@ -110,6 +110,10 @@ final class MessageOutboxStore {
     /// Delivery/read acknowledgments received before a deferred cold-load
     /// reveals the durable queue. Applied to every merge before persistence.
     private var pendingRemovalMessageIDs = Set<String>()
+    /// Peer-scoped acknowledgments received before a deferred cold-load
+    /// reveals the durable queue. Unlike the legacy global tombstones above,
+    /// these must not remove a colliding message ID queued for another peer.
+    private var pendingScopedRemovalMessageIDs: [PeerID: Set<String>] = [:]
     private var recoveryHandler: (@MainActor (Snapshot) -> Void)?
     /// Recovery loaded durable state that MessageRouter has not merged yet.
     /// While true, router saves must union with `cachedSnapshot` instead of
@@ -195,7 +199,7 @@ final class MessageOutboxStore {
                 ? (pendingSnapshot ?? [:])
                 : Self.merge(durable, pendingSnapshot ?? [:]))
             diskState = .loaded
-            if pendingSnapshot != nil || !pendingRemovalMessageIDs.isEmpty {
+            if pendingSnapshot != nil || hasPendingRemovalsLocked {
                 if persistSnapshotAndClearRemovalsLocked(cachedSnapshot) {
                     pendingSnapshot = nil
                     pendingSnapshotIsAuthoritative = false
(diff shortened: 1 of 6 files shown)


jackjackbits:
## Review: merge with nits

**Alias derivation verified sound.** Mesh peer IDs are provably noise-key-derived (`BLEService` sets `myPeerID = PeerID(str: fingerprint.prefix(16))`, matching `PeerID(publicKey:)`), so `toShort()` is an authenticated alias rather than a guess. GeoDM convKeys (`nostr_` + 16 hex) have `noiseKey == nil`, so geohash DMs never enter the alias/outbox-clearing path — no Nostr-pubkey-as-noise-key confusion. Threading is consistent (coordinator + MessageRouter are `@MainActor`; the store mutates all four snapshots under its own lock; scoped tombstones are applied in every load/save branch and cleared only after a successful write). The `BitchatMessage` rewrite in `consolidateAccountConversationAliases` passes every stored property — nothing dropped. Security is a net improvement: ack-driven removal is now scoped to aliases derived from the authenticated sender, strictly narrowing what a peer can clear.

### MINOR — the stated invariant only holds on the Nostr path
`ChatDeliveryCoordinator.swift:57`: the mesh ack path still calls the global `messageRouter.markDelivered(messageID)`, which clears a colliding message ID from every peer's queue and records a global tombstone. The PR body's claim that delivered/read acks are "scoped to authenticated peer aliases" is therefore true only for Nostr/account acks — a mesh peer's delivery ack can still erase another peer's queued message with the same ID. Not a regression, but the description overstates the coverage.

### MINOR — transport flips can ping-pong the canonical ID
`consolidateAccountConversationAliases` runs only on Nostr-path events and uses `isPeerConnected` (no retention window) to pick the target. A Nostr ack arriving during the BLE reachability-retention window (peer briefly "not connected" but mesh-active) migrates the whole conversation and the open sheet to the stable ID; the next mesh DM migrates it back. Self-healing and correct, but each flip rewrites every inbound message via per-message `upsertPrivateMessage` — O(n) store churn per flip on long threads.

### NITs
- `MessageRouter.markDelivered(_:for:)`: `recordRemoval` + `persistOutbox()` run unconditionally, so every authenticated stale/unknown ack costs a tombstone insert plus an encrypted-file write.
- `ChatViewModel.markMessageDelivered(_:for:)` (scoped overload) doesn't call `mediaTransferCoordinator.confirmPrivateMediaDelivery`, unlike the mesh-path overload — so a mesh-sent private media message acked over the Nostr account path never confirms media bookkeeping. Pre-existing gap, but the two same-named overloads now have silently different side effects; worth unifying or commenting.
- `scopedAckWhileColdLoadIsLockedPreventsOnlyTargetPeerResurrection` leans on two `await Task.yield()` to let the recovery handler's MainActor hop run — mildly flake-prone.
- `MockChatPrivateConversationContext.handOffSelectedPrivateChat` omits production's `|| selectedPrivateChatPeer == newPeerID` re-target branch.

Tests are genuinely good here — the store/router tests use the real store with simulated protected-data denial and assert end state through persistence and relaunch, rather than mirroring the implementation.

PR #1462, opened 2026-07-25
Retry unacknowledged DMs after Noise replacement

jackjackbits:
## Summary

- retain a securely transmitted private message until an authenticated delivery/read transition accepts it
- retry only messages that actually left through the replaced Noise session when the peer authenticates again
- bind receipt terminalization to the authenticated peer and its stable/ephemeral aliases
- preserve FIFO ordering across alias-split outboxes and prevent synchronous acknowledgements from resurrecting retained messages
- bound retries by the existing attempt cap, TTL, and per-peer outbox limit, with visible terminal failure
- replace fixed wall-clock polling in two VoiceRecorder concurrency tests with bounded async gates, eliminating scheduler-starvation false failures while retaining a five-second regression timeout

## User-visible fix

A DM sent into a stale local Noise session no longer disappears when the remote app has restarted. The same message ID is retried after the replacement handshake, so the receiver deduplicates it and the sender can advance to delivered/read.

## Compatibility

- no wire-format, message-type, or capability change
- Android and older iOS clients continue receiving the ordinary encrypted private-message packet
- normal first-handshake sends remain owned by BLE and are deliberately excluded from the router retry, preventing duplicate delivery to legacy peers
- remote receipts cannot clear retry state belonging to another peer, even if message IDs collide

## Validation

- VoiceRecorder async-gate validation: 100/100 focused stress iterations passed; the exact #1462 parallel gate and final strict one-worker gate each passed all 1,911 Swift Testing cases
## Current stack

This is layer 13 of 14 and targets `codex/cleanup-dm-alias`.

### DIFF
--- a/bitchat/App/ConversationStore.swift
+++ b/bitchat/App/ConversationStore.swift
@@ -442,6 +442,40 @@ final class ConversationStore: ObservableObject {
     @discardableResult
     func setDeliveryStatus(_ status: DeliveryStatus, forMessageID messageID: String) -> Bool {
         guard let ids = conversationIDsByMessageID[messageID] else { return false }
+        return applyDeliveryStatus(status, forMessageID: messageID, among: ids)
+    }
+
+    /// Applies an authenticated delivery/read receipt only to the supplied
+    /// direct-conversation aliases. A colliding message ID in another peer's
+    /// conversation (or a public timeline) must not inherit the receipt.
+    ///
+    /// Stable and ephemeral aliases can temporarily hold the same message
+    /// instance during handoff. The shared helper republishes every targeted
+    /// alias even when the first mutation already changed that instance.
+    @discardableResult
+    func setDeliveryStatus(
+        _ status: DeliveryStatus,
+        forMessageID messageID: String,
+        inDirectPeerAliases peerIDs: Set<PeerID>
+    ) -> Bool {
+        guard !peerIDs.isEmpty,
+              let indexedIDs = conversationIDsByMessageID[messageID] else {
+            return false
+        }
+        let allowedIDs = Set(peerIDs.map { ConversationID.directPeer($0) })
+        return applyDeliveryStatus(
+            status,
+            forMessageID: messageID,
+            among: indexedIDs.intersection(allowedIDs)
+        )
+    }
+
+    private func applyDeliveryStatus(
+        _ status: DeliveryStatus,
+        forMessageID messageID: String,
+        among ids: Set<ConversationID>
+    ) -> Bool {
+        guard !ids.isEmpty else { return false }
         var applied = false
         var skipped: [ConversationID] = []
         for id in ids {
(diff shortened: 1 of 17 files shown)


jackjackbits:
## Review: request changes (1 major)

The peer-scoping work is a genuine security improvement (peer-bound receipts close a cross-conversation clear/promote vector; bridge dedup now stores hashes rather than raw IDs; retries reuse the message ID so receivers dedup), threading is clean (no new main↔bleQueue sync cycles), and the router tests are strong.

### MAJOR — mesh acks can no longer clear the durable outbox after relaunch
`bitchat/ViewModels/ChatDeliveryCoordinator.swift:186` (`updateAcknowledgedMessageDeliveryStatus`).

The new gate requires `context.isOutgoingPrivateMessage(messageID, toAny: aliases)`, which checks conversation-store presence. But conversations are **in-memory only**, while `MessageOutboxStore` is explicitly durable across app kills — that's its entire purpose.

Failure scenario: DM queued/sent → app force-quit → relaunch restores the outbox but **not** the conversation → peer authenticates → retry delivers the message → peer's `.delivered` Noise ack arrives → `isOutgoingPrivateMessage` returns false → ack discarded (logged as "unknown message" at `ChatTransportEventCoordinator.swift:468`) → `router.markDelivered` never runs → the message is re-sent on every flush/auth event, the receiver re-acks each time (all rejected), until the 8-attempt cap drops it as **failed despite having been delivered**.

The old code called `markMessageDelivered(messageID)` *before* the store check, so the first ack always cleared retention.

That the gate is stricter than intended is confirmed by the inconsistency with the Nostr path: `ChatPrivateConversationCoordinator.handleDelivered` still calls `context.markMessageDelivered(messageID, for: aliases)` unconditionally before its conversation lookup.

**Fix:** call `markMessageDelivered(messageID, from: peerIDAliases)` even when the conversation lookup fails — the router-side clear is already peer-scoped and safe. Gate only the *UI status transition* on conversation presence.

**Test gap:** no coverage for an ack arriving for a durable outbox entry whose conversation is absent post-relaunch — the tests' `seedPrivateChat` setup sidesteps exactly this path.

### MINOR — dead but dangerous parameter
`MessageRouter.swift:412,424-451`: `clearRetainedMessage(_:allowedPeerIDs:)` is only ever called with `nil`. The parameter's filter branch pairs a scoped in-memory removal with an **unscoped** `outboxStore?.recordRemoval(messageID:)` tombstone, so any future caller passing aliases silently reintroduces the cross-peer tombstone bug Codex flagged. Please delete the parameter.

### MINOR — alias coverage gap
Retry/ack scoping uses {ephemeral peerID, live-session noise key}. An outbox entry keyed under a peer's *previous* ephemeral ID (pre-rotation) is neither retried nor cleared, lingering to the 24h TTL then surfacing `onMessageDropped`. Bounded and self-healing, but worth a comment/test.

### NITs
- `PeerMessageKey`'s manual `==`/`hash(into:)` duplicate synthesis — delete.
- Whitespace-only hunks in `NoiseEncryptionService.swift`; the VoiceRecorder test deflake is unrelated scope creep (declared in the body, but still).

Behavior change worth confirming: peers whose clients never send delivery acks were previously trusted outright and are now re-sent up to 8× across reconnects before being marked failed.

PR #1463, opened 2026-07-25
Make ordinary Noise reconnects atomic and race-safe

jackjackbits:
## Summary

- make cached ordinary Noise XX reconnect preparation, retirement, and handoff atomic
- keep an established transport receive-only while an inbound replacement handshake proves identity, then either promote the new generation or restore the old one
- deterministically resolve crossed initiators without sending stale message 1 packets
- add bounded timeout, collision-grace, rollback, cooldown, and one-shot recovery behavior
- clear local Noise transport state when CoreBluetooth powers off

## Compatibility

This uses the existing ordinary Noise XX wire format only. It does not advertise or act on the reserved non-destructive-replacement capability bit, so deployed iOS and Android clients continue to interoperate without understanding a new handshake mode.

## Validation

## Current stack

This is layer 4 of 14 and targets `codex/fix-private-media-e2e`.

---
**Known tradeoff (disclosed per review):** handshake message 1 is unauthenticated, so a spoofed 32-byte msg1 quarantines the victim's established transport — outbound pauses for up to `ordinaryResponderHandshakeTimeout` (20s), then rollback + forced rekey, repeatable after the 60s cooldown (~25% duty cycle per peer pair). This is deliberate and bounded: main is strictly worse (the session is destroyed outright by the same packet), and the cooldown/no-extension bounds are test-pinned.

**Review fix included:** timeout-restores no longer drain outbound queues under potentially-discarded keys; draining defers until the convergence retry completes (restore reason plumbed as `NoiseSessionRestoreReason`), with a deterministic interleaving test (`timeoutRestoredSessionDefersQueueDrainUntilConvergence`). Identity-mismatch restores drain immediately, unchanged.

### DIFF
--- a/bitchat/Noise/NoiseSecurityConstants.swift
+++ b/bitchat/Noise/NoiseSecurityConstants.swift
@@ -36,6 +36,10 @@ enum NoiseSecurityConstants {
 
     // Noise XX message 1 contains only the initiator's 32-byte ephemeral key.
     static let xxInitialMessageSize = 32
+
+    // Bounds the receive-only rollback quarantine created by an unauthenticated
+    // inbound message 1. A lost message 3 must not strand outbound traffic.
+    static let ordinaryResponderHandshakeTimeout: TimeInterval = 20
     
     // Session timeout - sessions older than this should be renegotiated
     static let sessionTimeout: TimeInterval = 86400 // 24 hours
--- a/bitchat/Noise/NoiseSecurityConstants.swift
+++ b/bitchat/Noise/NoiseSecurityConstants.swift
@@ -37,9 +37,29 @@ enum NoiseSecurityConstants {
     // Noise XX message 1 contains only the initiator's 32-byte ephemeral key.
     static let xxInitialMessageSize = 32
 
+    // Bounds an ordinary initiator whose message 1 or 2 is lost.
+    static let ordinaryHandshakeTimeout: TimeInterval = 10
+
     // Bounds the receive-only rollback quarantine created by an unauthenticated
     // inbound message 1. A lost message 3 must not strand outbound traffic.
     static let ordinaryResponderHandshakeTimeout: TimeInterval = 20
+
+    // A released client may immediately retry after both crossed initiators
+    // yielded. Give that unilateral retry a brief head start before the
+    // patched side spends its one bounded recovery.
+    static let handshakeCollisionRecoveryDelay: TimeInterval = 0.2
+
+    // Rate-limited recovery remains actionable without spinning.
+    static let handshakeRateLimitRecoveryDelay: TimeInterval = 60
+
+    // Covers only reordering between a winning message 3 and the losing
+    // crossed message 1.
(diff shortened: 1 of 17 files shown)


jackjackbits:
## Review: request changes (1 major)

**The atomicity design is sound** — quarantine/promote/rollback is genuinely race-safe, the crossed-initiator tie-break is a consistent total order, deadlock analysis found no reverse sync edge (lock chain stays one-directional: bleQueue→collectionsQueue, bleQueue→managerQueue, messageQueue→{managerQueue, collectionsQueue}), and the tests exercise real interleavings rather than happy paths. The races it closes are real on main.

### MAJOR — timeout-rollback drains outbound queues under discarded keys
`BLEService.swift` `onSessionRestoredWithGeneration` + `NoiseSessionManager` responder-timeout body.

Scenario: peer restarts and sends msg1 → we quarantine the old transport and reply msg2 → **peer processes msg2 and establishes new keys**, but msg3 is lost → our 20s responder timeout restores the quarantined OLD generation and requests convergence recovery (both dispatched `global().async` → messageQueue, unordered). If the restore handler wins the race, `handleNoisePeerAuthenticated(old generation)` passes its `withCurrentSessionGeneration` guard and drains `sendPendingMessagesAfterHandshake` + typed-payload queues **encrypted under keys the remote no longer holds** — DMs, read receipts and invites are dequeued and silently undecryptable.

This is the same issue as the unaddressed Codex P2 inline comment on `BLEService.swift:4745`, independently re-derived.

**Fix direction:** on *timeout*-restores, defer queue draining until the mandatory convergence retry completes. Terminal identity-mismatch restores are safe to drain immediately (the counterpart never established), so gate on the restore reason rather than suppressing draining wholesale.

**Test gap:** `failedInboundReconnectRestoresAndDrainsTypedPayloadQueue` only covers the identity-mismatch restore — the safe case. The timeout-restore drain race is untested.

### MINOR — undisclosed bounded DoS worth stating in the PR body
Handshake packets are unauthenticated, so a spoofed msg1 quarantines an established transport: `canDeliverSecurely` goes false and outbound pauses for up to `ordinaryResponderHandshakeTimeout` (20s), repeatable after the 60s cooldown — roughly a 25% duty cycle per peer pair. This is a deliberate tradeoff (main is *worse*: the session is destroyed outright; the base's candidate model had zero outbound impact) and the cooldown bounds are test-pinned. It just isn't disclosed anywhere — please note it in the description.

### MINOR — post-rollback cooldown can drop a genuine restart
If the one-shot recovery retry is cancelled (`isPeerReachable` false) or its msg1 is lost, restored-established + cooldown silently ignores legitimate fresh initiations for up to 60s.

### NIT
`handleNoiseHandshake`'s wasEstablished/isEstablished sandwich is three separate managerQueue syncs on a concurrent queue — two concurrent handshake packets can skip or spuriously fire `noteNoiseSessionCleared`. Fails closed today, but the transition belongs inside the manager's barrier.

Note this layer should land together with #1464 — see that PR.

PR #1464, opened 2026-07-25
Preserve early Noise ciphertext across reconnect promotion

jackjackbits:
## Summary

- serialize Noise handshake promotion/restoration with BLE generation-bound transport state
- defer a small bounded set of ciphertext packets that arrive before the exact authenticated generation is ready, then retry once after promotion or rollback
- preserve receive nonces by checking transport readiness and authorization before decrypt
- make malformed, replayed, oversized, forged, and rate-limited ciphertext drop-only so attacker-controlled bytes cannot evict a healthy session
- make same-generation rollback drain deferred ciphertext and pending typed/private-message queues once without repeating capability proof or announce side effects

## Bounds and compatibility

The deferred buffer is capped at 4 packets per peer, 32 globally, the maximum valid private-file ciphertext plus 256 KiB total, and the ordinary responder timeout. This changes only local ordering and failure handling: there are no new message types, payload values, capability bits, encodings, or Android/legacy-visible wire changes.

## Validation

## Current stack

This is layer 5 of 14 and targets `codex/noise-reconnect-atomic-recovery`.

### DIFF
--- a/bitchat/Noise/NoiseSessionManager.swift
+++ b/bitchat/Noise/NoiseSessionManager.swift
@@ -155,6 +155,16 @@ final class NoiseSessionManager {
         }
     }
 
+    /// Whether this peer has an inbound ordinary XX responder that still
+    /// needs message 3 before its receive keys can become authoritative.
+    func isAwaitingResponderHandshakeCompletion(for peerID: PeerID) -> Bool {
+        managerQueue.sync {
+            guard let session = sessions[peerID] else { return false }
+            return session.role == .responder
+                && session.getState() == .handshaking
+        }
+    }
+
     /// Transfers one bounded recovery generation to whatever ordinary XX
     /// handshake currently owns the peer, or starts the generation's single
     /// retry. The request token prevents stale transport callbacks from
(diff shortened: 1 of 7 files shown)


jackjackbits:
## Review: merge with nits — **but land as a unit with #1463**

Code and tests are sound. Verified explicitly: no ABBA deadlock (lock chain is messageQueue(barrier) → managerQueue → {collectionsQueue | rate-limiter leaf}; no `collectionsQueue.sync` block calls into noiseService/`messageQueue.sync`/`main.sync`; `deferredLock` is never held across call-outs and drains decrypt outside the lock); no nonce desync (readiness + `authorizeDecrypt` run inside `managerQueue.sync` *before* `session.decrypt`, so rejection throws before a nonce is consumed — tests prove rejected ciphertext still decrypts later and spends no rate budget); no replay hazard (buffer removed under lock before retry, `isDeferredRetry` prevents re-defer loops, panic wipes the buffer twice so pre-panic ciphertext can't replay into the post-panic identity).

### MAJOR (structural, not a code defect) — this PR repairs its own base
Roughly half of it fixes races introduced by #1463: `transportGenerationNotReady` exists only because #1463 added generation-bound transport state, and this PR revises #1463's semantics (removes `noteNoiseSessionCleared` on inbound message 1; rewrites #1463's own restore test from "3 outbound + capability proof + announce" to "2 + neither").

**Landing #1463 alone ships a known-broken window** — early ciphertext dropped after promotion, and the session-eviction DoS still live. Squash the two together or merge them atomically.

### MINOR — bounded global-budget DoS
`deferCiphertext`'s global budget (32 packets / ~1.3 MB) is shared across peers. An unauthenticated attacker can spoof message-1 initiations for fake peer IDs (each entering `isAwaitingResponderHandshakeCompletion`), then send garbage ciphertext (`authenticationFailure` is deferrable in that state) to exhaust the global budget so a legitimate peer's early private-file ciphertext is dropped. Throttled by the handshake and per-peer/global rate limiters (`rateLimitExceeded` is drop-only, not deferrable), memory-bounded, and a drop merely equals pre-PR behavior — hence minor, but worth a per-peer reservation.

### MINOR — silent loss by design
A deferred packet whose generation dies before drain fails its single retry and is dropped with only a log; private-file has no application-level retry (the PR body notes this). The buffer is sized for one max file but doesn't guarantee that file's delivery.

### MINOR — decrypt-failure self-heal removed; verify on device
Dropping clearSession+re-handshake for `authenticationFailure`/`invalidCiphertext`/`replayDetected` is defensible — transport uses an explicit 4-byte nonce with a sliding replay window, so auth failure on an established session means wrong keys or forgery, never loss-induced desync, and sessions aren't persisted. But any unforeseen divergence now black-holes silently instead of self-healing.

### NITs
- All Noise packets now dispatch as messageQueue barriers and the already-on-queue path re-dispatches async instead of processing inline, reordering Noise vs non-Noise packets queued in between (Noise-relative FIFO is preserved).
- Stray blank line in `PublicCaptureDelegate` (BLEServiceCoreTests).

Test gap: `transportGenerationNotReady` is covered with an injected readiness closure rather than end-to-end through BLEService's real `privateMediaSessionGenerations` install.

Worth stating for the record: the **ancestor bug on main is real and worth fixing regardless of this stack's fate** — on main any decrypt failure calls clearSession, so a single forged byte can evict an established session.

PR #1465, opened 2026-07-25
Deliver live DM transport events synchronously

jackjackbits:
## Summary

- deliver typed transport events within the existing single MainActor hop instead of adding a nested asynchronous task
- update the conversation store before inbound private-message ACK completion, so an already-open DM window refreshes immediately
- expose an optional synchronous message-acceptance delegate while preserving one-shot delivery to legacy/non-synchronous delegates
- keep blocked or invalid messages unacknowledged and acknowledge accepted duplicates safely

## Compatibility

This is an internal delegate/event-ordering change only. It changes no packets, message types, capabilities, encryption, encoding, or wire behavior, so Android and older iOS clients are unaffected.

## Validation

## Current stack

This is layer 6 of 14 and targets `codex/noise-handshake-promotion-ordering`.

### DIFF
--- a/bitchat/Services/BLE/BLEService.swift
+++ b/bitchat/Services/BLE/BLEService.swift
@@ -4303,16 +4303,54 @@ extension BLEService {
 
     private func emitTransportEvent(_ event: TransportEvent) {
         notifyUI { [weak self] in
-            self?.deliverTransportEvent(event)
+            _ = self?.deliverTransportEvent(event)
         }
     }
 
+    /// Delivers a transport event to the installed delegates and reports
+    /// whether acceptance was confirmed.
+    ///
+    /// For `.messageReceived`, returns `true` only when a
+    /// `SynchronousMessageTransportEventDelegate` synchronously confirmed
+    /// acceptance of the message (duplicates count as accepted). Returns
+    /// `false` when acceptance cannot be confirmed: the sink blocked the
+    /// message, the content was empty, or only a non-synchronous delegate is
+    /// installed so delivery happens without confirmation. Downstream logic
+    /// MUST NOT treat `false` as safe to acknowledge — a `false` return
+    /// means do not ACK.
+    ///
+    /// For all other events, returns `true` when any delegate received the
+    /// event and `false` when no delegate is installed.
     @MainActor
-    private func deliverTransportEvent(_ event: TransportEvent) {
+    @discardableResult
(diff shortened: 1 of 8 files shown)


jackjackbits:
## Review: merge with nits

**The deadlock concern is unfounded — verified.** "Synchronous" here is intra-main-actor only. `notifyUI` remains a fire-and-forget `Task { @MainActor }` after a panic-generation capture; there is no `main.sync` from bleQueue anywhere on the head (the only `DispatchQueue.main.sync` is init-time, before any queues run), and the comment at `scheduleBluetoothStatusSample` preserving "bleQueue must never block on main" is intact. The handlers now running inline call back into BLEService only via methods that dispatch async, or main→bleQueue `.sync` with a `getSpecific` reentrancy guard — the same direction and the same code that already ran on main pre-PR, one runloop tick earlier. Lock graph unchanged.

**It also fixes a real ordering race** rather than creating one: previously `.messageReceived` was deferred via a nested `runOnMain` Task while `.messageDeliveryStatusUpdated` ran synchronously when already on main, so a status update emitted after a message could apply before the message existed in the store and be dropped by `setDeliveryStatus`'s guard. Relative event order at emit sites is preserved, and the new switch replicates every branch of the base default impl including the vouch retry on `.peerListUpdated`.

### MINOR — undocumented contract on the new Bool
`BLEService.deliverTransportEvent` returns an "accepted" verdict that is discarded at every call site in this layer (`_ = ...`). It only becomes load-bearing in #1466, which gates the private-media delivery ACK on `== true`. The failure mode there: any future non-synchronous `eventDelegate` (or the legacy `BitchatDelegate`) returns `false` for `.messageReceived`, so private media would silently never ACK and the sender would resend forever. No live bug today — ChatViewModel is the only delegate ever assigned and it's synchronous — but please document the contract ("false = cannot confirm acceptance; downstream must not ACK") before #1466 lands.

### MINOR — no BLEService-level test of delegate selection
Nothing covers the matrix: sync-eventDelegate preferred, one-shot delivery (non-sync eventDelegate receives the event but returns false, legacy delegate not double-called), legacy-delegate-as-sync fallback. `BLEServiceCoreTests` already has `_test_emitTransportEvent` and a non-sync capture delegate to build on; the new coverage is only at the coordinator/ViewModel layer.

### NIT
The "Synchronously" naming reads scarier than the reality (intra-main-actor only); the doc comment on `didReceiveMessageSynchronously` mitigates this.

No wire/protocol change, no new UI strings.

PR #1466, opened 2026-07-25
Persist authenticated private-media delivery receipts

jackjackbits:
## Summary

- derive stable private-media message IDs from fields already present in the deployed file packet, without adding Android-rejected TLV tags
- durably record accepted encrypted media before UI delivery and ACK, so relaunches and sender retries cannot create duplicate bubbles or files
- correlate media delivery/read receipts to the stable ID
- advertise authenticated capability bit 9 for automatic receipt/retry support while retaining bit 8 as the encrypted-media contract
- preserve legacy/raw receive behavior outside the durable receipt ledger

## Compatibility

Existing wire values remain unchanged: encrypted private file `0x20`, authenticated peer state `0x21`, and signed raw migration fallback `0x22`. Bit-8-only clients still receive stable IDs and ACKs. Older/raw clients remain accepted but do not enter the retry ledger. Reserved bit 10 remains decodable but is never advertised or used.

## Validation

## Current stack

This is layer 7 of 14 and targets `codex/live-dm-transport-events`.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -30626,6 +30626,378 @@
         }
       }
     },
+    "content.delivery.reason.private_media_capability_unresolved" : {
+      "comment" : "Failure reason shown when the peer's support for encrypted media could not be confirmed before sending",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "تعذّر تأكيد دعم الوسائط المشفّرة"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "এনক্রিপ্ট করা মিডিয়া সমর্থন নিশ্চিত করা যায়নি"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "Unterstützung für verschlüsselte Medien konnte nicht bestätigt werden"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Could not confirm encrypted media support"
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "No se pudo confirmar la compatibilidad con multimedia cifrada"
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "پشتیبانی از رسانه رمزنگاری‌شده تأیید نشد"
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "Hindi makumpirma ang suporta sa naka-encrypt na media"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "needs_review",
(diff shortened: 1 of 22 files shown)


jackjackbits:
## Review: merge with nits (1 design major to fix in-stack)

**Security model verified sound.** Media only enters the ledger from the authenticated Noise `privateFile` path; the ACK goes out inside the established Noise session, so a third party cannot forge or inject a receipt. The stable ID binds sender+recipient short IDs plus a filename carrying ≥64-bit (burst) / 122-bit (UUID) entropy that never appears outside the Noise ciphertext, so IDs are unguessable to observers, and Noise replay protection + the ledger make replayed media idempotent (re-ACK, no new bubble/file). Blocked senders are dropped before any disk write without poisoning the reservation. **No wire-format change** — capability bit 9 rides the existing variable-length bitfield whose decoder ignores unknown bits, so rolling upgrade is safe in both directions and bit-8-only/legacy peers keep working.

### MAJOR — one corrupt receipt record silently disables ALL inbound private media, forever
`BLEPrivateMediaReceiptStore.loadIndexIfNeeded`.

Any single decode/read/structural failure aborts index install → every `state(for:)` returns `.unavailable` → the handler withholds save, UI and ACK for **every** stable-ID arrival from **every** peer. Nothing ever quarantines or repairs the bad record (`removeRecord` only runs after a successful scan), so it persists across relaunches indefinitely.

It's deliberate, tested fail-closed design and `.atomic` writes make torn records unlikely — but the failure mode is "all incoming images and voice notes silently vanish and the sender never sees delivered," with no recovery path and nothing surfaced. Please quarantine unreadable records to a sidecar name and fail closed only for that ID (or surface a notice).

### MINOR — `resetForPanic()` is wired to the wrong instance
The `defer { privateMediaReceipts.resetForPanic() }` added to `BLEIncomingFileStore.panicWipe` runs on `PanicRecoveryOperations.live()`'s own private `BLEIncomingFileStore()`. The instance that actually reads/commits receipts — `BLEService.incomingFileStore` — never has its in-memory `runtime.records`/`volatileTombstones` cleared on panic.

Near-inert today (panic rotates the local peer ID, and `.accepted` lookups re-verify file existence) but it contradicts the store's own comment and becomes load-bearing once #1468 wires tombstoned deletion. Note `panicWipeClearsCachedPrivateMediaReceiptDecisions` passes only because it wipes and reads on the *same* instance — it doesn't cover the production wiring.

### MINOR — capacity/TTL eviction is untested
`capacityVictim`, `overflowVictims`, the 4096-per-kind caps, 7-day TTL expiry and `addVolatileTombstone` overflow have no direct tests. Growth is bounded and fine (≤ ~8k small JSONs); the logic just isn't covered.

### MINOR — stable IDs derive from rotating short peer IDs
A sender retry after either side rotates (or after receiver panic) yields a different ID → duplicate bubble and an uncorrelatable ACK. No worse than the previous random-ID behavior, but the "retries cannot create duplicate bubbles" guarantee only holds *within a peer-ID generation* — worth a doc note.

### Localization — needs fixing here
`content.delivery.reason.private_media_capability_unresolved` (introduced by this PR) is **absent from `bitchat/Localizable.xcstrings`**, so it renders English-only in all 29 locales. #1393's coverage guard doesn't catch it because it only validates keys already present in the catalog. #1467 adds `private_media_delivery_unconfirmed` with the same problem — please add both keys here.

### NIT
`state(for:)` holds the store lock across disk I/O, and the first lookup scans/decodes the whole receipt directory.

PR #1467, opened 2026-07-25
Retry confirmed private media after reconnect

jackjackbits:
## Summary

- retain a bounded encrypted private-media send after local BLE completion until a correlated delivery receipt arrives
- retry only after the exact current Noise generation authenticates receipt capability bits 8 and 9
- re-check generation/capability ownership at policy resolution, retry start, message-queue admission, and encryption
- bound retained packet count, bytes, lifetime, retries per message, and retries per reconnect
- emit a visible terminal failure and release retained bytes/transfer ownership when limits, expiry, cancellation, or downgrade end retry

## Compatibility

Automatic retry is never authorized by public announces, bit-8-only peers, raw legacy files, or stale sessions. Existing Android/older clients continue their current one-send behavior. The sender always uses the existing encrypted private-file wire shape and never introduces or advertises reserved bit 10.

## Validation

## Current stack

This is layer 8 of 14 and targets `codex/private-media-receipts`.

### DIFF
--- a/bitchat/Services/BLE/BLEService.swift
+++ b/bitchat/Services/BLE/BLEService.swift
@@ -1145,6 +1145,29 @@ final class BLEService: NSObject {
         collectionsQueue.sync { peerRegistry.capabilities(for: peerID) }
     }
 
+    func authenticatedPrivateMediaReceiptSessionGeneration(
+        to peerID: PeerID
+    ) -> UUID? {
+        let normalizedPeerID = peerID.toShort()
+        let currentNoiseGeneration =
+            noiseService.sessionGeneration(for: normalizedPeerID)
+        return collectionsQueue.sync {
+            guard let generation =
+                    privateMediaSessionGenerations[normalizedPeerID],
+                  generation == currentNoiseGeneration,
+                  let authenticated =
+                    authenticatedPeerStates[normalizedPeerID],
+                  authenticated.sessionGeneration == generation,
+                  authenticated.capabilities.contains(.privateMedia),
+                  authenticated.capabilities.contains(
+                    .privateMediaReceipts
+                  ) else {
+                return nil
+            }
+            return generation
+        }
+    }
+
     private func privateMediaPolicyFingerprint(
         for peerID: PeerID,
         expectedSessionGeneration: UUID?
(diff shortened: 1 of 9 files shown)


jackjackbits:
## Review: merge with nits

**No media duplication** — verified. Retries are gated to bit-9-authenticated, exact-generation sessions at four boundaries (policy resolution, retry start, messageQueue admission, encryption), and the base layer's receiver dedup (stable ID + durable reservation ledger + tombstones) absorbs and re-acks duplicates. Legacy/bit-8-only peers can never receive a retry. **No deadlock** — the coordinator is `@MainActor`, `.peerConnected` arrives via `notifyUI`'s async hop (never main.sync from bleQueue), and `authenticatedPrivateMediaReceiptSessionGeneration` reads noiseService before `collectionsQueue.sync` with no nested lock hold. Expiry Tasks are token-guarded and cancelled on discard/terminal/panic.

### MINOR (i18n) — new string missing from the catalog
`content.delivery.reason.private_media_delivery_unconfirmed` is **not in `bitchat/Localizable.xcstrings`** — it will render English-only in all 29 locales. The reused `private_media_capability_unresolved` has the same problem (originates in #1466; this PR adds two more call sites). #1393's coverage guard passes silently because it only validates keys already present. Please add both — ideally in #1466 where the first was introduced.

### MINOR — a dropped policy callback can wedge retries for a peer
`ChatMediaTransferCoordinator.resolveReconnectRetries`: `peersResolvingReconnectRetry[peer]` is removed only inside the `resolvePrivateMediaSendPolicy` completion. If a completion is ever dropped (BLEService's `messageQueue.async { guard let self else { return } }` on teardown), an entry with `replacingActiveTransfer == true` permanently blocks all future retry resolutions for that peer until `resetForPanic`. Records then expire to visible failure, so it degrades gracefully — but a per-entry timeout, or removal on `peerDisconnected`, would be more robust.

### MINOR — redundant resend on link flap
A raw `.peerConnected` with an unchanged Noise generation (session survived the flap) still schedules a full-file retry with no grace delay for an in-flight ack. The receiver dedups so nothing surfaces, but bandwidth is wasted on flapping links. Bounded (2/message, 2/reconnect) and reconnect-gated, so no storm.

### MINOR — pessimistic status semantics worth confirming
A locally-completed send whose receipt never arrives flips `.sent` → `.failed("Delivery could not be confirmed")` after 120s even though the receiver may hold the file (ack lost, no reconnect). The copy is accurate and a late receipt self-heals, but please confirm this is the intended default.

### NITs
- `Transport.swift`'s default `sendFilePrivateReceiptRetry` is a silent no-op; a non-BLE transport routed here would never clear `activeTransferID`, retaining bytes forever (unreachable today).
- Retry attempts suppress `.started`/`.updated` events, so a retry after a mid-transfer cancel shows the stale `partiallyDelivered(X/Y)` until completion.
- `terminalizeReconnectRetry`: the second `removeValue != nil` guard is dead.

Tests are good (7 coordinator + 2 E2E with injected clock/ID factory, incl. bit-8-only never retaining and the deferred messageQueue boundary re-checking bit 9 after rotation). Gaps: no test for a receipt arriving during the initial transfer preventing the armed retry, and nothing catches the missing xcstrings keys.

PR #1468, opened 2026-07-25
Make private-media deletion transactional

jackjackbits:
## Summary

- make individual media deletion and `/clear` receiver state transactional with a write-ahead deletion journal as the durable batch commit point
- cancel active transfer and retained reconnect-retry ownership before tombstone persistence or UI removal
- unlink only exact, journal-owned regular files inside incoming media directories
- preserve shared, pending-delivery, active voice, reused, legacy-ambiguous, and non-regular paths
- recover partially materialized deletions after crash and keep accepted/tombstoned receipt state fail-closed when storage is unreadable
- serialize overlapping clears, identity migration, UI insertion, allocator reservation, quota eviction, and payload unlinking

## Compatibility

Deletion is entirely local; no deletion packet or wire-format change is introduced. Existing `0x20` encrypted files, `0x22` legacy/raw files, Android clients, and older iOS clients remain readable. Ambiguous legacy payloads are removed from UI but left for bounded quota cleanup instead of risking deletion of a newer file.

## Validation

## Current stack

This is layer 9 of 14 and targets `codex/private-media-retry`.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -1,6 +1,42 @@
 {
   "sourceLanguage" : "en",
   "strings" : {
+    "content.system.media_delete_refused" : {
+      "comment" : "System message shown in the affected chat when an explicit media delete or /clear was refused and bubbles/files were kept",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : { "stringUnit" : { "state" : "needs_review", "value" : "تعذّر حذف بعض الوسائط. جرّب حذف الوسائط الأقدم أولاً." } },
+        "bn" : { "stringUnit" : { "state" : "needs_review", "value" : "কিছু মিডিয়া মুছে ফেলা যায়নি। আগে পুরোনো মিডিয়া মুছে ফেলার চেষ্টা করুন।" } },
+        "de" : { "stringUnit" : { "state" : "needs_review", "value" : "einige medien konnten nicht gelöscht werden. versuche zuerst, ältere medien zu löschen." } },
+        "en" : { "stringUnit" : { "state" : "translated", "value" : "some media could not be deleted. try deleting older media first." } },
+        "es" : { "stringUnit" : { "state" : "needs_review", "value" : "no se pudieron eliminar algunos archivos multimedia. prueba a eliminar primero los más antiguos." } },
+        "fa" : { "stringUnit" : { "state" : "needs_review", "value" : "برخی رسانه‌ها حذف نشدند. ابتدا رسانه‌های قدیمی‌تر را حذف کنید." } },
+        "fil" : { "stringUnit" : { "state" : "needs_review", "value" : "hindi ma-delete ang ilang media. subukang i-delete muna ang mas lumang media." } },
+        "fr" : { "stringUnit" : { "state" : "needs_review", "value" : "impossible de supprimer certains médias. essaie d'abord de supprimer les médias plus anciens." } },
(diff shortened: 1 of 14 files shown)


jackjackbits:
## Review: request changes (2 majors)

The journal machinery itself verified sound: the `.atomic` write is a genuine commit point (temp+rename on APFS), materialization is idempotent and retried on recovery, double-delete is a no-op, there is no lock-order inversion (`payloadCoordination.lock` → `runtime.lock`, never reversed), panic clears journal cache + reservations + clear queue, delete-vs-in-flight-arrival is handled by failing the reservation when the path is pending, and pathless per-ID tombstones mean a reused basename can never be deleted by a stale tombstone. Tests are real behavioral coverage (~1.9k of the 3.4k added lines), not scaffolding.

### MAJOR 1 — `/clear` destroys outgoing media mirrored into another conversation
`ChatViewModel.swift` `continueClear`, outgoing block (~L2194-2210).

After tombstone commit, `outgoingMediaIDs` are removed from **every** direct conversation and `cleanupOutgoingLocalFile` unlinks the backing file — unconditionally. Incoming stable IDs get `hasRemainingCopy`/`capturedExclusiveIDs` alias protection (with a test: `clearPrivateChatKeepsMediaReferencedByAnotherConversation`); outgoing gets none.

Scenario: an identity-alias handoff mirrors an outgoing media message into a second direct conversation → `/clear` on the first removes the second conversation's bubble **and deletes its file**.

This is the open Codex P2 inline comment ("Preserve shared outgoing media aliases during clear"), still unaddressed at head 11afe965. Fix: mirror the incoming `hasRemainingCopy` logic for outgoing IDs, and add the missing alias test — its absence is why this survived.

### MAJOR 2 — explicit delete now leaves legacy incoming plaintext on disk (needs sign-off)
`removeUntombstonedMediaMessage` + the `/clear` path deliberately leave any non-stable-ID incoming payload on disk "for bounded quota cleanup" (100MB, oldest-first — i.e. potentially indefinitely). On main, delete/clear unlinks by basename (`context.removeMessage(withID:cleanupFile:true)`, main `ChatMediaTransferCoordinator.swift:337/342`).

These files are **decrypted plaintext** (Noise decrypts on the wire; `saveIncomingFile` writes decoded content). So on a privacy-focused app, deleting media received from Android or older-iOS senders removes the bubble but retains the bytes. That's a regression versus main.

The irony: this PR builds exactly the machinery that would make unlinking safe (`pendingDeliveryPaths` + reservations) and uses it for stable IDs but not here; `cleanupIncomingLocalFile` exists and is never called in this layer. Either gate an unlink on not-pending/not-reserved, or get explicit maintainer sign-off on the retention policy.

### MINOR — fail-closed deletion is silent and can wedge permanently
Every refusal path only `SecureLogger.error`s — no system message. Worse, `recordDeleted(messageIDs:)` fails the **whole batch** if any target path intersects `protectedPayloadRelativePaths` or another ID's record, so a basename collision between a stable receipt path and a surviving legacy bubble makes `/clear` of that chat a permanent silent no-op, with the only workaround (delete the legacy bubble first) never surfaced. Low exploitability (filenames embed timestamps/UUIDs), but the failure mode is invisible.

### MINOR — failed `/clear` leaves zombie outgoing sends
`clearPrivateChat` cancels send ownership *before* the journal commit. If the commit fails, bubbles are preserved (tested) but their transfers are dead — stuck "sending" forever with no retry.

### MINOR — recovery is lazy, not eager
`recoverDeletionJournal` runs only when the receipt store is next touched, so after a crash between commit and unlink the bytes persist until the next media operation. Consider a launch-time pass.

### NIT
`save()` holds `payloadCoordination.lock` across full receipt-directory load + journal recovery (incl. unlink retries) on the messageQueue barrier.

PR #1475, opened 2026-07-26
fix(favorites): normalize PeerID with toShort() in favorite status lo…

kalp-cg:
### Summary of Changes
Fixes a `PeerID` prefix mismatch in `FavoritesPersistenceService.getFavoriteStatus(forPeerID:)`.

- Normalizes incoming `peerID` with `.toShort()` before comparing against `PeerID(publicKey: pubkey)`.
- Ensures prefixed peer IDs (`mesh:`, `noise:`) properly match stored favorites.

### DIFF
--- a/bitchat/Services/FavoritesPersistenceService.swift
+++ b/bitchat/Services/FavoritesPersistenceService.swift
@@ -188,7 +188,8 @@ final class FavoritesPersistenceService: ObservableObject {
     func getFavoriteStatus(forPeerID peerID: PeerID) -> FavoriteRelationship? {
         // Quick sanity: peerID should be 16 hex chars (8 bytes)
         guard peerID.isShort else { return nil }
-        for (pubkey, rel) in favorites where PeerID(publicKey: pubkey) == peerID {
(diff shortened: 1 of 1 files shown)


jackjackbits:
Thanks for digging into the favorites lookup — but as written this is a no-op. The function already starts with `guard peerID.isShort else { return nil }`, and `isShort` requires a 16-hex `bare`. `toShort()` only converts when `noiseKey != nil`, which requires a 64-hex bare — impossible after that guard — so `peerID.toShort()` always returns `self` here.

It also wouldn't fix the scenario in the description: a `mesh:`-prefixed short ID passes `isShort` (the prefix is stripped for `bare`), but `toShort()` doesn't strip the prefix, and `PeerID`'s synthesized `Equatable` compares prefix + bare — so the mismatch would persist. If a caller is actually passing prefixed IDs into this lookup, the right fix would be comparing `.bare` (or re-wrapping as an empty-prefix PeerID); a failing test demonstrating the mismatch would make that an easy merge. Closing this one since it doesn't change behavior.

PR #1476, opened 2026-07-26
Add oldest-first disk quota for outgoing media

Ayush7614:
## Summary
- Apply a separate 100 MB oldest-first quota to outgoing voice notes, images, and files (incoming already had this budget).
- Enforce the outgoing quota before writing default Application Support media paths in `ImageUtils`, `VoiceRecorder`, and `VoiceCaptureSession`.
- Hold `QuotaByteReservation` headroom for multi-step writes and release it on cancel / encode-write failure / panic wipe.
- Protect in-progress outgoing capture paths from concurrent eviction (register in the pending-delivery exclusion set until stop/cancel) — `voice_<…>.m4a` is not covered by the `voice_live_` prefix.
- Wire `BLEService` / `ChatLiveVoiceCoordinator` / AppRuntime age sweep through `BLEIncomingFileStore.shared` so reservation and eviction exclusions are visible process-wide (tests that need isolation still inject a temp `baseDirectory`).
- Update privacy docs to match size quotas **in addition to** 7-day age retention, and add focused unit tests including decisive `reservingBytes` cases.

## Test plan
- [x] `xcodebuild` macOS Debug build with `CODE_SIGNING_ALLOWED=NO` succeeded
- [x] `BLEIncomingFileStoreOutgoingQuotaTests` (10 tests) passed on macOS
- [ ] Manual: fill outgoing media near 100 MB, send another image/voice note, confirm oldest outgoing file is removed and incoming files are untouched
- [ ] Manual: start a long voice capture near quota, trigger another outgoing write, confirm the in-progress capture file is not deleted mid-recording


jackjackbits (changes requested):
The idea is sound and complementary to what's on main, but the branch is based on a merge base from before several structural changes, and the conflicts are substantive rather than textual:

1. **Rebase must preserve main's deletion coordination.** `enforceQuota` on main now takes `payloadCoordination.lock` and excludes files under active `deletionReservations` (#1468 transactional deletion) and `pendingDeliveryPaths`. This rewrite predates all of that — keeping the PR's version on rebase would silently strip those protections from the incoming path too. Please express the outgoing scope inside main's current `enforceQuota` shape (lock + protected paths + `voice_live_` exclusion).

2. **Enforcement must run through the shared store instance.** `PayloadCoordination` is per-instance, and this PR triggers eviction from fresh `BLEIncomingFileStore()` instances constructed inside `ImageUtils.makeOutputURL`, `VoiceRecorder`, and `PTTLiveVoiceSession` — those can never see reservations registered on the long-lived store the BLE/deletion services use, so the exclusions would be vacuous even once added.

3. **Docs need redoing on top of #1484**, which landed a 7-day age expiry across both incoming and outgoing trees and rewrote `PRIVACY_POLICY.md` / `docs/privacy-assessment.md`. The PR's doc edits amend the pre-#1484 text and would contradict main. (#1484 also reduces the urgency — outgoing is now bounded to a 7-day window — but a size cap within that window is still worth having.)

4. After rebase, please add a test proving outgoing eviction honors deletion reservations and the live-capture prefix.

The test coverage you already have is good — this is worth finishing.

Ayush7614:
Thanks for the detailed review @jackjackbits  — all four points are addressed in `c2c89d0` (rebased onto current main):

1. **Deletion coordination preserved.** Outgoing scope now lives inside the existing `enforceQuota` shape: `payloadCoordination.lock`, `deletionReservations` ∪ `pendingDeliveryPaths` exclusions, and `voice_live_` skip.
2. **Shared store instance.** Added `BLEIncomingFileStore.shared` and wired BLEService / live-voice / AppRuntime age sweep / ImageUtils / VoiceRecorder / PTT fallback writers through it so reservation state is visible to eviction.
3. **Docs redone on top of #1484.** Privacy policy + assessment now describe separate 100 MB size quotas for incoming and outgoing *in addition to* the 7-day age retention.
4. **New tests.** Coverage for pending-delivery + live-capture protection on outgoing eviction, plus deletion-reservation isolation on the same store instance (`BLEIncomingFileStoreOutgoingQuotaTests`, 4 tests passed on macOS).

Chessing234:
quota reservation idea makes sense. please confirm reservingBytes gets released if compression fails or the person cancels mid-capture, and that panic wipe clears in-flight reservations.

Ayush7614:
Good catch @Chessing234 — confirmed and fixed in bc2bc09f.

`reservingBytes` on `enforceQuota` was always call-scoped (headroom for a synchronous write, nothing held afterward). Multi-step writers (image encode→disk, voice capture, live PTT) now take an explicit `QuotaByteReservation` and release it on every exit path:

- **Compression / write failure** — `ImageUtils` reserves after successful JPEG encode and `defer`-releases even if the disk write throws.
- **Cancel mid-capture** — `VoiceRecorder` and `PTTLiveVoiceSession` release on cancel / start failure / stop; live inbound assemblies release on `cancelAssembly` / `finalize`.
- **Panic wipe** — `panicWipe` clears all in-flight `byteReservations` (alongside pending-delivery and deletion reservations). Stale tokens after wipe are no-ops.

Added focused tests in `BLEIncomingFileStoreOutgoingQuotaTests` (release-on-failure pattern, panic clears held bytes, held bytes tighten eviction). All 8 suite tests passed on macOS.

Ayush7614:
Rebased onto current `main` and resolved the conflicts with the file-protection work (#1552):

- Kept `mediaProtectionAttributes` / `migrateFileProtectionIfNeeded` / `performMediaMaintenance`
- Routed maintenance through `BLEIncomingFileStore.shared` so age expiry shares coordination with quota writers
- Preserved outgoing quota + held `QuotaByteReservation` release paths on top of the protection attrs

`BLEIncomingFileStoreOutgoingQuotaTests` (8 tests) passed on macOS after the rebase.

Ayush7614:
cc: @jackjackbits

vincenzopalazzo (commented):
**Production Readiness: NO-GO** — the quota itself looks right; the hole is that in-progress outgoing captures aren't protected from it.

The oldest-first outgoing enforcement is well placed (single choke point before each write) and the tests cover the cross-media-kind and incoming-vs-outgoing scoping nicely.

### Issues & Suggestions

- **Medium:** `bitchat/Features/voice/VoiceCaptureSession.swift` (`makeOutputURL`, ~L237-251) and `bitchat/Features/voice/VoiceRecorder.swift` (`makeOutputURL`, ~L300-311) — files being actively recorded (`voice_<burstID>.m4a`, `voice_<ts>_<uuid>.m4a`) are protected from `enforceQuota` neither by the `voice_live_` prefix nor by `pendingDeliveryPaths`. Eviction is oldest-first so the live file is hit last — but if most of the tree is protected (large pending-delivery set), a concurrent `enforceOutgoingQuota` can delete the file a recording session is writing to. Suggest registering the capture path in `pendingDeliveryPaths` (or using the live prefix) until the send completes.
- **Medium:** `bitchatTests/Services/BLEIncomingFileStoreOutgoingQuotaTests.swift` — the scenarios use 105MB against the 100MB quota, so a single oldest eviction satisfies the target whether or not `reservingBytes: 10MB` is honored; the tests pass even if the reservation is ignored entirely. Please size one case so the reservation decides the outcome (e.g. 98MB total with `reservingBytes: 10MB` → must evict; same 98MB with `reservingBytes: 0` → must not).
- **Cleanup:** the `incomingFileStore`/`fileStore` defaults changed from per-instance `BLEIncomingFileStore()` to `.shared` (`BLEService.swift` ~L512, `ChatLiveVoiceCoordinator.swift` ~L161). That's clearly intentional (quota exclusions need the shared store), but it changes isolation for any test constructing these with defaults — worth one line in the PR description.
- **Cleanup:** `ImageUtils.makeOutputURL(outputDirectory:reservingBytes:)` silently ignores `reservingBytes` when a custom directory is passed. Production callers all use the default path, so this is test-only surface — document it or drop the parameter from the private helper.

### Summary

Solid approach; please close the in-progress-capture eviction hole and make the reservation test decisive.

---

Ayush7614:
Addressed the production-readiness review in 48b89c45:

1. **In-progress capture protection** — `VoiceRecorder` / `PTTLiveVoiceSession` / `ImageUtils` now call `beginEvictionProtection` / `endEvictionProtection` (same exclusion set as `pendingDeliveryPaths`) for the duration of the write/capture, so a concurrent `enforceOutgoingQuota` cannot unlink an open `voice_<…>.m4a` / JPEG destination. Added `outgoingQuotaSkipsProtectedInProgressCapture`.
2. **Decisive reservation test** — `outgoingQuotaReservingBytesDecidesEviction` uses a single 98 MB file: stays with `reservingBytes: 0`, must evict with `reservingBytes: 10 MB`. Same sizing for the held-byte reservation test.
3. **Shared store default** — called out in the PR description (`BLEService` / `ChatLiveVoiceCoordinator` / AppRuntime maintenance use `.shared`; temp-directory tests still construct their own instance).
4. **ImageUtils custom directory** — documented that a non-nil `outputDirectory` is a test/custom escape hatch that skips quota entirely.

All 10 suite tests passed on macOS.

Ayush7614:
Updated the branch onto current `main` and fixed the failing **Periphery scan**:

- `QuotaByteReservation.scope` / `.bytes` were assign-only (values already live in the coordination map keyed by id). The token is now an opaque `id` only, matching `PrivateMediaDeletionReservation`.
- Rebased cleanly onto latest `main` (including the already-merged #1477 / #1478 and follow-ups).

`BLEIncomingFileStoreOutgoingQuotaTests` (10 tests) still pass on macOS.

Ayush7614:
cc: @jackjackbits

PR #1477, opened 2026-07-26
Add macOS camera QR scanning for peer verification

Ayush7614:
## Summary
- Replace the macOS paste-only verification scanner placeholder with a real `AVCapture` QR camera preview, matching iOS.
- Keep paste/validate as a macOS fallback when no camera is available or permission is denied.
- Share scanner coordinator logic across platforms and extend the ViewSmoke camera test to macOS.
- Check `AVCaptureDevice.authorizationStatus(for: .video)` before creating `AVCaptureDeviceInput`, and gate the smoke test on prior authorization so test runners do not trigger TCC prompts / CI hangs.
- Show a one-line hint over the preview when the camera is unavailable or permission is denied.
- iOS layout note: `VerificationSheetView` scanner content uses `.frame(minHeight: 280)` instead of `.frame(height: 280)` so the sheet can grow with Dynamic Type / macOS paste fallback content.

## Test plan
- [x] `xcodebuild` macOS Debug build with `CODE_SIGNING_ALLOWED=NO` succeeded
- [x] `ViewSmokeTests` including `cameraScannerView_previewAndCoordinatorSmoke` passed on macOS
- [x] Manual on Mac: open verification sheet, allow camera, scan another peer QR, confirm verification request fires
- [x] Manual: deny camera permission and confirm paste/validate path still works
- [ ] Quick iOS render check of the verification sheet scanner branch after the `minHeight` change

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -70439,6 +70439,192 @@
         }
       }
     },
+    "verification.scan.camera_unavailable" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "الكاميرا غير متاحة — الصق رمز QR أدناه."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Camera unavailable — paste a QR below."
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Kamera nicht verfügbar — QR unten einfügen."
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Camera unavailable — paste a QR below."
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Cámara no disponible — pega un QR abajo."
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Camera unavailable — paste a QR below."
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Camera unavailable — paste a QR below."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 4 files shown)


Ayush7614:
Thanks for the review — agreed.

Sandboxed macOS builds need `com.apple.security.device.camera` in addition to `NSCameraUsageDescription`. Added that entitlement in `bitchat/bitchat-macOS.entitlements` in 338e351 so the AVCapture QR path can open the camera after the user grants permission.

jackjackbits:
This is close to merge — the security semantics are preserved by construction (both platforms funnel through the identical `verifyScannedPayload` → fingerprint flow, nothing forked), `NSCameraUsageDescription` is in the shared Info.plist the macOS target uses, and the sandbox camera entitlement from the Codex round is confirmed in the diff. Three asks:

1. **Check authorization before creating the capture input.** `CameraScannerCoordinator.setup` builds `AVCaptureDeviceInput` before consulting `AVCaptureDevice.authorizationStatus(for: .video)`. With the smoke test now running on macOS, any test runner on a Mac with a camera triggers a TCC consent prompt for the runner process — the same class of problem we just fixed for keychain in #1413, and a potential hang on CI hosts with a virtual camera. Check status first (or gate the smoke test on authorization).
2. Optional polish: with no camera or denied permission, the view shows a black 260-pt rectangle above the paste fallback — a one-line hint there would help.
3. The iOS layout change (`frame(height: 280)` → `frame(minHeight: 280)` in `VerificationSheetView`) is unadvertised — please mention it in the description and give the sheet a quick render check on iOS.

Then this merges.

Ayush7614:
@jackjackbits Thanks for the review!

Addressed the three asks in 3ad7492:

1. **Auth before capture input** — `CameraScannerCoordinator.setup` now consults `AVCaptureDevice.authorizationStatus(for: .video)` first and only constructs `AVCaptureDeviceInput` after `.authorized` (or after a granted `requestAccess`). The smoke test also skips `setup` unless status is already `.authorized`, so test runners won’t trip a TCC prompt / CI hang.
2. **Unavailable hint** — when there’s no camera or permission is denied, the preview shows a one-line “Camera unavailable — paste a QR below.” overlay above the paste fallback.
3. **iOS `minHeight`** — called out in the PR description (`frame(height: 280)` → `frame(minHeight: 280)`). Still worth a quick iOS sheet render check on your side before merge.

PR #1478, opened 2026-07-26
Localize hardcoded macOS image picker copy

Ayush7614:
## Summary
- Move `MacImagePickerView` user-facing English literals into `Localizable.xcstrings`.
- Add translations for all supported app locales so localization coverage stays complete.
- Leave `#Preview`-only helper text unlocalized.

## Test plan
- [x] `LocalizationCoverageTests` passed on macOS
- [x] Manual: open macOS image picker UI and confirm title/select/cancel/panel message resolve from the catalog


yhyya11:
[Uploading CarTube.html…]()

PR #1482, opened 2026-07-26
Create SECURITY.md for security policy

Cyril-the-Real:
Added a security policy document outlining supported versions and vulnerability reporting.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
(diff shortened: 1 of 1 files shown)


jackjackbits:
Thanks for the interest in getting a security policy in place — but this is the unedited GitHub template: the placeholder instructions are still in the prose, and the version table (5.1.x / 4.0.x) doesn't correspond to this project's actual releases (currently 1.7.x). Publishing it as-is would give people misleading guidance and no real reporting channel.

If you'd like to take another run at it, a useful SECURITY.md for this project would need: a real contact/reporting channel, which release lines actually receive fixes, and expectations for response time. Closing this one in the meantime.

PR #1492, opened 2026-07-26
Stop broadcasting the adjacency graph and the authorship marker

jackjackbits:
Two Tier 1 radio-metadata leaks that need **no cross-platform agreement**, because each only changes what this device chooses to emit. Companion: #1487 (peer ID rotation, which does need agreement).

## Announces no longer carry the neighbour list

The TLV held up to ten 8-byte peer IDs, so a **single** passive receiver could reconstruct the local adjacency graph — who is standing next to whom — with no need for several receivers or RSSI trilateration. In a crowd that's the most sensitive thing the radio layer gives away, and unlike the identity keys it isn't required for the protocol to work.

Backward compatible in both directions: an empty list omits the TLV entirely rather than emitting a zero-length one, the decoder already treats its absence as "no topology offered", and lists from other peers are still parsed so a mixed network behaves sensibly.

**The cost is source routing**, and it's worth being explicit. `MeshTopologyTracker` builds its adjacency map from these lists, so with everyone silent there are no routes to compute and directed traffic floods instead — which is already the documented fallback whenever a route fails. More airtime for directed sends in dense meshes; no correctness change.

Left as a `TransportConfig` constant rather than a user setting because it's a protocol trade-off, not a preference. If you'd rather keep source routing, it's a one-line flip and the reasoning is in the constant's doc comment.

## Public broadcasts no longer always originate at the maximum TTL

`ttl == messageTTLDefault` was a reliable "this device wrote it" marker to any direct listener — that discloses *authorship*, not merely presence, which is the more damaging inference.

Origin TTL is now drawn from `5...7`. In a dense graph relays already clamp broadcasts to 5, so an origin emitting 5 is indistinguishable from relayed traffic; in a sparse chain a 6 could be an origin or one hop from a 7. The floor deliberately isn't below the dense-graph clamp, since lower would cost reach without buying ambiguity that clamp doesn't already provide.

Signature-safe and needs no agreement: TTL is excluded from the signed bytes (`toBinaryDataForSigning` zeroes it so relays can decrement without invalidating), so a peer on any version simply sees a smaller starting TTL and relays it normally.

**Announces keep the fixed TTL on purpose.** Three link-binding paths read a maximum-TTL announce as "direct link", and an announce's sender ID already identifies the device — so there's nothing to hide there and something to break.

## Not done here: padding — and why

The third item on my earlier list was extending padding beyond Noise frames and fixing the gap where a frame needing more than 255 bytes of padding is emitted unpadded (payloads of 241–256, 497–768 and 1009–1792 bytes ship at exact length today).

It looked unilateral. It isn't. `BitchatPacket.toBinaryDataForSigning()` calls `BinaryProtocol.encode(unsignedPacket)`, whose `padding` parameter defaults to `true` — so **the padding bytes are inside the signed material for every signed packet**. Changing the padding algorithm changes the signed byte stream, and signature verification then fails against any peer that hasn't changed it identically.

So padding is frozen by cross-platform signature compatibility and belongs with the coordinated work. Recorded in `docs/privacy-assessment.md` and in #1487's open questions rather than attempted here. I'd rather flag that than ship a change that silently partitions the mesh.

## Testing

- 8 new tests: TTL stays in range, isn't always the maximum, honours an injected randomizer, survives a degenerate range without trapping, and the range sits between the dense clamp and the default; the neighbour flag defaults off, an empty list omits the TLV while everything else round-trips, and received lists still parse.
- Full suite green: 1953 tests in 207 suites.
- `docs/privacy-assessment.md` updated, since the project's own checklist treats announce fields as privacy-surface changes — the "bounded set of short direct-neighbor identifiers" line it listed is no longer true.

**Needs device testing:** directed delivery in a multi-device mesh now that source routing has no topology to work with — the expected outcome is "slower, still delivered", and that should be confirmed on real hardware rather than assumed.


jackjackbits:
Note for the Android side, since this one is easy to read as iOS-internal and isn't quite:

**The neighbour-list change has a cross-platform effect even though the wire change is backward compatible.** Omitting the TLV is safe for any decoder — it's optional, and lists from other peers are still parsed — but `MeshTopologyTracker` builds its adjacency map *from peers' announces*. So once iOS stops advertising, Android's source routing loses topology for every iOS peer and falls back to flooding for those destinations. That's the documented fallback and not a correctness problem, but it will show up as more airtime for directed sends in dense meshes, and it'd be misleading to have you discover it as a regression.

Three ways to go, and I don't think this PR should decide it unilaterally:

1. Android matches (stops advertising too) — source routing effectively retires, and the adjacency leak is closed on both platforms.
2. Android keeps advertising — Android→Android routes keep working, and the leak persists for Android users, which is the population most likely to be on a cheap phone at a protest.
3. Neither advertises by default, but the list returns later restricted to Noise-authenticated links, so topology is shared only with peers you've actually handshaked with. That keeps most of the routing benefit without broadcasting the graph to passive listeners.

I'd lean (3) as the eventual answer and (1) as the interim, but it's genuinely your call — the whole point of the leak is that a *single* passive receiver reconstructs who is standing next to whom, and that's worse for Android users than for iOS ones in most of the places this app matters.

The TTL half needs nothing from you: TTL is excluded from the signed bytes, so a peer on any version just sees a smaller starting TTL and relays it normally.

Chessing234:
good privacy cut — neighbour tlv + fixed max ttl authorship marker were real passive leaks.

two notes: source routing will degrade to flood until enough peers still announce neighbours (fine if documented as intentional). also the privacy win is incomplete in mixed rooms until android stops emitting the same tlv — worth a tracking issue if one isn't open yet.

RodsF1:
I merged this onto current `main` to try it out (clean merge — the `BLEService` split since July doesn't touch the TTL/announce call sites, and the wiring guard passes: all six drawn sites survive, the ten remaining fixed-TTL sites are each a documented exclusion, full suite green at 2030 tests). Two things ship *inconsistent* with the change unless they land alongside it — happy to open a PR against this branch with both if useful:

**1. The whitepaper still describes the closed leaks as live.** `docs/privacy-assessment.md` was updated here, but `WHITEPAPER.md` wasn't: §4.2 still says packets "originate with TTL 7", §4.3 still says announcements "carry up to 10 direct-neighbor IDs", §8 lists both as current metadata leaks, and §9 lists the neighbour-list one as future work. I'd correct all four — and deliberately *not* write the TTL change up as a win, matching how `TransportConfig` and the privacy assessment already frame it: relays strictly decrement, so the top of the range is still origin-only (~one message in three, ~87% of senders self-identified within five). A whitepaper that overstates a mitigation is worse than one that omits it.

**2. The topology map goes quietly wrong — and not in the obvious way.** I expected it to go blank; it doesn't. `currentMeshTopology()` calls `refreshLocalTopology()` first, which records *this device's own* links from live connections rather than from received announces, so those edges survive while remote peers contribute nothing. The map therefore degrades to a **star centred on self** — which doesn't read as degraded, it reads as "a mesh where nobody else is connected to anybody." Someone diagnosing patchy connectivity would take that as a finding. The caption makes it worse: it still says "estimated from gossiped neighbor lists (up to 10 per peer)", now false twice over (the edges are local, and nothing is gossiped). I'd reword the caption across all 30 locales to say the links shown are your own, and fix the three doc comments (`BLEService`, `AppChromeModel`, `MeshTopologyView`) that still describe edges as gossiped claims. No behaviour change — the star is the honest picture of what a device can still see, worth keeping as long as it's labelled partial.

One thing worth surfacing for the reviewers rather than hiding in a doc: with neighbour lists suppressed network-wide, no source route is ever *confirmable* (confirmation needs both endpoints to claim the edge), so directed traffic always floods. That's the documented fallback, not a regression — but it makes source routing effectively dead code in a fully-upgraded mesh, which seems like a call the maintainers should make consciously. I noted it in §9 as follow-up ("restore source routing without publishing adjacency").

RodsF1:
Opened #1700 against this branch with both fixes (whitepaper consistency + the honest topology map). Sits directly on the branch tip; net diff is just the two commits.

Chessing234 (commented):

RodsF1:
Merge-order note: #1700 targets this branch rather than main, so it'd disappear when this lands — on a squash with the head branch deleted GitHub closes a dependent PR instead of retargeting it (#1655 and #1404 both went that way).

PR #1495, opened 2026-07-26
fix(arti): relay until both halves close to avoid truncating responses

aashish-thapa:
The SOCKS bridge relays both direction with select! which drops the surviving copy the moment either side hit EOF. 
When a client half-closes its write side after sending a request, the client to tor copy completes and the tor to client
copy is cancelled before the response even arrives silently truncating it.

So its better to switch to `tokio::io::copy_bidirectional` which propagates the half close and keeps forwarding until both
direction closes. I have also added a regression test over in-memory duplex streams that fails on the old select! and passes
now.

### DIFF
--- a/localPackages/Arti/arti-bitchat/src/socks.rs
+++ b/localPackages/Arti/arti-bitchat/src/socks.rs
@@ -7,7 +7,7 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 
 use arti_client::{TorClient, IntoTorAddr};
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tokio::net::TcpStream;
 use tor_rtcompat::PreferredRuntime;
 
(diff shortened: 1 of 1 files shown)


aashish-thapa:
@jackjackbits will appreciate if you can review this.

jackjackbits (changes requested):
Thanks for this — the diagnosis is correct and well-observed. The `select!` over two one-way copies really does cancel the tor→client copy at the first EOF, and a client that half-closes after sending its request gets its response truncated by task cancellation. The write-up and the failing-on-old-code test are appreciated. Requesting changes on two blocking points, both specific to how Tor differs from TCP here.

## 1. `copy_bidirectional` re-creates the truncation at the Tor protocol layer

`tokio::io::copy_bidirectional` shuts down the write half of the peer when one direction hits EOF. For the Tor stream, that shutdown is not a TCP half-close: Arti's `DataWriter::poll_shutdown` flushes and then closes the stream target with `sent_end = true` — see tor-proto 0.38 `client/stream/data.rs` (`poll_flush_impl` with `should_close: true`, ~lines 845–861), and tor-proto's own test asserting that closing the writer emits a **`RELAY_END` cell** on the wire (`client/circuit.rs`, `close_stream_helper`, ~1824–1857).

Tor streams are not half-closeable. An exit that receives END closes its TCP connection to the destination and tears down the stream. So in exactly the scenario this PR targets — client half-closes right after the request, response not yet sent — the new code sends END while the server is still thinking, and the response dies at the exit instead of in the local task cancellation. C-tor's own SOCKS edge handles application FIN by *deferring* END until both directions are finished, not by propagating it.

The in-memory duplex test can't catch this: `tokio::io::duplex` genuinely supports half-close, so `copy_bidirectional` looks correct over it. It models TCP, not Tor.

**Suggested shape** — a manual relay with asymmetric EOF handling:

- client→tor hits EOF: stop that copy and flush, but do **not** shutdown the Tor writer; keep relaying tor→client.
- tor→client hits EOF (exit sent END or the destination finished): shutdown the client's write half (TCP does support half-close) and end the relay.
- Either direction errors: end the relay.
- Dropping both streams at the end sends the END, at the right time.

For the test: keep the current one (it pins the local-cancellation bug), and add one with a wrapper stream around the "tor" side that fails the test if `poll_shutdown` is called before that side has itself reached EOF — that's the property `copy_bidirectional` violates.

## 2. As merged, the fix would be inert

The app links the prebuilt, checked-in `Frameworks/arti.xcframework` (see `Package.swift` `binaryTarget` and `docs/ARTI-BINARY-PROVENANCE.md`); this PR changes only the Rust source, so the shipped binary would still contain the `select!` code. No action needed from you beyond the code change — we'll run the provenance rebuild flow when this lands — but flagging it so nobody assumes a green CI run means the fix shipped. (CI currently doesn't compile or test this crate at all; that's a gap on our side we'll address separately.)

## Minor

- `relay()` now propagates errors instead of logging and returning `Ok(())` — that's fine (the accept loop in `lib.rs` logs per-connection failures), just noting it's a deliberate behavior change.

Happy to re-review quickly once the relay shape is reworked — the bug is real and worth fixing properly.

aashish-thapa:
@jackjackbits Thanks for the review.  reworked to an asymmetric relay per your shape.

Chessing234:
good catch on half-close. a short comment near the select! explaining why both directions must drain would help the next reader

Chessing234:
relaying until both halves close is the right fix for truncated responses. is there a stall timeout if the remote half-closes and never finishes the body?

aashish-thapa:
@Chessing234 

Thanks for looking. On the comment, there's one on `relay()` and one right above the loop. One correction though: it's not that both directions have to drain. It's asymmetric on purpose. The response path ends the relay; the request path finishing only stops that copy. Waiting on both would pin the connection open any time a client keeps its write half open after the response is already delivered.

On the stall:  no, there's no timeout. `copy(tor_read, client_write)` just waits  and `handle_socks_connection` is spawned bare in `lib.rs` with nothing around it. The old `select!` did return early in that case, but only because it was cutting the response short, so it was never really a timeout.

An idle timeout on the response path is the right shape; a total-duration one would kill the long-lived relay connections. @jackjackbits want that here or as a follow-up? It's a behavior change past the truncation fix, so I'd rather you call it.

PR #1498, opened 2026-07-26
BLE transport architecture V3: one engine domain, capability ports, feature-owned state

jackjackbits:
## What this is

The V3 transport restructuring (design + remaining roadmap in `docs/BLE-ARCHITECTURE-V3.md`). Eleven commits, each independently green (suite grows 1,949 → 1,964 tests, wall-clock unchanged).

## Concurrency model (commits 1–4)

- **Lock-backed peer registry + local announce state** — every main-actor Transport read (`isPeerConnected`, nicknames, snapshots, capability queries) reads a lock instead of blocking on `collectionsQueue` behind in-flight transport work.
- **Link buffers to bleQueue** — `pendingPeripheralWrites`/`pendingNotifications`/`pendingWriteBuffers` become bleQueue-confined next to the CoreBluetooth objects they feed; the notification drain stops invoking CB from a transport queue.
- **One serial engine queue** — the concurrent message queue + collections queue collapse into a single serial domain; every barrier flag and per-field ownership comment deletes (~98 cross-queue hops removed). `onEngine` documents and debug-enforces the sync-edge order `main → engine → bleQueue/crypto`; its trap caught two latent inversions during migration (announce-rebind fetching engine state from a bleQueue critical section; noise session-generation closures sync-re-entering the engine) — both fixed structurally.

## Review findings + a real product defect (commit 8)

All pass-6 findings are resolved: Periphery symbols gone, the two `messageQueue.sync` bypasses and the three panic fences route through `onEngine`, the registry ownership comments state the real writer set, and `BLEQueueContractTests` pins the ordering contract (only `onEngine` may sync-enter the engine; no `main.sync` from transport code; the collections queue stays deleted).

The `timeoutRestoredSession` CI flake exposed a genuine defect, fixed here rather than papered over: the capability-proof watchdog kept draining the parked outbound queues when it fired during a timeout-restore — encrypting them under restored keys the counterpart may have discarded, the exact silent loss `deferOutboundUntilConvergence` exists to prevent. Deferred peers are now tracked, the watchdog drain respects the rule, and the test fires the watchdog deterministically inside the deferred window.

## Feature-owned state (commits 6, 9)

`BLEMeshPingTracker` (/ping probes + per-link response budget) and `BLEPrivateMediaSessionStore` (the six generation-keyed private-media maps + convergence deferral, whole-transition methods under a leaf lock) — both with direct unit tests of their security invariants. The private-media store also removes the last routine main→engine sync reads and turns the noise-critical-section transitions into ordinary leaf-lock calls.

## App boundary (commits 5, 10)

No app code names `BLEService` anymore, and the god-protocol is split: `Transport` keeps only the core every transport implements; the mesh-only surface is eleven capability protocols (`MeshFileTransferring`, `MeshVoiceStreaming`, `MeshCourierTransporting`, `MeshGroupMessaging`, `MeshBoardBroadcasting`, `MeshDiagnosing`, `MeshVerifying`, `MeshPublicArchiving`, `MeshBridgingTransport`, `PanicResettingTransport`, `BluetoothStateReporting`) discovered with `as?`. The inert-defaults extension is deleted; call sites that leaned on a default keep its safe floor explicitly.

## What remains (documented, deliberately not in this PR)

The link-layer extraction behind `LinkEvent`/`LinkCommand` ports needs the link-auth boundary designed first (`noiseAuthenticatedLinkOwners` atomicity with bindings — see roadmap item 1), and the sans-I/O engine + simulator rides after it. Note: this PR must merge before #1521 (BLE power), which collides semantically in the scanPlan region.

## No wire changes

Packet formats, signing, identity binding, courier tags untouched.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -85,7 +85,8 @@ final class AppChromeModel: ObservableObject {
     /// neighbor claim but never announced to us) fall back to a short ID.
     func meshTopologyDisplayModel() -> MeshTopologyDisplayModel {
         let mesh = chatViewModel.meshService
-        guard let snapshot = mesh.currentMeshTopology() else { return .empty }
+        guard let diagnostics = mesh as? MeshDiagnosing,
+              let snapshot = diagnostics.currentMeshTopology() else { return .empty }
         let nicknames = mesh.getPeerNicknames()
 
         let nodes = snapshot.nodes.map { peerID -> MeshTopologyDisplayModel.Node in
--- a/bitchat/Services/BLE/BLEEngineScheduler.swift
+++ b/bitchat/Services/BLE/BLEEngineScheduler.swift
@@ -1,0 +1,39 @@
+import Foundation
+
+/// Schedules deferred engine work: relay jitter, announce delays, protocol
+/// deadlines (ping, capability proof), notification retry backoff, and
+/// fragment pacing.
+///
+/// This is the transport's only source of engine-side delay. Production
+/// wraps the engine queue's `asyncAfter`; tests inject a manually advanced
+/// clock so timer-driven behavior is asserted deterministically instead of
+/// racing the wall clock — product constants used as deadlines are exactly
+/// the hidden-elapsed-deadline flake class the test-timing hygiene rules
+/// exist to contain.
+protocol BLEEngineScheduling: AnyObject {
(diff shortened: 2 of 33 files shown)


yhyya11 (commented):
شجرة العمل-بل-الهندسة المعمارية-v3

PR #1499, opened 2026-07-27
Route favorite notifications through the MessageRouter outbox

vincenzopalazzo:
Stacked on #1408 — draft until it merges; review only the top commit (265028c). Will rebase onto main once the alias-flush machinery lands. Picks up the remaining kernel from #1040 (ref: https://github.com/permissionlesstech/bitchat/pull/1040#issuecomment-5085806219).

## Summary

- `MessageRouter.sendFavoriteNotification` was fire-and-forget: a favorite toggle sent while the peer was offline or mid-handshake was silently dropped — for mesh-only peers, a missed npub exchange.
- Now builds the same `[FAVORITED]:<npub>` / `[UNFAVORITED]:<npub>` payload the transports built internally and routes it through `sendPrivate`: the outbox retains it until an ack, and #1408's alias flush drains the queue on reconnect/authentication under either PeerID form — no key normalization needed.
- `MessageRouter` gains an optional `NostrIdentityBridge` to stamp our npub (previously stamped inside the transports' own `sendFavoriteNotification`), wired through `ChatViewModelBootstrapper`.
- Wire format unchanged: receivers already parse the prefix from PM content and never display it.
- Transport-level `sendFavoriteNotification` stays for the `UnifiedPeerService` mesh-only fallback when the router isn't wired yet.

Known semantic: a retained favorite that never gets acked eventually fires `onMessageDropped` like any unacked PM; its message ID backs no chat row, so the UI handler is a no-op.

## Test plan

- [x] `swift build --build-tests` + `swift test --skip-build --parallel --skip PerformanceBaselineTests` — 1958 tests green
- [x] Rewrote `sendFavoriteNotification_routesThroughOutboxAsPrivateMessage`: reachable peer → exactly one PM with the `[FAVORITED]` prefix, no transport-level favorite call
- [x] New regression test `sendFavoriteNotification_whileOffline_queuesAndFlushesOnReconnect`: toggle offline → nothing sent → `flushOutbox` on reconnect delivers one `[UNFAVORITED]` PM

🤖 Generated with [goose](https://github.com/block/goose)


Chessing234:
routing favorites through the outbox is cleaner. can the same favorited/unfavorited enqueue twice under both ephemeral and stable peer aliases after reconnect?

Chessing234 (commented):
the underlying bug is real — `sendFavoriteNotification` on main picks a transport and drops the notification on the floor if there isn't one, and for a mesh-only favorite that's a lost npub exchange. routing it through the outbox is the right shape. three consequences of reusing `sendPrivate` that i think need answering before this lands, none of which the body covers.

**a favorite toggle now gets handed to couriers.** `sendPrivate` falls through to the courier path for a peer that isn't securely reachable (`MessageRouter.swift:222-229`, then `depositToCouriers`, up to `maxCouriersPerMessage = 3`). so favoriting an offline peer now seals `[FAVORITED]:<your npub>` to that peer and hands it to three third parties to physically carry, plus a bridge drop when enabled. the payload is sealed, but the fact that you have mail for that specific noise key is not, and the user's mental model of "i tapped a star" doesn't include recruiting couriers. main's fire-and-forget did none of this. if the intent is only "survive a reconnect", the outbox retention gets you that without the courier hop — worth either scoping favorites out of `depositToCouriers` or saying explicitly that this is intended.

**each toggle takes a slot in a 100-deep FIFO that evicts real DMs.** every call builds a fresh `UUID().uuidString`, so toggles never coalesce the way a resend of the same ID does in `enqueue` (`:468`). toggle a favorite on and off 50 times while the peer is offline and `queue.count > maxMessagesPerPeer` starts evicting from the front — and the front is the oldest mail, i.e. the user's actual queued messages, dropped via `dropMessage` with `onMessageDropped`. control-plane traffic and user messages now share one bounded queue with no priority between them. keying the favorite on a stable per-peer message ID (so a re-toggle replaces rather than appends, which `enqueue` already supports) would collapse that to one slot and would also make the last write win, which is the semantics you actually want for a boolean.

**the plaintext now goes to disk.** `MessageOutboxStore`'s own header says "Nothing else in the app persists message plaintext" and that it keeps the key in the keychain for exactly that reason. after this, toggling a favorite writes `[FAVORITED]:<npub>` plus the peer's noise key into that store. it's encrypted at rest, so this is a note rather than an objection — but it means a favorite is now a durable record of "this identity favorited this peer", and it's worth one line in the body since the store's comment is currently an accurate statement about DMs only.

one thing i checked and am satisfied with: the Nostr path doesn't regress. `NostrTransport.sendFavoriteNotification` builds `[FAVORITED]:<npub>` and embeds it through `NostrEmbeddedBitChat.encodePMForNostr`, which is the same thing `sendPrivateMessage` does with the same content, so routing through `sendPrivate` produces the same wire bytes. `recipientNickname: ""` also matches what `BLEService.sendFavoriteNotification` already passed.

and a smaller one on the refactor: the `PeerMessageKey` comment now says "Both properties are read directly now — `peerID` by the alias-scoped sweep in `flushOutbox(forAliases:)`". that's true on this branch, but it makes a periphery-directive removal depend on a specific call site continuing to exist. if that sweep is ever inlined the ignore has to come back, which is exactly the kind of thing the original comment existed to record. i'd keep the directive and the explanation and just add the new reads to it.

(reviewed the top commit only, as asked — this is still stacked on #1408.)

PR #1500, opened 2026-07-27
Consolidate base64url helpers into a single BitFoundation utility

### LINKED ISSUE #642: Chore: Consolidate `base64url` helpers into a single utility
# Chore: Consolidate `base64url` helpers into a single utility

## Problem:

Multiple `base64url` encode/decode implementations exist across the codebase, increasing drift and edge-case bugs.

- Duplicated logic in `NostrProtocol`, `NostrEmbeddedBitChat`, `ChatViewModel`.
- Padding handling differs per call site.
- Harder to audit/patch for security or correctness.
- Extra test and maintenance surface area.

## Proposed Solution:

Create a shared `base64url` utility and migrate call sites.

1. Add `Protocols/Base64URL.swift` with `encode(_:)` and `decode(_:)`.
2. Replace local helpers/usages in `NostrProtocol`, `NostrEmbeddedBitChat`, `ChatViewModel`.
3. Add minimal unit tests for padding/invalid inputs.
4. Update docs/comments to reference the single source.

## Benefits:

Consistency and fewer edge cases in encoding paths.

- Single, auditable implementation.
- Reduced maintenance and code drift.
- Clearer error handling for malformed inputs.

viditkulsh:
## Problem

Four hand-rolled base64url implementations existed: `Base64URLCoding` in the Nostr folder, private duplicates in `NostrEmbeddedBitChat` and `CashuTokenDecoder`, and an inline `replacingOccurrences` chain in `VerificationService`. Padding handling differed per call site — exactly the drift #642 warns about.

## Change

- Move `Base64URLCoding` into BitFoundation (the shared-components package) and make it public.
- Route all four call sites through it and delete the duplicates (net −20 lines of app code).
- The unified `decode` keeps `CashuTokenDecoder`'s padding normalization — strip any `=` then re-pad, reject length ≡ 1 (mod 4) — because external wallets emit both padded and unpadded forms. It accepts a superset of what the old Nostr-side decode accepted and returns identical results for every input the old decoder handled. Encoding is byte-for-byte identical to all three prior copies, so no wire format changes.

## Tests

New `Base64URLCodingTests` cover alphabet substitution, padded and unpadded input, every padding length round-trip, and malformed-input rejection (`len % 4 == 1`, invalid characters, embedded whitespace).

Verified locally with `swift build` on BitFoundation plus test vectors cross-checked against an independent implementation; full suite runs in CI.

Fixes #642


jackjackbits (changes requested):
The consolidation itself verified out completely — every replaced site is wire-safe (courier tags stay hex, gift-wrap stays standard base64, the base64url encoder is character-identical to each deleted private helper). One ask, which is also what all four red CI jobs trace back to:

**`bitchatTests/Nostr/Base64URLCodingTests.swift` is stale** — it still references the type at its old location, but the utility now lives in BitFoundation, so the test target no longer compiles (hence the BitFoundation, app, iOS-sim, and Periphery failures). Update or move that test file alongside the type and CI should go green.

FYI: #1529 was closed as a duplicate in favor of this PR, so once this ask is addressed we can move quickly.

Chessing234:
clean chore. double-check Cashu still accepts the legacy dotted charset after the shared helper swap

Chessing234:
jack’s ask checks out — Base64URLCodingTests.swift still imports the old app-target path after the type moved into BitFoundation, which is why those four CI jobs fail. moving/updating that test file next to the BitFoundation type should clear it.

Chessing234 (commented):
i ran the two decoders side by side against real Foundation rather than reading them, and the "identical results for every input the old decoder handled" claim holds — 3000 random payloads round-tripped through canonical encoding give byte-identical output from both. so the consolidation itself is safe.

but the widening isn't only padding, and the body doesn't mention the part that matters. `=` is stripped globally, not just from the tail, so interior padding characters are now accepted:

```
input  3q2-7wEC        old deadbeef0102   new deadbeef0102
input  3q2-7wEC==      old deadbeef0102   new deadbeef0102
input  3q=2-7wEC       old nil            new deadbeef0102
input  3=q=2=-=7=w=E=C old nil            new deadbeef0102
```

(trailing `=` was already accepted on main — `Data(base64Encoded:)` tolerates the over-padding the old re-pad produced — so that half of the "wallets emit both forms" rationale was already true.)

the consequence is that an unbounded number of distinct strings now decode to the same bytes, at two network-facing call sites: `NostrProtocol.swift:641` (`v2:` private envelope) and `NostrInboundPipeline.swift:581` (`bitchat1:` embedded packet). both producers are bitchat itself and both emit canonical unpadded output, so leniency buys nothing there and costs the property that one packet has one encoding. i'd split it: a strict `decode` for the wire formats (reject any `=`, reject `len % 4 == 1`) and an explicit lenient entry point for `CashuTokenDecoder`, where the input is a user-pasted token from a third-party wallet and leniency is the point.

the length bounds in front of both call sites are unaffected either way — stripping only shortens the string, so `encoded.count <= maxEncoded` and `ciphertext.utf8.count <= maximum...` stay conservative.

two smaller things:

the branch is CONFLICTING, and it's a one-liner: #1645 landed a `guard status == errSecSuccess` immediately above the `nonceB64` line you're replacing in `VerificationService.swift:87-89`. that's the only conflicting hunk.

and #642 asks for a single utility — there's a fifth copy at `bitchatTests/CashuTokenDecoderTests.swift:20`. it's test-only, but it's the copy that would keep a future encoder change from being caught, since the tests would drift with it.

PR #1504, opened 2026-07-27
Add keyboard navigation for command suggestions on macOS

### LINKED ISSUE #233: Add arrow key navigation for command suggestions (macOS)
Currently you can only click on suggestions. This would add keyboard navigation similar to other autocomplete interfaces.

viditkulsh:
## Problem

The command suggestion panel is click-only (#233). Every other autocomplete surface supports keyboard navigation.

## Change

On macOS, arrow keys now move a highlight through the suggestions, and return/tab insert the highlighted command (with the trailing space, ready for arguments). Typing continues to filter as before.

Implementation note: arrow keys never reach SwiftUI modifiers while the composer's single-line field editor has focus — it consumes `moveUp:`/`moveDown:` itself — and the macOS 13 deployment target rules out `.onKeyPress`. Navigation therefore uses an `NSEvent` local monitor installed while the panel is visible and removed when it disappears. The monitor is narrowly scoped:

- only plain (unmodified) up/down/return/tab are consumed, and only while the selectable list is showing;
- the usage-reminder row (shown once arguments have begun) stays informational — return passes through and sends the composed command;
- iOS behavior is untouched.

The selection highlight resets whenever the filtered list changes, so it never outlives its list.

## Status

Draft until I attach a screen recording from a device run; logic is complete and the existing `ViewSmokeTests` mount exercises the monitor install/remove lifecycle.

Fixes #233

### DIFF
--- a/bitchat/Views/Components/CommandSuggestionsView.swift
+++ b/bitchat/Views/Components/CommandSuggestionsView.swift
@@ -6,6 +6,9 @@
 //
 
 import SwiftUI
+#if os(macOS)
+import AppKit
+#endif
 
 struct CommandSuggestionsView: View {
     @EnvironmentObject private var privateConversationModel: PrivateConversationModel
@@ -13,6 +16,16 @@ struct CommandSuggestionsView: View {
     @ThemedPalette private var palette
 
     @Binding var messageText: String
+
+    /// Row highlighted for keyboard navigation (macOS). Reset whenever the
+    /// filtered list changes so the highlight never outlives its list.
+    @State private var selectedIndex = 0
+    #if os(macOS)
+    /// Arrow keys never reach SwiftUI while the composer's field editor has
+    /// focus (the single-line field consumes moveUp:/moveDown: itself), so
+    /// navigation uses a local key monitor scoped to the panel's lifetime.
+    @State private var keyMonitor: Any?
+    #endif
 
     /// The command already typed in full, once arguments have begun.
     private var typedCommandAlias: String? {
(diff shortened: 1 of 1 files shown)


Chessing234:
handy for #233. maybe keep return-to-send when the panel isnt focused / has no highlight, so plain send still works

jackjackbits:
update now that the mention-panel sibling landed: #1542 merged with a reusable `AutocompleteKeyboardNavigationModifier` (ContentComposerView.swift), so the asks for this PR when you pick it back up:

1. **reuse the shared modifier instead of a private monitor in CommandSuggestionsView** — one mechanism for both panels. pass `isActive: { !filteredCommands.isEmpty }`, `onMove`/`onAccept`/`onDismiss`, and delete the local `keyMonitor` plumbing.
2. **one landmine to respect if you adapt rather than reuse**: the modifier's `isActive` is a `() -> Bool` on purpose. #1542's first version took a plain `Bool`, and the monitor closure froze the value captured at install time — the panel that opened *after* install never intercepted a key. your current code is actually immune (you recompute `filteredCommands` through the binding on every event — the right instinct), but any refactor that passes visibility as a value reintroduces it.
3. **key-convention convergence, now pinned by the merged behavior**: arrows move (mentions wrap; this PR clamps — please wrap to match), return/tab accept, **escape dismisses** (currently unhandled here). one wrinkle specific to commands: when the typed text already equals the highlighted suggestion, accept is a no-op — return should fall through to send in that case, or a fully-typed `/clear` costs a dead keypress (this also addresses the earlier keep-return-to-send comment).
4. un-draft + rebase onto current main when ready — the composer file moved under it.

Chessing234 (commented):
the diagnosis in the body is right and #233 is still open, so this is worth having. the problem is that main grew a general version of exactly this mechanism after you opened it, and this branch predates it.

`AutocompleteKeyboardNavigationModifier` in `ContentComposerView.swift:415` landed with #1542 on 1 aug. it is the same NSEvent local monitor, the same keycodes (126/125/36/48), the same "return nil consumes the event" contract — and its doc comment credits this pr twice ("the same reason command suggestions (#1504) use an `NSEvent` local monitor" and "Standard autocomplete navigation (aligned with #1504)"). so the seam you'd want already exists, it's already applied to the composer's TextField at line 84, and it takes `isActive` / `onMove` / `onAccept` / `onDismiss` closures. rebasing onto that instead of hand-rolling a second monitor is a much smaller diff and keeps the two panels on one convention, which is what the comments on both sides now promise.

adopting it also fixes two things this branch is missing:

**escape doesn't dismiss.** the modifier handles keycode 53; `handleKeyDown` here doesn't. once you've typed `/` the panel is up and the keyboard has no way to get rid of it — you have to delete back past the slash. that's the one navigation key a user will reach for after arrows.

**return and tab are consumed unconditionally.** here it's `case 36, 48: accept(...); return nil`, where the modifier is `return onAccept() ? nil : event`. the difference matters if `accept` can ever be a no-op — and it's the pattern the composer's `.onSubmit` at line 98 is already written against (`if ... completeSelectedSuggestion(...) { return }` then fall through to send).

i did check the case i expected to break and it's fine: with `/msg @al` in the composer, both monitors are installed at once — the mention one on the TextField and yours on the panel — but `typedCommandAlias` is non-nil there so `handleKeyDown` returns the event and the mention monitor gets it. so the two don't currently fight. it's still two monitors where main now has one.

one thing that applies to both implementations rather than just this one, so treat it as an observation: `addLocalMonitorForEvents` is app-wide and neither handler looks at `event.window`. while the panel is visible, a plain tab or return anywhere else in the app — another window, a sheet — is swallowed. probably worth a `guard event.window === <composer window>` in the shared modifier rather than fixing it twice.

smaller: `selectedIndex` and the `.onChange` reset are compiled on iOS too, where `rowHighlight` never reads them, so the state is written and never used off macOS. easy to fold into the `#if os(macOS)` block.

on the status line — this is still a draft and the recording isn't attached, but CI is fully green on the current head (all seven checks), so the only thing between it and review is the rebase.

PR #1508, opened 2026-07-27
fix: enable autocorrect in the message composer

### LINKED ISSUE #969: Enable auto correction when typing a message
Currently it seems to be disabled

Chessing234:
## Summary
- composer autocorrect is **token-aware** (#969): on for prose, off while the current token starts with `/`, `@`, or `#`
- avoids fighting `/commands`, `@mentions`, and `#channels`, and keeps those tokens out of keyboard learning
- nickname / geohash fields elsewhere stay fully autocorrect-disabled

Closes #969

## Test plan
- [ ] type a normal sentence — autocorrect suggestions appear
- [ ] type `/hel` — no autocorrect rewriting the command
- [ ] type `hi @al` — no autocorrect on the mention token; prose before it still fine
- [ ] type `#u4` — channel token left alone
- [ ] after `@alice ` (trailing space) — autocorrect re-enabled for the next word

### DIFF
--- a/bitchat/Utils/ComposerAutocorrect.swift
+++ b/bitchat/Utils/ComposerAutocorrect.swift
@@ -15,6 +15,10 @@ import Foundation
 /// of those sigils; leave it on otherwise. The composer TextField only exposes
 /// the text (not a live selection), so callers pass the caret — typically the
 /// end of the string, matching autocomplete.
+///
+/// Mid-string caret is a known v1 limitation: when callers only know
+/// `text.count`, moving the caret back into an existing `/` `@` `#` token
+/// still leaves autocorrect on. Plumb selection through when SwiftUI exposes it.
 enum ComposerAutocorrect {
     /// Sigils that mean "exact token, don't rewrite me".
     static let specialPrefixes: Set<Character> = ["/", "@", "#"]
--- a/bitchat/Utils/ComposerAutocorrect.swift
+++ b/bitchat/Utils/ComposerAutocorrect.swift
@@ -1,0 +1,45 @@
+//
+// ComposerAutocorrect.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Foundation
+
+/// Token-aware autocorrect policy for the message composer (#969).
+///
+/// Autocorrect helps with prose but fights `/commands`, `@mentions`, and
+/// `#channels`. Disable it while the token under the cursor starts with one
+/// of those sigils; leave it on otherwise. The composer TextField only exposes
+/// the text (not a live selection), so callers pass the caret — typically the
+/// end of the string, matching autocomplete.
+enum ComposerAutocorrect {
+    /// Sigils that mean "exact token, don't rewrite me".
+    static let specialPrefixes: Set<Character> = ["/", "@", "#"]
+
+    /// Whether `.autocorrectionDisabled` should be on for this caret position.
+    static func shouldDisable(for text: String, cursorPosition: Int) -> Bool {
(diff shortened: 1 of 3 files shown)


jackjackbits (changes requested):
A blanket autocorrect enable conflicts with how the composer is used: it's also the command line, and autocorrect fights /commands, @mentions, and #channels (see the maintainer guidance in #969, which prescribes token-aware behavior). There's also a privacy dimension — enabling autocorrect feeds message content into the system keyboard's learning.

**Ask: token-aware autocorrect** — disabled while the current token starts with /, @, or #, enabled otherwise. That gets the typing comfort without breaking command entry. If SwiftUI's per-token control is too coarse, disabling when the message starts with "/" covers the worst of it.

Chessing234:
switched to token-aware autocorrect — off while the current token starts with /, @, or #. ci green.

Chessing234:

vincenzopalazzo (commented):
**Production Readiness: GO** — pending a quick on-device check of the keyboard behavior noted below.

The token-aware policy is small, well-scoped, and the sigil set matches how the composer actually parses commands/mentions/channels.

### Issues & Suggestions

- **Medium:** `bitchat/Views/ContentComposerView.swift` (~L74-76) — `shouldDisableAutocorrect` flips `.autocorrectionDisabled` and `.textInputAutocapitalization` on the live first responder as the caret crosses token boundaries. UIKit applies changed text-input traits to an active field by reloading them, which on some iOS versions produces visible keyboard flicker or a brief first-responder reset. Please verify on-device: type `hello @` back and forth across the boundary and watch the keyboard. If it flickers, consider only updating the traits when the computed value actually changes (it already does) plus debouncing, or applying via a UIKit representable.
- **Cleanup:** the caret is always passed as `messageText.count` (end of string), so moving the caret back into an existing `/`, `@`, or `#` token mid-string still gets autocorrect. The docstring calls this out, so it's a known limitation — fine for v1, worth a tracking note if mid-string editing of mentions matters.

### Summary

Good fix for #969; the only thing between this and merge is the on-device keyboard check.

---

Chessing234:
thanks for the readiness pass — pushed a short debounce on the autocorrect/capitalization trait flip so crossing `/` `@` `#` doesn't thrash uikit on every keystroke. also documented the mid-string caret limit you called out. i don't have the physical device rig for the flicker check here; if you or jack can tap `hello @` once on a phone that would close the last gate.

Chessing234:
debounce for trait flips is on the branch now — still only needs that quick on-device `hello @` keyboard check vincenzo called out.

jackjackbits (changes requested):
the debounce commit is sound and addresses the trait-thrash concern from the last round — one thing blocks it.

1. the periphery red is this PR's own: `shouldDisableAutocorrect` (ContentComposerView.swift:139-142) lost both use sites to `appliedAutocorrectDisabled` and is now dead — delete it.
2. nit: `autocorrectTraitTimer` isn't invalidated in `.onDisappear` (only the autocomplete timer is) — a pending 0.12s tick can fire after the view is gone; invalidate it alongside.

still want one on-device keyboard pass before merge, per the earlier review.

Chessing234:
Addressed the latest review:

1. Deleted dead \`shouldDisableAutocorrect\` (use sites already go through \`appliedAutocorrectDisabled\`).
2. Invalidate \`autocorrectTraitTimer\` alongside the autocomplete timer in \`.onDisappear\`.

Still need an on-device keyboard pass before merge — I don’t have that hardware in this loop; happy to re-check anything that looks off after you poke it.

Chessing234:
rebased onto main. jack's two asks were already on tip — dead `shouldDisableAutocorrect` gone, `autocorrectTraitTimer` invalidated in onDisappear. ready for another look.

PR #1512, opened 2026-07-27
fix: add setting to disable BLE wake-on-proximity

### LINKED ISSUE #1427: Random/intermittent modal - "The 'iPhone' accessory would like to open 'bitchat'"
I've been seeing a random/intermittent modal - "The 'iPhone' accessory would like to open 'bitchat' since the last release, and also on the current release. Is this expected? My iPhone might be in developer mode / enabled for development. Not sure if that is relevant.

Image

This is on iOS 26.5.2

Chessing234:
## Summary
- add a privacy setting to opt out of background wake-on-proximity pending BLE connects
- default stays on so the mesh remains reachable while backgrounded
- addresses the iOS 26 "accessory would like to open bitchat" prompt from #1396

Closes #1427

## Test plan
- [ ] leave the setting on — backgrounding still arms pending connects as before
- [ ] turn the setting off in App Info → Privacy — backgrounding no longer arms pending connects
- [ ] confirm the toggle persists across relaunch

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -3349,6 +3349,378 @@
         }
       }
     },
+    "app_info.settings.proximity_wake.subtitle" : {
+      "comment" : "Subtitle explaining wake-on-proximity and the ios 26 accessory open prompt",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "عندما يكون bitchat في الخلفية، اطلب من النظام إعادة الاتصال بأقران الشبكة الحديثين حتى يتمكن هاتف قريب من إيقاظ التطبيق. على ios 26 قد يظهر تنبيه \"يريد ملحق فتح bitchat\" — آمن للسماح، أو عطّله إن كنت تفضّل عدم رؤيته. مفعّل افتراضيًا لتبقى الشبكة قابلة للوصول."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "bitchat ব্যাকগ্রাউন্ডে থাকলে সিস্টেমকে সাম্প্রতিক মেশ পিয়ারদের সাথে আবার সংযুক্ত হতে বলুন যাতে কাছের ফোন অ্যাপ জাগাতে পারে। ios 26-এ সেই জাগানোতে \"অ্যাকসেসরি bitchat খুলতে চায়\" প্রম্পট দেখা যেতে পারে — অনুমতি দেওয়া নিরাপদ, অথবা দেখতে না চাইলে বন্ধ করুন। মেশ পৌঁছানো যায় রাখতে ডিফল্টে চালু।"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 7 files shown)


jackjackbits (changes requested):
The rework is in good shape — relocating the guard into `BLERadioController.armPendingBackgroundConnects` after the V3 merge was exactly right: it's the single choke point both call sites go through, and defaulting the setting to ON preserves mesh reachability. One blocker left:

**Catalog entries for the two new strings.** `app_info.settings.proximity_wake.title` and `.subtitle` have no `Localizable.xcstrings` entries at all — every non-English locale will show English. This passes CI silently today (`LocalizationCoverageTests` only validates keys already present in the catalog), and then breaks CI on the next catalog regeneration. Please add entries for all locales — same ask we've made on the sibling PRs.

Two non-blocking nits:
- No `BLEProximityWakeSettings.reset()` in the panic-wipe path — inconsistent with `MeshEchoSettings`, `NotificationPrivacySettings`, and `NostrRelaySettings`, which all reset on wipe. One line plus a test.
- Toggling off doesn't cancel already-armed pending connects, only prevents future arming (takes effect on the next background transition). Acceptable in practice since you can't toggle while backgrounded — a code comment would prevent someone "fixing" it later.

Heads-up: #1509 also adds a settings card to `AppInfoView`; whichever of the two merges second will need a trivial rebase.

Chessing234:
Addressed:

1. **Catalog entries** for `app_info.settings.proximity_wake.title` / `.subtitle` across all 30 locales.
2. **`BLEProximityWakeSettings.reset()`** on the panic-wipe path (plus a small settings test).
3. Comment on `armPendingBackgroundConnects` clarifying that toggling off only prevents future arming.

Ready for another look.

Chessing234:
localization + panic reset are on the branch now; also noted that toggling off only affects the next background arm. ci green.

Chessing234:
@jackjackbits Re-review round — tip \`91bb72b\`:

**Addressed**
- \`app_info.settings.proximity_wake.title/subtitle\` added to \`Localizable.xcstrings\` for all 30 locales.
- \`BLEProximityWakeSettings.reset()\` wired into panic wipe (\`ChatViewModel.panicClearAllData\`) + unit test.
- Comment + \`cancelPendingWakeConnects()\` on opt-out / settings notification — already-armed pending connects are cancelled, not just future arming.

CI green on tip.

Chessing234:
@jackjackbits ready for another look — addressed your leftover asks on tip:

1. `proximity_wake` title/subtitle are in `Localizable.xcstrings` for all locales.
2. `BLEProximityWakeSettings.reset()` is on the panic-wipe path (with test).
3. Disabling proximity wake also cancels in-flight pending connects (with comment).

Chessing234:
proximity_wake title/subtitle are in the catalog on tip — ready for another look.

Chessing234:
@jackjackbits re-review please — blocker + nits addressed (e8625b00, ed13ba26):

- i18n: app_info.settings.proximity_wake.title / .subtitle now have real entries ×30 locales (de "aufwachen wenn nahe peers zurückkehren", ja "近くのピアが戻ったら起動する", ar "الإيقاظ عند عودة الأقران القريبين").
- panic wipe: BLEProximityWakeSettings.reset() is now called in the wipe path (ChatViewModel.swift:1640), consistent with MeshEchoSettings / NotificationPrivacySettings / NostrRelaySettings — with a test.
- in-flight connects: toggling the setting off cancels already-armed pending connects (ed13ba26), not just future arming — with a code comment so it isn't "fixed" later.

Chessing234:
rebased onto main and dropped an empty ci-retrigger commit. the conflict was the string catalog; resolved it by re-adding just the two proximity-wake keys to main's copy rather than carrying the old file, so it stays +372/-0 there. suite green locally, 2018 → 2022.

Chessing234:
tip already has the catalog entries ×30, panic-wipe reset + test, and the toggle-off / in-flight connect comment (plus cancel on disable). still on main, ci green — ready for re-review.

PR #1514, opened 2026-07-27
fix: normalize nicknames to Unicode NFC for matching

### LINKED ISSUE #214: Unicode normalization mismatch for usernames & message filtering
Nicknames or filters using accents (like café) typed differently (café) may not match.

Chessing234:
## Summary
- store local nicknames in Unicode NFC form
- compare nicknames / mentions / geo lookups with an NFC + casefold key so `café` (NFC) and `café` (NFD) resolve to the same peer

Closes #214

## Test plan
- [ ] set nickname to an accented name — persists in NFC
- [ ] `/msg` / mention a peer whose name was typed in the other normalization form
- [ ] `NicknameUnicodeTests` pass

### DIFF
--- a/bitchat/Services/AutocompleteService.swift
+++ b/bitchat/Services/AutocompleteService.swift
@@ -55,10 +55,10 @@ final class AutocompleteService {
         
         let fullRange = match.range(at: 0)
         let captureRange = match.range(at: 1)
-        let prefix = nsText.substring(with: captureRange).lowercased()
+        let prefix = nsText.substring(with: captureRange).bitchatNicknameKey
         
         let suggestions = peers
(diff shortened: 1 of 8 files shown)


jackjackbits:
Closing in favor of #1502, which normalizes nicknames to NFC at the storage and ingest boundaries (including migrating persisted geo-nicknames) rather than at comparison time. Two issues with the approach here: Swift's `==` on String already compares canonical equivalence, so the added comparison-time normalization is a no-op for the reported symptom; and the `lowercased()` in the matching key would make `/msg` resolve to the wrong peer when two nicknames differ only by case. One of the new tests also fails in CI. Thanks for flagging the Unicode angle — the fix is landing via #1502.

Chessing234:
ok np

PR #1515, opened 2026-07-27
fix: Stop a spoofed fragment from truncating an in-flight reassembly

FayezBast:
## The problem

Fragment packets are unauthenticated and bypass the packet deduplicator, so any peer in
range can emit a `.fragment` addressed to a stream it does not own — both halves of
`BLEFragmentKey(sender, fragmentID)` are attacker-choosable. `BLEFragmentAssemblyBuffer.append`
trusted every arriving header rather than the first one it accepted, which gave three primitives:

**1. Truncation — one packet.** Completion and reassembly both read `header.total` off the
arriving fragment:

```swift
fragments.count == header.total     // completion gate
(0..<header.total).reduce(...)      // reassembly range
```

A stream holding indices 0–2 of ten receives one injected fragment with `total = 3` at any
index below 3. The buffer declares `.complete` and hands the receive pipeline a truncated
payload as though it were the whole message.

**2. Assembly wipe — one packet.** The size ceiling was read off the arriving header too
(`assemblyLimit(for: header.originalType)`), and exceeding it wiped the entire assembly. No
accumulation is needed to reach it: `BinaryProtocol` accepts a compressed payload up to
`maxFramedFileBytes` at a ratio up to 50000:1, so a ~25-byte fragment on the wire legally
inflates to ~1.1 MiB of `fragmentData` — over `maxPayloadBytes` on its own. Separately,
`projectedSize` counted a duplicate index's bytes again even though storing at that index
replaces rather than adds, so replayed honest duplicates could reach the same wipe.

**3. Index poisoning.** `fragmentsByKey[key]?[index] = header.fragmentData` overwrote
unconditionally — a duplicate at a held index carrying different bytes replaced the honest
bytes, last-wins.

## What this changes

`(total, originalType, isBroadcast)` are pinned in `Metadata` when a stream's first fragment
is accepted. Every later header is checked against them and rejected as `.conflicting` without
mutating state; completion, the reassembly range, and the size ceiling all read pinned metadata.
An index already held is immutable — a duplicate carrying different bytes is rejected, not
applied. `projectedSize` counts only bytes that will actually be newly stored, and an oversized
fragment no longer destroys an assembly it did not start.

This rejects nothing a conformant peer emits: a real sender restates the same `total`,
`originalType`, and scope on every fragment of a stream, and retransmits identical bytes
per index.

## What this does not change

**It does not make reassembly unforgeable.** An earlier revision of this PR claimed an attacker
could no longer corrupt an assembly, only destroy it. That was wrong and is removed. First-wins
protects indices that are *already held*. A fragment that reaches an index before the honest one
does is the first accepted value for that index, and the stream reassembles around it —
`HON0 HON1 HON2 EVL3 HON4` still yields the attacker's bytes at index 3.

What changes is which attacks stay cheap. Corrupting *part* of a payload is still one packet — the
attacker wins the race to an index the sender has not filled yet, and those bytes are used. What is
no longer cheap is premature completion and full replacement: the attacker must supply the stream's
pinned `total`, cannot shorten it, and must win that race at every index it wants to control.
Destroying a stream is no longer possible at all — an injected fragment can no longer wipe or
truncate an assembly, only fail to be stored.

The remainder is not closable at the receiver. With unauthenticated fragments there is nothing to
tell the variants apart. Closing it needs authenticated fragment envelopes or a sender-key-bound
stream ID — the same class of work as the peer-ID-rotation spec in #1487. The type doc on
`BLEFragmentAssemblyBuffer` states this scope so the code stops overselling itself; happy to send
the matching §5.3 wording to #1519.

## Relay behavior is deliberately unchanged

An earlier revision gated relay and gossip archiving on local assembly state. That was a net loss
and is dropped. Losing the race locally is not evidence a fragment is forged — only that it arrived
second *here*. Withholding it would let a poisoned relay both spread the injected copy and censor
the honest one, including from peers that never saw the conflict and from `REQUEST_SYNC` recovery.
Against a one-shot injector that is strictly worse than last-wins relay; it only helps against a
persistent flooder. First-wins stays where it belongs — this node's storage layer.

## Two smaller holes closed along the way

- **A fragment too large to ever be stored is rejected before `prepareAssembly` runs.** Starting an
  assembly evicts the oldest in-flight stream to make room, so the inflate-to-1.1 MiB packet above
  was otherwise still a one-packet eviction primitive against a full table.

- **Streams that hit their ceiling stop drawing `REQUEST_SYNC`.** They are retained rather than
  discarded now, but cannot complete on what they hold, so they should not burn wire-filter slots.
  The mark is provisional, not a verdict: the fragment that tripped the ceiling may be the injected
  one, so any later fragment that does store clears it. Without that, one injected oversize packet
  would permanently disable recovery for a healthy stream — the same collateral-suppression shape
  as the relay gate above.

## Tests

- Each pinned field (`total`, `originalType`, broadcast scope) is rejected on conflict, and the
  original assembly still completes afterwards.
- A conflicting duplicate cannot overwrite a held index; an exact duplicate is not counted twice
  toward the budget.
- An oversized fragment at an unused index does not discard the assembly it targeted.
- An oversized first fragment is rejected without evicting the oldest stream from a full table.
- An over-budget stream is excluded from `stalledBroadcastFragmentIDs`, and regains eligibility
  once a fragment stores.
- Two senders sharing a fragment ID stay isolated. This passes only because `BLEFragmentKey`
  carries the sender; without that field the new first-wins checks would turn an ID collision into
  mutual rejection, so it is worth pinning explicitly.

### DIFF
--- a/bitchat/Services/BLE/BLEFragmentAssemblyBuffer.swift
+++ b/bitchat/Services/BLE/BLEFragmentAssemblyBuffer.swift
@@ -53,19 +53,61 @@ struct BLEFragmentHeader: Equatable {
     }
 }
 
+/// Reassembles inbound fragment streams keyed by `(sender, fragmentID)`.
+///
+/// Fragments are unauthenticated and bypass the packet deduplicator, so any
+/// peer in range can address a stream it does not own. The buffer treats the
+/// first accepted header as authoritative for the stream's shape and refuses
+/// to let a later fragment restate it or rewrite an index already held.
+///
+/// This bounds the damage; it does not eliminate it. An injected fragment
+/// that reaches an index before the honest one does *is* the first accepted
+/// value for that index, and one such packet is enough to corrupt the
+/// reassembled payload. What first-wins buys is cost against premature
+/// completion or full replacement: the attacker must supply the stream's
+/// pinned total and win the race at every index it needs to control. Closing
+/// the gap needs authenticated fragment envelopes or a sender-key-bound
+/// stream ID — the receiver has nothing to distinguish the variants by.
 struct BLEFragmentAssemblyBuffer {
+    enum ConflictReason: Equatable {
+        case total(expected: Int, actual: Int)
+        case originalType(expected: UInt8, actual: UInt8)
+        case broadcastScope(expected: Bool, actual: Bool)
+        case fragmentData(index: Int)
+    }
+
     enum AppendResult: Equatable {
         case stored(header: BLEFragmentHeader, started: Bool)
         case complete(header: BLEFragmentHeader, reassembledData: Data, started: Bool)
         case oversized(header: BLEFragmentHeader, projectedSize: Int, limit: Int, started: Bool)
+        case conflicting(header: BLEFragmentHeader, reason: ConflictReason)
     }
 
     private struct Metadata {
         let total: Int
+        let originalType: UInt8
         let timestamp: Date
         let isBroadcast: Bool
         var lastFragmentAt: Date
         var lastResyncRequestAt: Date?
+        /// Set when a fragment would have pushed the stream past its size
+        /// limit. The assembly is kept (see `append`) but cannot complete on
+        /// what it holds, so it stops consuming REQUEST_SYNC filter slots
+        /// until a fragment actually stores and clears the mark.
+        var exceededBudget: Bool = false
+
+        func conflictReason(for header: BLEFragmentHeader) -> ConflictReason? {
+            if header.total != total {
+                return .total(expected: total, actual: header.total)
+            }
+            if header.originalType != originalType {
+                return .originalType(expected: originalType, actual: header.originalType)
+            }
+            if header.isBroadcastFragment != isBroadcast {
+                return .broadcastScope(expected: isBroadcast, actual: header.isBroadcastFragment)
+            }
+            return nil
+        }
     }
 
     private var fragmentsByKey: [BLEFragmentKey: [Int: Data]] = [:]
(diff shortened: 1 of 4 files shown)


jackjackbits (changes requested):
First: we verified all three bugs empirically with a side-by-side harness against main, and they're real — actually worse than the PR claims. One-packet truncation and the oversize-wipe each need only a single tiny packet (a 25-byte compressed fragment inflates to 1.1 MiB via the 50000:1 ratio ceiling and wipes any in-flight assembly, no race needed). Pinning `(total, originalType, isBroadcast)` at first-accept, reading pinned metadata for completion and the ceiling, and counting only stored bytes is the right fix, and it rejects nothing a conformant peer emits. Dropping the barrier hop is a no-op (that branch was dead). We want this to land. Three things first:

**1. Rebase on main.** #1498 (BLE V3) moved `handleFragment` within `BLEService.swift` (~6886 → 6637), but the code your diff targets is verbatim intact — `BLEFragmentAssemblyBuffer.swift`, `BLEFragmentHandler.swift`, and both test files merge clean. The conflict is purely textual; a mechanical rebase lands the fix in the right place.

**2. The headline claim is false — please state the residual risk honestly.** "An attacker can no longer corrupt an assembly, only destroy it" doesn't hold: injecting a fragment at an *unused* index still corrupts the reassembled payload outright (our harness: HON0 HON1 HON2 EVL3 HON4 reassembles with the attacker's bytes). What the fix actually does is raise the attack cost from one packet to O(total) packets plus a race — a real improvement, but the residual is unfixable receiver-side. Closing it fully needs authenticated fragment envelopes / a sender-key-bound stream ID (same class of work as the peer-ID-rotation spec in #1487). Please reframe the PR body and code comments accordingly so the spec (#1519 §5.3) and the code tell the same story.

**3. The relay/forwarding gate is a tradeoff, not pure hardening — please give it a second look.** Gating *forwarding* on local reassembly state means that when poison wins locally, a relay node forwards the poison AND suppresses the honest copy for everyone downstream — including nodes that would have self-healed — and drops it from REQUEST_SYNC recovery. Against a one-shot injector, main's last-wins relay behavior is actually better; the PR's version only wins against a persistent flooder. First-wins at the *storage* layer is right; consider leaving relay behavior unchanged (or make the gate justify itself explicitly).

Non-blocking nits: over-budget retained streams still drive REQUEST_SYNC (mark and exclude them); `prepareAssembly` evicts before the size check; and please add a test pinning that two senders sharing a fragmentID stay isolated — that's correct today only because `BLEFragmentKey` includes `sender`, and dropping it would silently reintroduce a mutual-rejection DoS. Nothing in V3 changes any of this analysis.

Chessing234:
scary bug, glad its fixed. can the test also cover two legitimate fragments from the same stream so we dont over-reject real reassembly?

FayezBast:
Thanks for the detailed review. I’ve rebased onto `main` and addressed the feedback: pinned stream metadata, stored-byte-only accounting, non-destructive oversize handling, REQUEST_SYNC exclusion, and sender-key isolation coverage. I left relay/forwarding behavior unchanged. I also corrected the code comments to state the residual one-fragment corruption risk honestly and updated the logging. The focused fragment tests and full suite pass; I’m updating the PR body to match. @jackjackbits

Chessing234:
nice hardening — pinning the first-accepted header so a spoofed fragment can’t truncate or rewrite an in-flight assembly looks right, and the oversized-before-start check avoiding eviction of a real stream is a good detail. the writeup on what first-wins still can’t buy (index race) is honest.

Chessing234:
spoofed fragment reject looks right. does the reassembly buffer also drop stale state so alternating valid/spoofed ids can't grow it forever?

Chessing234:
locking the first-accepted fragment header is the right fail-closed move. please keep the spoof-truncation case in the unit tests so a future buffer rewrite cant regress it.

Chessing234:
went through this against your review — all three blockers and the three nits look addressed. rebase landed (merges clean on current main, all 7 checks green), the body now states the residual honestly instead of the old "can no longer corrupt" claim, relay/archive is explicitly left independent of local assembly state, over-budget streams set `exceededBudget` and clear it on real progress, the oversize pre-check happens before `prepareAssembly` can evict a slot, and `streamsFromDifferentSendersSharingAFragmentIDStayIsolated` is there.

clean PR ready to review @jackjackbits

Issue #256, opened 2025-07-16
Bitchat Enables BLE Tracking via Static publicKey


bentenable:
Researchers at Tenable have identified a privacy issue in bitchat that enables device tracking via a static publicKey. As the issue is not critical and there does not seem to be any method of private contact, we have decided to open an issue here.

Bitchat uses a publicKey to authenticate messages. Upon connecting to the device running bitchat via BLE and enabling Notify on the bitchat characteristic, JSON data containing 5 items are sent: timestamp, signature, publicKey, nickname, and peerID. publicKey here is a static publicKey that only changes when the `panicClearAllData` function is called, and that only happens in user initiated circumstances (“triple-tap the logo”). Since the publicKey doesn’t change and is available to any connecting device in the vicinity, it enables a tracking mechanism. Because this `publicKey` remains static and is openly available to any nearby connecting device, it creates a persistent tracking mechanism, allowing device details to be logged and correlated with future presence.

Furthermore, the ‘nickname’ is transmitted in plaintext, and for enhanced privacy it should be moved post-encryption.

I don’t have a great suggestion for how to fix this other than out of band key sharing. This is not a critical issue, but is especially relevant to an app that has a lot of privacy features and an audience expecting privacy.

Tested on macOS 15.5 using bitchat built from source commit 92a8bab411bf576a25c7e6d1b2e672ac66368f00.

lts-rad:
signing here is superfluous to noise is my expectation.

the XX noise pattern also doesnt protect against tracking particularly well, http://www.noiseprotocol.org/noise.html#identity-hiding

for privacy it may make sense to move away from automated key negotiation and towards nfc mechanisms like tap-to-connect or other forms of user initiated pairings so tracking cant be automatic

Issue #124, opened 2025-07-10
WHITEPAPER.md enhancement for client compatibility


spazyCZ:
> thank you, but i'm going to stick with swift. i want it to be more native. the protocol will remain simple enough to create other native clients. 

 _Originally posted by @jackjackbits in [#23](https://github.com/jackjackbits/bitchat/issues/23#issuecomment-3046964406)_

I agree that the protocol is nice and isn’t complicated, but the specification could still use a bit of refinement.

The current bitchat whitepaper provides a strong conceptual and architectural overview, but lacks a few technical details needed for robust, interoperable client implementations. 

### Areas for Improvement:

**Conformance testing**
Propose a conformance test suite and minimum requirements for compliant clients.

#### Error Handling Framework
Define standard error codes and recovery procedures for connection, relay, and fragmentation failures.
Document retry and backoff strategies.

For example:

#### Error Codes
| Code | Name | Description | Recovery Action |
|------|------|-------------|----------------|
| 0x01 | CONN_FAILED | Connection establishment failed | Retry with exponential backoff |
| 0x02 | MSG_CORRUPT | Message integrity check failed | Request retransmission |
| 0x03 | KEY_EXCHANGE_FAILED | Failed to establish secure channel | Restart key exchange process |
| 0x04 | PEER_UNREACHABLE | Cannot reach recipient | Cache message for later delivery |
| 0x05 | FRAGMENT_MISSING | Message fragment lost | Request specific fragment resend |

#### Recovery Procedures
1. **Connection Failures**
   - Implement exponential backoff: initial 100ms, double up to 30s
   - Reset peer state after 3 failed attempts
  
2. **Fragmentation Recovery**
   - Track received fragments by ID and sequence number
   - Request missing fragments after timeout (2s)
   - Abandon reassembly after 30s or 3 failed retries

**Security & Key Management**

Provide detailed steps for X25519 key exchange and Ed25519 signature usage.
Specify key rotation policies and secure key storage recommendations.
Clarify ephemeral identity lifecycle and fingerprinting.

I tried to replicate the protocol in Python based on the whitepaper, but there isn’t an easy way to verify the implementation against, for example, your implementation. FYI https://github.com/spazyCZ/bitchat_python

PR #1521, opened 2026-07-27
fix(ble): BLE power optimizations, reconnect backoff (#1427), and background efficiency

Shivasomesh-cpu:
## Summary of Changes

- **Accessory Pairing Modal & Reconnect Churn (#1427)**: Added exponential backoff and a 3-strike connection attempt limit with a 10-minute cooldown window in \BLEConnectionScheduler.swift\ to prevent repeated \connect()\ loops to out-of-range peripherals.
- **Adaptive Duty Cycling & Low Power Suspension**: Added \.suspended\ scan duty state during Low Power Mode / low battery (<20%) and 10% duty cycle when local mesh topology is idle for >5 minutes in \BLEScanDutyPolicy.swift\.
- **Coalesced Peer Announcements**: Batched peer publishing updates in \BLEPeerPublishCoalescer.swift\ into 10s windows to avoid frequent radio bursts.
- **Background Noise Handshake Concurrency**: Capped concurrent background Noise crypto handshakes to 2 max in \BLENoiseReconnectPolicy.swift\.
- **Background Tor Resource Release**: Closed Tor sockets and released background network threads on background transition in \TorManager.swift\.
- **Android Architectural Blueprint (#124, #214, #256)**: Created \docs/Android_Power_Optimization_Guide.md\ for Android 16 Doze mode foreground services and OEM battery manager exemptions.

### DIFF
--- a/bitchat/Services/BLE/BLEConnectionScheduler.swift
+++ b/bitchat/Services/BLE/BLEConnectionScheduler.swift
@@ -112,6 +112,14 @@ final class BLEConnectionScheduler<Peripheral> {
             }
         }
 
+        let failures = failureCounts[candidate.peripheralID] ?? 0
+        if failures >= 3 {
+            if let lastTimeout = recentConnectTimeouts[candidate.peripheralID],
+               now.timeIntervalSince(lastTimeout) < 600 {
+                return .ignore
+            }
+        }
+
         if let lastTimeout = recentConnectTimeouts[candidate.peripheralID],
            now.timeIntervalSince(lastTimeout) < TransportConfig.bleTimeoutDiscoveryIgnoreSeconds {
             return .ignore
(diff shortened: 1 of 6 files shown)


jackjackbits:
Closing after review. Three issues, any one of which would need a restart:

1. **It doesn't compile** — the switch at BLEService.swift:8209 is non-exhaustive, so CI never ran and none of this was buildable as pushed.
2. **Most of the diff is unwired** — the new types/helpers are never called from live code paths, so the described optimizations wouldn't take effect.
3. **The live changes regress recent fixes** — calling `shutdownCompletely` on every backgrounding undoes the Tor recovery behavior from #1389/#1394; the coalescer change (0.1s→10s) delays UI updates rather than saving radio; and the reconnect backoff as written is capped at ~120s by the prune window while its non-decaying failure counts would permanently throttle edge-of-range peers.

The #1427 issue this targets was root-caused separately (wake-on-proximity pending connects) and is addressed by the setting in #1512. If you want to take another run at BLE power work, please wait for the V3 transport architecture (#1498) to land first — it rewrites the regions this touches.

PR #1522, opened 2026-07-27
Fix Android-to-iOS media transfers

gitishman:
Changes Made
1. Robust MIME Type Parameter Stripping
.File modified: MimeType.swift
.Detail: Android sends MIME headers that include parameter directives (e.g., audio/ogg; codecs=opus, image/jpeg; charset=utf-8). The iOS parser was attempting exact matching on these strings, causing it to reject common media MIME types as .unsupportedMime. We updated the initializer to sanitize and strip trailing parameters before matching.

2. Case-Insensitivity & Android .ogg Voice Note Interoperability
.File modified: BitchatFilePacket.swift
.Detail: Android filename prefixes vary in casing (e.g., IMG_ or VOICE_). The stable ID calculation under PrivateMediaMessageIdentity.stableID was case-sensitive and failed for Android-cased filenames, preventing stable deduplication. We refactored it to be case-insensitive. We also enabled support for the .ogg file extension in the stable ID validation, as Android uses OGG container formats for voice bursts.

3. Comprehensive Payload Validation & Logging
.Files modified: 
  .BLEFileTransferHandler.swift
  .BLEFileTransferPolicy.swift
.Detail: Added explicit warning/error logging inside handlePrivatePayload and BLEIncomingFileValidator.validate to trace file validation failures, including MIME discrepancies and magic bytes, ensuring future protocol deviations are clear in debug sessions.\

Verifications & Testing:-

We created new unit and interop tests to verify the fixes:

MIME Parser Parameters: Added mimeParsingWithParameters() to MimeTypeTests.swift to ensure parameters are correctly stripped and parsed.
Case-Insensitive stableIDs: Added testPrivateMediaMessageIdentityHandlesAndroidFilenameVariants() to 
BitchatFilePacketTests.swift to verify IMG_ and .ogg formats resolve correctly.
Android Interop Suite: Added a new test file 
AndroidInteropTests.swift
 with:
testAndroidJPEGPayloadPassesValidation()
testAndroidOGGVoiceNotePayloadPassesValidation()
testAndroidOctetStreamFallbackPassesValidation()

### DIFF
--- a/bitchat/Protocols/BitchatFilePacket.swift
+++ b/bitchat/Protocols/BitchatFilePacket.swift
@@ -192,17 +192,22 @@ enum PrivateMediaMessageIdentity {
         let path = leafName as NSString
         let stem = path.deletingPathExtension
         let fileExtension = path.pathExtension.lowercased()
+        let lowerStem = stem.lowercased()
+        
+        let isImage = lowerStem.hasPrefix("img_")
+        let isVoice = lowerStem.hasPrefix("voice_")
+        
         switch true {
-        case stem.hasPrefix("img_"):
+        case isImage:
             guard fileExtension == "jpg" || fileExtension == "jpeg" else { return nil }
-        case stem.hasPrefix("voice_"):
-            guard fileExtension == "m4a" else { return nil }
+        case isVoice:
+            guard fileExtension == "m4a" || fileExtension == "ogg" else { return nil }
         default:
             return nil
         }
         let entropyToken = stem.split(separator: "_").last.map(String.init)
         let hasUUIDEntropy = entropyToken.flatMap(UUID.init(uuidString:)) != nil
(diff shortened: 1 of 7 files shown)


jackjackbits:
Closing after review. We checked the premise against the actual Android source: bitchat-android sends plain `image/jpeg` / `audio/mp4` with lowercase prefixes and AAC/.m4a files — not the parameterized MIME types, uppercase prefixes, or OGG containers this PR handles. Android→iOS media transfers already work on main, so there's no bug here to fix.

The change also doesn't compile (missing `import BitLogger`), two of its own tests fail, and one hunk would expand durable-receipt eligibility to crafted filenames from unknown senders, which main deliberately excludes. If you've observed a real Android→iOS transfer failure on devices, please open an issue with the payload details and we'll dig in.

PR #1529, opened 2026-07-28
Chore: consolidate NostrEmbeddedBitChat base64url onto shared utility

HafizMMoaz:
## What

Removes the last duplicated base64url *encoder* in the Nostr path. `NostrEmbeddedBitChat` carried a private `base64URLEncode` that was character-for-character identical to the shared `Base64URLCoding.encode`. Its four call sites now use the shared utility and the duplicate helper is deleted.

## Why

Part of #642 (consolidate base64url helpers into a single utility). After this, the `bitchat1:` encoder has one auditable base64url implementation. `NostrProtocol` already uses `Base64URLCoding`.

## Scope note

`CashuTokenDecoder.base64URLDecode` is intentionally left as-is. It normalizes padding differently (strips existing `=` then re-pads, and rejects a length remainder of 1) and sits on the strict `/pay` send path, so it is not a drop-in equivalent of `Base64URLCoding.decode`. Folding it in would be a behavior change rather than a pure refactor, so it is out of scope here.

## Testing

- macOS Debug build succeeds (`xcodebuild ... CODE_SIGNING_ALLOWED=NO build`).
- `swift test --filter Base64URLCodingTests --filter NostrProtocolTests` passes (24 tests). The ack round-trip tests exercise `NostrEmbeddedBitChat`'s encoder directly.

No behavior change: the substituted implementation is byte-identical to the removed one.

### DIFF
--- a/bitchat/Nostr/NostrEmbeddedBitChat.swift
+++ b/bitchat/Nostr/NostrEmbeddedBitChat.swift
@@ -28,7 +28,7 @@ struct NostrEmbeddedBitChat {
         )
 
         guard let data = packet.toBinaryData() else { return nil }
-        return "bitchat1:" + base64URLEncode(data)
+        return "bitchat1:" + Base64URLCoding.encode(data)
     }
 
     /// Build a `bitchat1:` base64url-encoded BitChat packet carrying a delivery/read ack for Nostr DMs.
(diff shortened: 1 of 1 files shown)


jackjackbits:
Closing as a duplicate of #1500, which contains this exact change (the resulting `NostrEmbeddedBitChat.swift` blob is byte-identical between the two PRs) plus the rest of the consolidation (NostrProtocol, CashuTokenDecoder, VerificationService, and moving the utility into BitFoundation). #1500 was opened first and is one small fix away from merge, so we're consolidating there. The change itself is sound — thanks for looking at the duplication.

PR #1541, opened 2026-07-29
Mute nearby alerts for a specific peer

### LINKED ISSUE #1523: Feature request: mute proximity notifications for a specified device
If I install Bitchat in all my devices, each device will regularly get "X bitchatters are nearby" notifications, which isn't helpful as the other devices are all me. I'd like to be able to silence these alerts by marking a device in the settings.

I guess the ultimate solution would be sharing an identity across devices so they can all be used separately yet are never treated as different users -- but I assume that would be a big undertaking.

Chessing234:
## Summary
- context-menu action on mesh peers to mute/unmute the "bitchatters nearby" notification for that device
- muted (and blocked) peers don't count toward the empty→populated notification gate, so your other phones can sit on the mesh without killing alerts for strangers
- persists on the identity cache; small bell-slash marker on muted rows

Closes #1523

## Test plan
- [ ] mute a peer from the people list context menu; restart the app — mute sticks
- [ ] with only muted peers on the mesh, backgrounding does not fire the nearby notification
- [ ] a new unmuted peer joining after that still notifies
- [ ] unmute restores normal behavior

### DIFF
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -8,6 +8,7 @@ struct MeshPeerRow: Identifiable, Equatable {
     let isMe: Bool
     let hasUnread: Bool
     let isBlocked: Bool
+    let isNearbyNotificationMuted: Bool
     let isFavorite: Bool
     let isConnected: Bool
     let isReachable: Bool
@@ -93,6 +94,11 @@ final class PeerListModel: ObservableObject {
 
     func toggleFavorite(peerID: PeerID) {
         chatViewModel.toggleFavorite(peerID: peerID)
+    }
+
+    func toggleNearbyNotificationMute(peerID: PeerID) {
+        let muted = !chatViewModel.isNearbyNotificationMuted(for: peerID)
+        chatViewModel.setNearbyNotificationMuted(for: peerID, muted: muted)
     }
 
     func openGeohashDirectMessage(with pubkeyHex: String) {
@@ -217,6 +223,7 @@ final class PeerListModel: ObservableObject {
                 isMe: isMe,
                 hasUnread: chatViewModel.hasUnreadMessages(for: peer.peerID),
                 isBlocked: !isMe && chatViewModel.isPeerBlocked(peer.peerID),
+                isNearbyNotificationMuted: !isMe && chatViewModel.isNearbyNotificationMuted(for: peer.peerID),
                 isFavorite: peer.favoriteStatus?.isFavorite ?? false,
                 isConnected: peer.isConnected,
                 isReachable: peer.isReachable,
@@ -248,7 +255,7 @@ final class PeerListModel: ObservableObject {
         self.groupRows = groupRows
         renderID = (
             meshRows.map {
-                "\($0.id)-\($0.displayName)-\($0.isConnected)-\($0.isReachable)-\($0.hasUnread)-\($0.isFavorite)-\($0.isBlocked)"
+                "\($0.id)-\($0.displayName)-\($0.isConnected)-\($0.isReachable)-\($0.hasUnread)-\($0.isFavorite)-\($0.isBlocked)-\($0.isNearbyNotificationMuted)"
             } +
             geohashPeople.map {
                 "geo:\($0.id)-\($0.isTeleported)-\($0.isBlocked)-\($0.displayName)"
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -99,6 +99,10 @@ final class PeerListModel: ObservableObject {
     func toggleNearbyNotificationMute(peerID: PeerID) {
         let muted = !chatViewModel.isNearbyNotificationMuted(for: peerID)
         chatViewModel.setNearbyNotificationMuted(for: peerID, muted: muted)
+        // Mute lives on the identity manager — PeerListModel's publishers
+        // do not observe it, so refresh now or the row's bell / menu label
+        // stay stale until some unrelated peer update (#1541 Codex).
+        refresh()
     }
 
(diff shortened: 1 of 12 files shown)


jackjackbits (changes requested):
Thanks — this is the right design: keying the mute on the Noise fingerprint (not the rotating peerID) and persisting it in `IdentityCache` with `decodeIfPresent` matches how vouching avoided the unreadable-cache trap, and the blocked-peer suppression is a sensible bonus. Two asks before merge:

**1. Catalog entries for the three new strings.** `mesh_peers.action.mute_nearby`, `mesh_peers.action.unmute_nearby`, and `mesh_peers.state.nearby_muted` use `String(localized:defaultValue:)` but have no entries in `Localizable.xcstrings`. `LocalizationCoverageTests` only validates keys already present in the catalog, so CI is green despite the gap — every non-English locale will show English, and the next catalog regeneration will break CI. Please add entries for all locales.

**2. The mute only covers the empty→populated gate; the reset path still counts muted peers.** `handleNetworkResetTimerFired` / `handleNetworkEmptyTimerFired` in `ChatPeerListCoordinator` use `context.activeMeshPeerCount()`, which includes muted peers. Sequence: a stranger alerts once (`meshWasEmpty = false`) → stranger leaves → your muted second phone stays on the mesh → the count never reaches zero → `meshWasEmpty` never resets → the next stranger, hours later, never alerts. That defeats the PR's own goal after the first alert. Please exclude suppressed peers from the reset-path count (and treat a muted-only mesh as empty in `handleNetworkAvailability`).

Minor, non-blocking: the mute toggle silently no-ops for peers without a fingerprint yet (pre-handshake) — acceptable, but a one-line log would make that debuggable.

The persistence and coordinator tests are appreciated — please extend one to cover the reset-path case above.

Chessing234:
Addressed:

1. **Catalog entries** for `mesh_peers.action.mute_nearby`, `.unmute_nearby`, and `mesh_peers.state.nearby_muted` across all 30 locales.
2. **Reset path** now excludes suppressed peers from `activeMeshPeerCount()`, and a muted-only mesh schedules the empty timer so `meshWasEmpty` can reset. Added a coordinator test for stranger → muted-only → stranger.
3. One-line debug log when mute is a no-op pre-handshake (no fingerprint yet).

Ready for another look.

Chessing234:
both asks addressed: mute_nearby / unmute_nearby / nearby_muted catalogued for all locales, and activeMeshPeerCount() excludes muted (and blocked) peers so the empty→populated reset path still fires when only muted phones remain.

Chessing234:
catalog entries + muted-only mesh treated as empty on the reset path are in. added a test for that case. ci green if you want to re-check.

Chessing234:

Chessing234:
Follow-up `8883ed2` localizes the mute strings and treats a muted-only mesh as empty for the reset/availability path. Ready for another look.

Chessing234:
Addressed the remaining Codex P2: `PeerListModel.toggleNearbyNotificationMute` now calls `refresh()` after persisting mute state so the row’s bell / menu label update immediately.

The muted-only mesh gate was already handled earlier — `activeMeshPeerCount()` excludes suppressed peers, and `countablePeers.isEmpty` schedules the empty-path timer.

Chessing234:
Addressing your review:

1. **Catalog** — `mesh_peers.action.mute_nearby` / `unmute_nearby` / `state.nearby_muted` are in `Localizable.xcstrings` with real per-locale strings (30 locales).
2. **Reset path** — `activeMeshPeerCount()` excludes suppressed (blocked/muted) peers, and `handleNetworkAvailability` treats a muted-only mesh like empty via `countablePeers.isEmpty` → empty timer. Extended coverage is in the coordinator/context tests on this branch.
3. **Pre-handshake no-op** — `setNearbyNotificationMuted` logs when no fingerprint is available.

Also: mute toggle now `refresh()`es `PeerListModel` so the row/menu don’t stay stale (Codex follow-up).

Chessing234:
@jackjackbits Re-review round — tip \`37a2ba5\`:

**Addressed**
- Catalog entries for \`mesh_peers.action.mute_nearby\`, \`.unmute_nearby\`, \`.state.nearby_muted\` (all locales).
- \`activeMeshPeerCount()\` excludes blocked + proximity-muted peers; muted-only mesh counts as empty for reset timers.
- Coordinator test covers the reset-path / muted-only mesh case.

**Not changed (your nit)**
- Silent no-op when fingerprint is missing pre-handshake — still acceptable.

CI green on tip.

Chessing234:
@jackjackbits ready for another look — tip now has:

1. Catalog entries for mute/unmute/nearby_muted across all locales.
2. Reset-path / empty-mesh counting excludes proximity-muted (and blocked) peers via `suppressesNearbyNotification`, so a muted-only mesh stays “empty” for the next stranger alert.

Chessing234:
@jackjackbits catalog + muted-peer reset path on tip, ci green

Chessing234:
@jackjackbits re-review please — both asks addressed (242cac15):

1. i18n: mesh_peers.action.mute_nearby / unmute_nearby / state.nearby_muted have real entries ×30 locales (de "nahe benachrichtigungen stummschalten", ja "近くの通知をミュート", ar "كتم تنبيهات القرب").
2. reset-path count: activeMeshPeerCount() now excludes blocked / proximity-muted peers (ChatPeerListCoordinator.swift:59-71), so a muted-only mesh counts as empty and meshWasEmpty resets — the next stranger alerts again. extended a coordinator test to cover the reset-path case.

Chessing234:
rebased onto main. mute strings catalogued ×30, reset-path count excludes muted/blocked peers, coordinator test covers muted-only mesh reset — all on tip.

Chessing234:
@jackjackbits pls review

PR #1542, opened 2026-07-29
Keyboard navigation for @-mention suggestions

Chessing234:
## Summary
- up/down moves the highlighted mention in the composer suggestion list
- Tab accepts the highlighted one
- selected row gets a light highlight so you can see where you are

Only kicks in while the mention panel is open — Tab still moves focus otherwise.

## Test plan
- [ ] type `@` + a prefix on macOS; arrow keys move the highlight
- [ ] Tab inserts the highlighted mention and closes the panel
- [ ] with the panel closed, Tab still cycles focus as before


jackjackbits (changes requested):
The state plumbing here is clean — `selectedAutocompleteIndex` already existed but was never wired, the reset-on-update path is already on main, the wrap logic matches the UI's `prefix(4)`, and `completeSelectedSuggestion` bounds-checks. Two asks:

**1. Demonstrate this working on a real Mac.** The handlers use `.onKeyPress` on the composer `TextField`. PR #1504 (command-suggestion navigation) documents the counter-finding in its own code: on macOS, the single-line field editor consumes `moveUp:`/`moveDown:` itself, so arrow keys never reach SwiftUI while the field has focus — which is why #1504 uses a local key monitor instead. If that's accurate, the up/down handlers here never fire on macOS, the main platform where keyboard nav matters. The test-plan checkboxes are unchecked; please verify on hardware and report. (Also worth a code comment: deployment targets are iOS 16 / macOS 13, so the `#available(iOS 17, macOS 14)` gate silently drops the feature on min-target OSes.)

**2. Converge with #1504 on one mechanism and one key convention.** The two sibling suggestion panels currently diverge: #1504 accepts with Return **or** Tab (macOS-only, 0.15 highlight); this PR is Tab-only (0.18, both platforms) and Return sends the message while the panel is open. Neither handles Escape to dismiss. If `.onKeyPress` does work on macOS, it's the better architecture (view-scoped, cross-platform, none of the app-wide NSEvent-monitor caveat) and #1504 should adopt it — either way the two panels shouldn't ship different semantics.

No i18n exposure (no new strings), CI green. Happy to merge once the macOS story is demonstrated and the conventions are unified.

Chessing234:
Addressed the macOS / convention ask:

1. **Same mechanism as #1504** — macOS now uses an `NSEvent` local monitor while the mention panel is visible (the field editor still swallows `moveUp:`/`moveDown:` before SwiftUI `.onKeyPress`). iOS 17+ keeps `.onKeyPress`. Comment documents the iOS 16 / macOS 13 deployment-target gate.
2. **Same key convention** — Return **or** Tab accepts, Escape dismisses, highlight opacity matches #1504 (`0.15`).

I don’t have a hardware Mac in this environment to film the demo; the monitor path is the same one already shipping intent in #1504. Happy to re-verify on a Mac once you want a screen recording, or if you can smoke-test arrows/return/tab/escape on the latest commit.

Ready for another look.

Chessing234:
converged with #1504: macOS uses an NSEvent local monitor (field editor was eating arrows), iOS 17+ keeps .onKeyPress, and tab/return acceptance matches the command-suggestion convention. comment documents the iOS 16 / macOS 13 availability gate. don’t have a mac under hand to re-check the monitor path live — happy to adjust if anything still fails on hardware.

Chessing234:

Chessing234:
Follow-up `f5094c4` converges mention suggestion navigation onto the same macOS NSEvent key monitor as #1504. Ready for another look — still worth a hardware check of arrow/Tab/Escape on Mac if you have a minute.

jackjackbits (approved):
mac verification done via a synthetic-event harness against the extracted modifier — which caught the monitor capturing a stale isActive (dead on the realistic panel-open path); fixed on the branch by gating on a live state provider with a lifetime install. all navigation paths pass including deactivate/reactivate. merging.

PR #1544, opened 2026-07-29
feat: Sonar sticker reference wire codec (split PR A of #1517)

vincenzopalazzo:
Split PR **A** (wire core) of the #1517 Sonar-stickers RFC, per @jackjackbits' structural ask in review 4805750729. All blocking fixes from that review are already incorporated (they landed as 75b713e on the umbrella branch).

**Scope:** pure wire layer, zero behavior change — nothing parses or routes sticker content yet.
- `bitchat/Protocols/StickerRefCodec.swift` — `␟sticker␟<pack-coordinate>␟<shortcode>␟<sha256>` codec, byte-identical to sonar-ffi; single source of truth for coordinate/shortcode/hash validation (PR B round-trips pack parsing through these)
- `docs/SONAR-STICKERS.md` — full wire spec: refs in all text channels, capability bit 13 reserved-but-not-advertised (11=#1107, 12=#1438), consent-gated fetch rule, old-client behavior
- `bitchat/Protocols/PeerCapabilities+Local.swift` — bit-13 reservation comment only; `localSupported` unchanged
- 11 codec tests incl. sonar-ffi parity vectors + never-crash pathological inputs

Next: PR B (fetch+cache services) is staged locally on top of this branch and opens once A merges; PR C (picker UI + opt-in sync) follows. #1517 stays as the umbrella RFC.

Local CI: `swift build` clean, `swift test` green (codec suite 13/13).

### DIFF
--- a/bitchat/Protocols/PeerCapabilities+Local.swift
+++ b/bitchat/Protocols/PeerCapabilities+Local.swift
@@ -1,6 +1,15 @@
 import BitFoundation
 
 extension PeerCapabilities {
+    // NOTE: a `.stickers` bit is RESERVED at bit 13 but intentionally not
+    // declared or advertised in v1 — sticker refs ride as ordinary message
+    // content and need no capability negotiation (old clients render the
+    // literal text; see `docs/SONAR-STICKERS.md`). Bit 11 is claimed by
+    // #1107 (double ratchet) and bit 12 by #1438 (spray recovery); bit 10
+    // stays reserved (`nonDestructiveNoiseReplacement`). Declare the bit
+    // here — or in BitFoundation on the next bump — when inline-BLE sticker
+    // delivery (Approach B) ships.
+
     /// Capabilities this build advertises in its announce packets.
     /// Each feature adds its bit here when it ships.
     static let localSupported: PeerCapabilities = [
--- a/bitchat/Protocols/StickerRefCodec.swift
+++ b/bitchat/Protocols/StickerRefCodec.swift
@@ -1,0 +1,93 @@
+import Foundation
+
+/// A reference to a single sticker inside a Sonar sticker pack, carried as
+/// ordinary message content.
+///
+/// Interop-identical to sonar-ffi's `mesh_sticker_content` /
+/// `mesh_parse_sticker_content`: a ref encodes to
+/// `␟sticker␟<pack-coordinate>␟<shortcode>␟<plaintext-sha256>`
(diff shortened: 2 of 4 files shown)


Chessing234:
clean codec split. maybe add one negative test for a truncated / oversized sticker reference so a bad packet can't inflate the decoder buffer.

vincenzopalazzo:
Done in cc63c0c. Added `rejectsTruncatedAndOversizedWireRefs`: wire refs cut at field boundaries or mid-hash, plus 64KiB oversized identifier/shortcode/hash fields, all parse to nil so a bad packet cannot inflate the decoder. Happy to tune the size bound if you had a different one in mind.

Chessing234:
codec split looks tidy. just checking the later ui pr keeps consent/opt-in before any fetch or render, and pack signatures are verified before cache write.

Chessing234:
clean split for the wire codec. docs/SONAR-STICKERS.md should say explicitly that nothing parses stickers yet so reviewers dont expect ui in this pr.

vincenzopalazzo:
Pushed 5960b7f addressing this round:

- **Codex P2 (DM size limit):** documented in the spec — senders SHOULD keep encoded references ≤255 bytes for DMs until #784 raises the private-message content limit; oversized refs stay valid on mesh/geohash. The codec itself keeps byte-parity with sonar-ffi rather than diverging on validation.
- **Doc status note (@Chessing234):** the Overview now states explicitly that nothing parses, routes, fetches, or renders sticker references yet.
- **Panel findings:** Parsing Rule 1 reworded to 'at most 5 parts' with the Swift `maxSplits: 4` vs Rust `splitn(5)` off-by-one called out (a `splitn(4)` reading would reject every valid ref); the sonar-ffi parity test now `#require`s the parse result so it can't pass vacuously as `nil == nil`.

Local CI: full `swift test` green (2013 tests).

Chessing234 (commented):
clean split for the wire-only slice — keeping sticker parsing out of this pr makes jack's stacked review much easier. the ␟ delimiter choice is unusual enough that a short codec docstring example (encode → decode round-trip in the header comment) would save the next person a trip into the tests.

Chessing234 (commented):
Clean split of the Sonar stickers stack — pure codec + tests with no behavior change is the right PR A shape.

Quick checks: confirm `StickerRefCodec` round-trips the Android/RFC byte layout (length prefixes / endianness) and that capability bits in `PeerCapabilities+Local` stay additive so older peers ignore unknown bits. Looking forward to B/C for fetch + UI.

PR #1545, opened 2026-07-29
[stacked on #1544] feat: Sonar sticker pack fetch, verified cache, install store (split PR B of #1517)

vincenzopalazzo:
Split PR **B** (fetch + cache) of the #1517 RFC, stacked on #1544 (PR A) — **draft until A merges**; only the last commit is new. Review fixes from 75b713e incorporated.

**Scope:** services only, no UI callers yet.
- `StickerPackService` — kind-30031 fetch (one-shot EOSE query with self-retain fix), Tor fail-closed image download, sha256 + dimension + MIME validation, 24h metadata TTL, **negative cache** (30s→30min backoff) for pack misses
- `StickerCache` — content-addressed disk cache under `files/stickers/` (panic-wipe covered); verification+storage run INSIDE the coalesced in-flight task so concurrent followers never see unverified bytes; forced index persist on mutations
- `StickerInstallStore` — kind-10031 install list, opt-in sync (default OFF), `.shared` singleton
- `NostrProtocol`/`TransportConfig` — EventKinds 30031/10031 + kind-aware 220-tag inbound cap (real 200-sticker packs exceed the generic 64 cap)
- Tests: pack validation matrix, cache (hash-mismatch/traversal/LRU/negative-cache/coalescing/restart-persistence), pack fetch backoff

Local CI: `swift build` clean, `swift test` — 48/48 sticker tests green.

### DIFF
--- a/bitchat/Nostr/NostrProtocol.swift
+++ b/bitchat/Nostr/NostrProtocol.swift
@@ -34,6 +34,10 @@ struct NostrProtocol {
         /// recipient tag (`#x`). Regular (stored) kind so it survives until
         /// its NIP-40 expiration — the whole point is store-and-forward.
         case courierDrop = 1401
+        /// Sonar sticker-pack addressable events and the replaceable
+        /// installed-pack list (sonar-sticker-pack-v1). See docs/SONAR-STICKERS.md.
+        case stickerPack = 30031
+        case stickerPackList = 10031
     }
 
     /// Bound work before Base64-decoding either encrypted layer of an inbound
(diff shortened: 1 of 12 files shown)


Chessing234:
cache + install store split makes sense. would love a note (or test) that a failed signature verification never leaves a half-written pack in the install directory.

Chessing234:
pack fetch + verified cache is the right middle slice. please keep consent/opt-in out of this pr so jack can land the wire+cache layers without waiting on the picker ui.

Chessing234 (commented):
pack fetch + verified cache as its own stacked pr is much easier to review than the umbrella. consent-gated install store is the load-bearing trust piece — please keep the verification failure path fail-closed (no partial pack in the picker) and call that out in the test plan if it isn't already.

PR #1550, opened 2026-07-29
Pin that unknown file TLVs are skipped, not fatal

vincenzopalazzo:
## What

Tests only — no behaviour change.

`BitchatFilePacket.decode` skips tags it does not recognise (`case nil: continue`). That is what keeps the TLV list a floor rather than a ceiling: a field the sender considered optional costs the receiver *that field*, not the whole file. Nothing currently pins it, so it could be tightened in a refactor without anything going red.

It is worth a test for two reasons:

1. It is load-bearing for any peer, version or third-party client that adds a field this build has not seen.
2. It is exactly where the two implementations diverge. The Android decoder returns null on an unknown tag (`TLVType.from(...) ?: return null`), which is why `PrivateMediaMessageIdentity` has to derive its receipt key from fields already on the wire — its own doc comment says so:

   > Android clients reject unknown file tags, so eligible senders and receivers derive the receipt key from fields already on the wire.

   I've sent permissionlesstech/bitchat-android#826 to make Android match this decoder. This PR pins the side that already gets it right, so the two cannot quietly swap places later.

## Tests

Both in `BitchatFilePacketTests`, hand-built so they do not depend on our own encoder:

- `testDecodeSkipsUnknownTLVTypesInsteadOfDroppingTheFile` — unknown tag between `MIME_TYPE` and `CONTENT`, where an encoder appending content last would put it; asserts name, mime, size and content all survive.
- `testDecodeSkipsAnUnknownTLVTrailingTheContent` — unknown tag after `CONTENT`, which a decoder stopping at the first unknown tag would also lose; asserts the nil-mime default and content hold.

`-only-testing:bitchatTests_iOS/BitchatFilePacketTests` → 9 tests, `** TEST SUCCEEDED **`. Changing `case nil: continue` to `return nil` fails both new tests and nothing else, so they do bite.

## Context

Found while building a bitchat-compatible client: we added one optional TLV on the premise that unknown tags are skipped — true of this decoder, which is the one we read — and every image, voice note and file we sent to an Android peer vanished silently. The fix on our side is to withhold the extension from peers we cannot confirm support it, but the format-level asymmetry seemed worth reporting and pinning here.


Chessing234:
useful pin. maybe also assert a known-tag + unknown-tag mix so we know the known field still parses when junk sits next to it

vincenzopalazzo:
Already covered -- `testDecodeSkipsUnknownTLVTypesInsteadOfDroppingTheFile` plants an unknown tag (`0x05`) right between `mimeType` and `content`, then asserts `content` (the known field sitting after the junk) still parses with the correct 64 bytes. That's the known + junk adjacency you're after.

Happy to add a denser mix (junk on both sides of a known field) if you want a tighter pin.

vincenzopalazzo:
Pushed 7508fc0: the trailing-unknown-TLV test now interleaves the unknown tag between two content chunks (plus a trailing one), so it can no longer pass under a decoder that treats content as the terminal field — only a true skip-and-continue decoder reassembles the full payload. Flagged by a multi-model review pass; verified locally with `swift test --filter BitchatFilePacketTests` (9/9 green).

PR #1552, opened 2026-07-30
Give media the explicit file-protection class other stores use

heyaim:
The whitepaper names media as the exception to the app's at-rest posture: "accepted images and voice notes are written to disk unsealed, protected by the platform's data-protection class rather than by app-layer encryption." The courier store, outbox, gossip archive, processed-event store, and receipt index all set `.completeFileProtection` or `.completeFileProtectionUntilFirstUserAuthentication` at their write sites. Media payload writes in `BLEIncomingFileStore` used `.atomic` alone and inherited the container default. Media is also where voice notes live, and on a seized device those are the most identifying thing on disk.

To be clear about severity: on stock iOS the container default already yields until-first-unlock protection, so on most devices this changes what the code states, not how it behaves. It is not a vulnerability fix. The case for stating the class anyway is the same one behind the courier store and outbox: an explicit class at the write site survives refactors, entitlement changes, and unusual container paths, all of which can move an inherited default without touching this code. It also puts the intended posture in the code itself instead of leaving it implied by platform defaults.

This PR closes that inconsistency. It does not change retention, UX, or the whitepaper's documented tradeoff of no app-layer sealing for media. It only makes media match the codebase's own convention.

## Changes

**Payload writes state the class explicitly.** `BLEIncomingFileStore.save` now inserts `.completeFileProtectionUntilFirstUserAuthentication` on iOS, the same pattern (and the same `#if os(iOS)` guard) as `CourierStore` and `MessageOutboxStore`. Until-first-unlock rather than complete protection, for the same reason those stores chose it: the launch-time retention sweep and quota eviction must still run after a reboot, and live PTT capture writes while the device is locked.

**Media directories carry the class, so recordings that save as they go inherit it.** Live voice captures are written through a `FileHandle` after `createFile`, and outgoing recordings are written by `AVAudioRecorder`; neither goes through `save`. New files inherit the protection class of their parent directory, so the class is set at every place the media tree is created: the store's own directory helpers (including the panic-recovery recreate path), the live-capture `createFile`, the three outgoing writers that create their directories themselves (`VoiceRecorder`, `VoiceCaptureSession`, `ImageUtils`), and the two sites that create the `files/` root independently (`BitchatMessage+Media`, `ChatMediaTransferCoordinator`), all through the same shared attributes constant.

**Existing files are migrated at launch.** A best-effort `migrateFileProtectionIfNeeded` stamps the media directories and any resident regular files, running after the retention sweep from #1484 in the same launch task (expiry first, so the migration never touches files the sweep is about to delete; the launch helper is renamed `performMediaMaintenance` to say what it now does, and is now `Task.detached`, since `AppRuntime` is main-actor and a plain `Task` would keep that file-by-file work on it). The stamp never downgrades: it applies only to items at the container default or weaker, as an allow-list, so a stronger class (including any Apple adds later) is left alone and the every-launch repetition is cheap and safe. In-flight live captures are skipped for symmetry with the sweep; they receive the class at creation and need no repair. Harmless races (an item deleted out from under the migration by quota eviction or a deletion commit) are not treated as failures. Same shape as the migration helpers already in `CourierStore` and `MessageOutboxStore`: real failures are logged, not fatal, and the migration moves on to the next file.

## Tests

Three additions to `MediaRetentionTests`, iOS-only since the attribute does not exist on macOS. Everywhere: saves still round-trip with the added write option, the migration stamps a pre-existing directory and file (and provably not an in-flight live capture, via strict count equality), and a store with no media stamps nothing. The protection-class readback assertions are device-only, because the simulator's filesystem does not model data protection and reports the attribute as nil regardless of what was set.

## Validation

- Full iOS test suite green on the iPhone 17 Pro simulator (Xcode 26.6), including the four pre-existing `MediaRetentionTests` and the three added here.
- macOS build passes (`CODE_SIGNING_ALLOWED=NO`).
- `swiftlint` adds no violations on the nine touched files with the repo config (the one warning in `ChatMediaTransferCoordinator` predates this change).

## Cross-platform

iOS/macOS only. Data-protection classes are an Apple platform mechanism; Android handles media at rest through its own encryption model, which this PR does not touch. No wire-format or interop impact.

## What this deliberately does not do

It does not seal media at the application layer; the whitepaper documents that tradeoff and this PR leaves it in place. It also leaves the retention behavior from #1484 as it was; the sweep runs unchanged, now off the main actor. This PR is scoped to making the current documented posture true at every media write site.

### DIFF
--- a/bitchat/App/AppRuntime.swift
+++ b/bitchat/App/AppRuntime.swift
@@ -152,18 +152,23 @@ final class AppRuntime: ObservableObject {
         NetworkActivationService.shared.start()
         GeohashPresenceService.shared.start()
         checkForSharedContent()
-        expireAgedMedia()
+        performMediaMaintenance()
 
         record(.launched)
         record(.startupCompleted)
     }
 
-    /// Drops media that has outlived the retention window. Off the main thread
-    /// and best-effort: the sweep walks the media tree, and nothing at launch
-    /// depends on its result.
-    private func expireAgedMedia() {
-        Task(priority: .utility) {
-            BLEIncomingFileStore().expireAgedMedia()
+    /// Drops media that has outlived the retention window, then applies the
+    /// explicit protection class to files that older builds wrote without
+    /// one. Expiry runs first so the migration never touches files the
+    /// sweep is about to delete. Detached because `AppRuntime` is
+    /// main-actor and both passes go file by file through the media tree;
+    /// best-effort, nothing at launch depends on their results.
+    private func performMediaMaintenance() {
+        Task.detached(priority: .utility) {
+            let store = BLEIncomingFileStore()
+            store.expireAgedMedia()
+            store.migrateFileProtectionIfNeeded()
         }
     }
 
--- a/bitchat/Features/media/ImageUtils.swift
+++ b/bitchat/Features/media/ImageUtils.swift
@@ -206,7 +206,7 @@ enum ImageUtils {
         } else {
             directory = try applicationFilesDirectory().appendingPathComponent("images/outgoing", isDirectory: true)
         }
-        try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true, attributes: nil)
+        try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true, attributes: BLEIncomingFileStore.mediaProtectionAttributes)
         return directory.appendingPathComponent(fileName)
     }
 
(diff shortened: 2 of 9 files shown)


Chessing234:
good alignment with the whitepaper. please confirm media files still get deleted on panic wipe under the new protection class

Chessing234:
also: worth a one-liner in PRIVACY_POLICY or the media docs noting the protection class so future contributors dont regress it

heyaim:
Confirmed. The protection class encrypts file contents, not the directory entry, so deletion is unaffected. The panic-wipe tests in `BLEFileTransferHandlerTests` still pass.

Agreed the class is worth writing down.

PR #1557, opened 2026-07-30
add flake.nix for reproducible Nix dev shell

tellmeY18:
Provides a nix develop shell with swiftlint, xcbeautify, and jq for local development. Includes shellHook to strip Nix stdenv vars (DEVELOPER_DIR, SDKROOT, etc.) that otherwise break Xcode/Swift SDK detection on macOS.

Also adds 'just develop' recipe as a convenience alias.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -14,6 +14,7 @@ default:
     @echo "  just build              Build the macOS app without signing"
     @echo "  just test               Run the SwiftPM test suite"
     @echo "  just test-ios           Run tests on the iPhone 17 simulator"
+    @echo "  just develop            Enter the Nix dev shell (swiftlint, xcbeautify, jq)"
     @echo "  just clean              Remove repo-local build artifacts only"
     @echo "  just nuke               Also remove nested package build caches"
     @echo "  just check              Validate the development environment"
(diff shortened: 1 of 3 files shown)


tellmeY18:
#1557

jackjackbits:
Thanks for the contribution, but closing after review — as filed this adds more maintenance surface than it removes:

- **Broken on one of its two declared platforms:** at the pinned nixpkgs rev, `pkgs.swiftlint` has no `x86_64-darwin` build (`sources.json` only carries `aarch64-darwin` + Linux), so `just develop` throws `Unsupported platform` on Intel Macs.
- **It can't pin what the project actually needs.** The real toolchain is Xcode/`xcodebuild`/`swift`, which Nix can't provide (the shellHook's whole job is getting the Nix SDK out of the way). Of the three packages it does pin: `xcbeautify` is used nowhere in the repo, `jq` only appears inside CI scripts, and `periphery` (the other CI lint tool) is absent.
- **The one correct pin drifts silently.** swiftlint 0.65.0 matches CI's docker pin today by coincidence; the flake tracks `nixpkgs-unstable`, so any `nix flake update` moves it with nothing tying it back to CI. This repo has no other Nix infrastructure or Nix-fluent maintenance to catch that.

If you'd like to resubmit, the shape that would work: drop `x86_64-darwin` (or gate swiftlint), use `mkShellNoCC` instead of the env-var unset-hack (note `nix develop --command $SHELL` re-sources rc files, which can reorder PATH and undo the `/usr/bin`-first fix), drop `xcbeautify`, add `periphery` and `just`, pin swiftlint directly (or a release branch) with a comment tying it to the CI version, and leave the Justfile untouched so non-Nix contributors see no change.

Issue #1558, opened 2026-07-30
Fix EnvironmentObject crash inside NavigationStack sheet


tellmeY18:
`ContentPeopleSheetView` crashed when presented inside a `NavigationStack` sheet
because several `@EnvironmentObject` properties weren't available in the sheet's
environment.

PR #1559, opened 2026-07-30
Fix EnvironmentObject crash inside NavigationStack sheet

tellmeY18:
ContentPeopleSheetView was missing @EnvironmentObject declarations for locationChannelsModel, peerListModel, publicChatModel, and privateInboxModel. Its child views inside the NavigationStack (ContentPeopleListView, MessageListView) require these objects, but SwiftUI's sheet presentation context can lose inherited environment objects during the view graph update cycle, causing EnvironmentObject.error() to fatal-error.

Re-inject all eight environment objects on the NavigationStack so child views can always resolve them regardless of environment propagation issues.

### DIFF
--- a/bitchat/Views/ContentSheetViews.swift
+++ b/bitchat/Views/ContentSheetViews.swift
@@ -27,6 +27,10 @@ struct ContentPeopleSheetView: View {
     @EnvironmentObject private var privateConversationModel: PrivateConversationModel
     @EnvironmentObject private var verificationModel: VerificationModel
     @EnvironmentObject private var conversationUIModel: ConversationUIModel
+    @EnvironmentObject private var locationChannelsModel: LocationChannelsModel
+    @EnvironmentObject private var peerListModel: PeerListModel
+    @EnvironmentObject private var publicChatModel: PublicChatModel
+    @EnvironmentObject private var privateInboxModel: PrivateInboxModel
     @Environment(\.scenePhase) private var scenePhase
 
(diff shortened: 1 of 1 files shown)


tellmeY18:
#1558

jackjackbits:
Closing — the premise doesn't hold up against the code:

- All ten models are injected at the app root (`BitchatApp.swift:35-44`), wrapping `ContentView`. The `.sheet` is attached inside `ContentView.body`, and SwiftUI sheet content inherits the environment at the attachment point. The sheet's children (`ContentPeopleListView`, `MessageListView`) already resolve six-plus environment objects through exactly this inheritance in shipping builds that have been field-tested extensively — a deterministic crash-on-present would have been hit by everyone.
- The change can't fix the failure mode it describes: the four new `@EnvironmentObject` properties read the same environment in order to re-inject it, so if the environment were actually missing, this code would fatal-error identically — it just moves the hypothetical crash site up one view.
- The one known env-propagation gap in this area (`navigationDestination` on early NavigationStack OSes) is already handled on main by explicit injection on the destination content (`ContentSheetViews.swift:183`), and modifiers on the NavigationStack wouldn't reach destinations anyway.
- Issue #1558 has no crash log, stack trace, OS version, or repro steps.

Happy to reopen if you can attach a symbolicated crash log with OS version and reproduction steps — that would mean something genuinely different is going on and worth understanding at the root rather than papering over with re-injection.

PR #1570, opened 2026-07-31
feat: quote messages into the composer from the context menu

Chessing234:
## Summary
- Extracts `MessageClipboard` for plaintext copy and quote formatting (testable without SwiftUI).
- Adds a **quote in composer** context-menu action beside copy so nearby chat can be referenced without retyping; existing drafts are preserved.
- Unit tests cover multiline quotes, system-sender handling, and non-destructive append.

## Test plan
- [ ] Long-press a message → copy message — pasteboard has the plaintext body.
- [ ] Long-press → quote in composer — composer gains a `> @sender:` block; prior draft text stays.
- [ ] System rows do not offer quote.
- [ ] CI Build & Test green.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -2051,36 +2051,186 @@
       "comment" : "System message shown in the affected chat when an explicit media delete or /clear was refused and bubbles/files were kept",
       "extractionState" : "manual",
       "localizations" : {
-        "ar" : { "stringUnit" : { "state" : "needs_review", "value" : "تعذّر حذف بعض الوسائط. جرّب حذف الوسائط الأقدم أولاً." } },
-        "bn" : { "stringUnit" : { "state" : "needs_review", "value" : "কিছু মিডিয়া মুছে ফেলা যায়নি। আগে পুরোনো মিডিয়া মুছে ফেলার চেষ্টা করুন।" } },
-        "de" : { "stringUnit" : { "state" : "needs_review", "value" : "einige medien konnten nicht gelöscht werden. versuche zuerst, ältere medien zu löschen." } },
-        "en" : { "stringUnit" : { "state" : "translated", "value" : "some media could not be deleted. try deleting older media first." } },
-        "es" : { "stringUnit" : { "state" : "needs_review", "value" : "no se pudieron eliminar algunos archivos multimedia. prueba a eliminar primero los más antiguos." } },
-        "fa" : { "stringUnit" : { "state" : "needs_review", "value" : "برخی رسانه‌ها حذف نشدند. ابتدا رسانه‌های قدیمی‌تر را حذف کنید." } },
-        "fil" : { "stringUnit" : { "state" : "needs_review", "value" : "hindi ma-delete ang ilang media. subukang i-delete muna ang mas lumang media." } },
-        "fr" : { "stringUnit" : { "state" : "needs_review", "value" : "impossible de supprimer certains médias. essaie d'abord de supprimer les médias plus anciens." } },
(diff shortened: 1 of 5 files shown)


jackjackbits (changes requested):
clean extraction and good tests — request-changes on four things, all small.

1. **i18n**: `content.message.quote` is missing from `Localizable.xcstrings` entirely — visible context-menu copy renders english in 29 locales (`content.message.copy` right above it has all 30). the coverage test only checks keys already in the catalog, so CI can't catch this one.
2. **decide the DM→public path**: `messageText` is one shared draft bound to both the public composer and the DM sheet, cleared only on send. quoting a DM stages its content (plus `@sender`) into a draft that silently travels to the public composer — one tap later it's broadcast. copy-paste allows the same, but this makes it a designed one-tap action. either scope the quote to the conversation it came from, drop it when the composer target changes, or state the accepted risk explicitly in the PR.
3. **decide the mention side effect**: the `> @sender:` header (and any `@name` inside the quoted body) is re-parsed as a live mention on send — the quoted person gets a mention ping, and third parties mentioned inside the quote get re-pinged. if reply-ping is wanted, choose it deliberately; otherwise a non-`@` header (`> sender:`) avoids it. this interacts badly with 2 (quoting a DM publicly names and pings the DM partner).
4. **bug**: `appendQuote` (MessageClipboard.swift:52-54) treats a trailing space like a trailing newline, so the quote marker lands mid-line and doesn't read as a quote. only `hasSuffix("\n")` should skip the newline insert — and please add the trailing-space case to `appendQuotePreservesExistingDraft`.

verified safe, for the record: quoted commands can never execute (every line gets the `> ` prefix and command dispatch is prefix-gated), and there's no composer length cliff — long content fragments at the BLE layer.

Chessing234:
Addressed the four review items:

1. **i18n** — \`content.message.quote\` is now in \`Localizable.xcstrings\` across all 30 locales.
2. **DM→public path** — chose \"drop when the composer target changes\": clearing \`messageText\` on \`selectedPrivatePeerID\` / \`selectedChannel\` changes so a quoted DM cannot silently travel into the public composer.
3. **Mention side effect** — deliberate non-ping: quote header is \`> sender:\` (not \`> @sender:\`). Quoted body lines stay behind \`> \`, so embedded \`@names\` are not live mention tokens either.
4. **Trailing space bug** — \`appendQuote\` only treats \`hasSuffix(\"\\n\")\` as already line-broken; added \`appendQuoteAfterTrailingSpaceStartsNewLine\` coverage.

Ready for another look.

Chessing234:
@jackjackbits Re-review round — tip \`12d009d\`:

**Addressed**
- \`content.message.quote\` in \`Localizable.xcstrings\` (all locales).
- \`appendQuote\` trailing-space bug fixed + test case added.
- Quote header uses plain \`> sender:\` (no \`@\`) to avoid mention re-pings.
- DM→public leak: \`messageText\` cleared on \`selectedPrivatePeerID\` / channel change so a DM quote cannot silently travel to the public composer.

CI green on tip.

Chessing234:
@jackjackbits ready for another look — tip has `content.message.quote` in the catalog, `appendQuote` only skips the separator on a trailing newline (trailing-space case covered by test), and quote/mention scoping from the earlier revision round.

Chessing234:
rebased onto main — quote catalog, plain sender header, draft clears on target change

Chessing234:
@jackjackbits quote fixes on tip, ci green

Chessing234:
rebased onto main. quote key in catalog, `> sender:` header (no @ re-ping), draft clears on target change, trailing-space appendQuote fix + test.

PR #1572, opened 2026-07-31
Deflake VoiceRecorderTests: replace timed semaphores with async events

jackjackbits:
## Problem

`VoiceRecorderTests.cancelWhileSessionAcquireIsInFlightNeverCreatesARecorder` is the last of the three recurring CI flakes — 7 sightings, three in the last two days alone (the #1506 and #1528 merge runs, and #1550's PR run this morning).

Root cause: `waitUntilActivationBegan` hardcodes a **5-second `DispatchSemaphore` timeout** — below the 10s house floor from #1491, and invisible to `TestTimingHygieneTests` because it's a raw semaphore wait rather than a helper-timeout parameter (this exact site was flagged in the July 26 triage as a starvation site the guard can't see). On a starved runner, the window expires before the recorder's session-acquire task gets scheduled.

## Fix — extracted verbatim from #1107

@mmalmi already fixed this inside #1107, but that PR is blocked on a BLE-V3 rebase and shouldn't hold CI stability hostage. This extracts just the `VoiceRecorderTests.swift` change: both test gates (activation and padding) drop their `DispatchSemaphore` + timeout for an untimed async-event wait (`VoiceRecorderAsyncEvent`, private to the file), so **no timing constant remains to starve** — the test framework's own timeout is the backstop. Same "make the flake class unrepeatable" standard as #1491/#1563/#1564.

When #1107 rebases, this file will already match its branch — no conflict, no divergence.

## Verification (count-checked via xcresulttool)

- 7/7 `VoiceRecorderTests` green on the iOS simulator.
- **7/7 × 5 consecutive runs under 16× CPU oversubscription.**

Test-only change; no production code touched.

### DIFF
--- a/bitchatTests/VoiceRecorderTests.swift
+++ b/bitchatTests/VoiceRecorderTests.swift
@@ -10,10 +10,41 @@ import Foundation
 import Testing
 @testable import bitchat
 
+/// One-shot event that bridges synchronous production seams to async tests
+/// without blocking a shared dispatch worker while waiting for the seam.
+private final class VoiceRecorderAsyncEvent: @unchecked Sendable {
+    private let lock = NSLock()
+    private var isSignaled = false
+    private var waiters: [CheckedContinuation<Void, Never>] = []
+
+    func wait() async {
+        await withCheckedContinuation { continuation in
+            let resumeImmediately = lock.withLock { () -> Bool in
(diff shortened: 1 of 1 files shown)


duckniii:
@jackjackbits  Thanks for this

PR #1573, opened 2026-07-31
feat: verification seal on private message sender rows

### LINKED ISSUE #1439: [UI only] Verification badge missing from chat; local alias field exists but unused (secure by default)
**App version:** current `main` (checked July 2026)

UI-only change, no protocol or core logic changes needed.

Following up on #42 and #370, closed as resolved against the current architecture. The verification system (`VerificationService`, `SecureIdentityStateManager`) exists now, but two gaps remain. Security features only help if they're easy to see and use — right now both of these are built but hidden.

---

### 1 — Verified badge not shown in chat

`isVerified` is only used in `FingerprintView.swift` and `VerificationViews.swift` — both profile screens. `ContentView.swift`, where messages actually render, never checks it. Users have to leave the chat to see if someone is verified. In practice, an impersonator looks identical to a real verified contact right in the conversation — the exact scenario verification is meant to prevent.

**Fix:** show a small badge next to the sender's name in chat messages and in peer list rows, using the existing `isVerified` value. Display-only change.

---

### 2 — Local alias field exists but nothing sets it

`IdentityModels.swift` already has:
```swift
var localPetname: String?   // User's name for this peer
```
Already read in `BLEService.swift`, `PrivateConversationModels.swift`, `VerificationModel.swift`, `ChatPeerIdentityCoordinator.swift` — but every place that creates a `SocialIdentity` passes `localPetname: nil`. No UI anywhere sets it.

**Fix:** add a way to rename a peer (long-press or edit icon) that saves through `SecureIdentityStateManager`. Read side is done, just needs a write path.

---

Both fit BitChat's secure-by-default goal: make the security that's already there visible and accessible, not opt-in and buried.

Happy to PR either part if welcome.

Chessing234:
## Summary
- Jack re-scoped #1439: peer-list/DM-header seals already exist; the remaining gap is sender names on message rows, with different trust semantics on public timelines.
- This adds the filled verification seal next to **private** message senders only (text + media rows), matching the peer-list verified glyph.
- Public mesh / geohash timelines stay undressed.

Closes #1439 (petname write path is in #1507; this is the remaining badge half, DM-scoped).

## Test plan
- [ ] Verify a peer, open their DM, send/receive a message — filled seal appears before the sender name
- [ ] Unverify — seal disappears on new renders
- [ ] Own private messages — no seal
- [ ] Public mesh messages — no seal even for verified peers

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -231,6 +231,15 @@ final class ConversationUIModel: ObservableObject {
                 self?.refreshComputedState()
             }
             .store(in: &cancellables)
+
+        // Verify/unverify while a DM is open must repaint existing rows —
+        // showsVerifiedSeal is computed per render, so forward the store change.
+        chatViewModel.peerIdentityStore.$verifiedFingerprints
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.objectWillChange.send()
+            }
+            .store(in: &cancellables)
     }
 
     private func refreshComputedState() {
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -128,6 +128,18 @@ final class ConversationUIModel: ObservableObject {
         message.sender == currentNickname || message.senderPeerID == chatViewModel.meshService.myPeerID
     }
 
+    /// Whether a private-message row should show the filled verification seal
+    /// next to the sender name (#1439). Scoped to DMs only — public timelines
+    /// have different trust semantics and stay undressed.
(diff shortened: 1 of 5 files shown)


Chessing234:
Addressed the Codex note: private rows keep the SF Symbol seal and no longer also append the in-string ✓ from `formatMessage` / `formatMessageHeader`.

Chessing234:
Codex P2 (double verified marker) is addressed on tip — `ChatMessageFormatter` skips the in-string ✓ for private messages where the SF Symbol seal is shown beside the sender.

jackjackbits (approved):
seal predicate verified against the persisted verified-fingerprint store; pushed the missing verified_sender catalog entry (30 locales) + live repaint on verify to the branch. merging.

PR #1575, opened 2026-07-31
fix: three-way logo panic gesture (instant / confirm / off)

Chessing234:
## Summary

Makes the undiscoverable `bitchat/` logo triple-tap a three-way setting: **instant**, **confirm first** (default), or **off**. Panic wipe resets it to confirm.

### 1. Rebased onto #1509's shape

`requestPanicWipe()` stays the single choke point and the confirmation dialog stays on `ContentView` where #1509 put it — this only folds the mode switch into that existing path:

```swift
func requestPanicWipe() {
    switch PanicWipeSettings.logoShortcutMode {
    case .confirm: showPanicConfirmation = true
    case .instant: panicClearAllData()
    case .off:     showPanicGestureDisabledAlert = true
    }
}
```

The duplicate dialog is gone and `ContentHeaderView` is **untouched** — main already routes the triple-tap through `requestPanicWipe()`, so there was nothing left to change there. The disabled-gesture alert is hosted next to your dialog on `ContentView`, for the reason your comment there gives: a host that can be covered or removed would take a pending dialog down with it.

### 2. The migration is deleted, and the default is `.confirm`

You were right and it's worse than a wrong default — the migration read `panic.confirmLogoShortcut` and `panic.logoShortcutEnabled`, neither of which was ever written, so *every* existing install fell through to `.instant`. On update, people who had confirm-before-wipe would have silently been moved to an instant wipe by a PR whose stated purpose was to give them more control.

Fixed by removing the migration rather than correcting it: with no key ever written there is nothing to migrate, and the branch is dead code that can only get the default wrong. An unparseable stored value now falls back to `.confirm` as well, so a corrupt preference can't leave the gesture unarmed either.

`reset()` restores `.confirm`, so a wiped device comes back armed rather than carrying the seized owner's `off` setting.

### 3. Mode-aware copy

`app_info.privacy.panic.description` ("…asks to confirm") and `app_info.settings.danger.panic_note` ("the logo does the same") are both true in confirm mode and wrong outside it. Rather than edit main's entries, the row and the caption select between main's existing key and new `_instant` / `_off` / `_disabled` variants.

### Catalog

Rebased onto current main, spliced surgically: **+2232/−0**, twelve keys, no pre-existing entry modified. Real translations in all 30 locales at `state: "translated"` — the nine you spot-checked in the Aug-9 rework are carried over unchanged; the three new ones (`logo_mode.subtitle`, rewritten because the default changed, plus the two privacy-row variants) are translated to the same bar.

## Verification

- `PanicWipeSettingsTests` now pins the thing that broke: a fresh install reads `.confirm`, the never-shipped legacy keys are ignored, an unknown stored value falls back to `.confirm`, and `reset` returns to `.confirm` from `off`.
- `PanicWipeSettings` is Foundation-only, so I compiled and ran it standalone under `-swift-version 5` — **13/13 assertions pass**, including suite isolation and every mode round-tripping. Clean under `-swift-version 6` too.
- Catalog re-parsed after the splice with every pre-existing entry compared for equality; `everyCodeReferencedKeyExistsInACatalog` re-run locally against both catalogs — 0 missing keys.

**Not run locally:** the Xcode suite. Command Line Tools only on this machine, so `xcodebuild` cannot build the app here — CI covers the build, app tests, simulator tests, SwiftLint and periphery. I have not exercised the picker or seen either dialog render.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -23,6 +23,8 @@ final class AppChromeModel: ObservableObject {
     @Published var showScreenshotPrivacyWarning = false
     /// Triple-tapping the logo asks first; the dialog lives on the header.
     @Published var showPanicConfirmation = false
+    /// Triple-tap while the gesture is off. See `requestPanicWipe`.
+    @Published var showPanicGestureDisabledAlert = false
     /// Mirrors `ChatViewModel.panicRecoveryBlocked` for the chrome: a wipe
     /// that did not commit must be visible, not just logged — the person who
     /// triggered it needs to know data may remain on the device.
@@ -120,11 +122,23 @@ final class AppChromeModel: ObservableObject {
         prepareForPanic = preparation
     }
 
-    /// Entry point for the header triple-tap: confirm before destroying.
-    /// The Settings-pane button has always confirmed; the gesture now goes
-    /// through the same dialog so a mis-tap can't wipe the device.
+    /// Entry point for the header triple-tap. The Settings-pane button has
+    /// always confirmed; the gesture goes through the same dialog by default
+    /// so a mis-tap can't wipe the device.
+    ///
+    /// The mode is resolved here rather than at the gesture site so that
+    /// instant / confirm / off is decided at one choke point.
     func requestPanicWipe() {
-        showPanicConfirmation = true
+        switch PanicWipeSettings.logoShortcutMode {
+        case .confirm:
+            showPanicConfirmation = true
+        case .instant:
+            panicClearAllData()
+        case .off:
+            // A silent no-op is the worst outcome here: someone triple-tapping
+            // under duress would believe the device was wiped. Say it wasn't.
+            showPanicGestureDisabledAlert = true
+        }
     }
 
     func panicClearAllData() {
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -30515,6 +30515,2238 @@
         }
       }
     },
+    "app_info.privacy.panic.description_instant" : {
+      "comment" : "Panic mode privacy row when the logo triple-tap wipes immediately",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "انقر الشعار ثلاث مرات لمسح كل البيانات (بدون تأكيد)"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "সব ডেটা মুছতে লোগোতে তিনবার ট্যাপ করুন (কোনো নিশ্চিতকরণ নেই)"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "logo dreifach tippen, um alle daten zu löschen (ohne rückfrage)"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "triple-tap logo to wipe all data (no confirmation)"
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "toca el logo tres veces para borrar todos los datos (sin confirmación)"
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "برای پاک‌کردن همهٔ داده‌ها سه بار روی لوگو ضربه بزنید (بدون تأیید)"
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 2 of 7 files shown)


jackjackbits (changes requested):
the feature is defensible — default-on and reset-on-wipe both check out, so a wiped device comes back with the gesture armed — but it needs a revision round.

1. **blocker — the CI red is real**: `Localizable.xcstrings` is invalid JSON — missing comma after the `panic_note_disabled` block (~line 14890), which kills `xcstringstool compile` and cascades into all four failing jobs. also fix the stray indentation around `logo_shortcut.subtitle` and `panic_button`.
2. **blocker — fake translations**: all 3 new keys ship the english string into every locale marked `"state": "translated"`. #1509 has real per-locale translations — same bar here.
3. **converge with #1509**: this collides with it hunk-for-hunk (same triple-tap closure, same danger-zone card, same wipe list, same xcstrings region) and introduces a second settings type for the same gesture — and "confirm before wipe" is meaningless while "gesture off". #1509 merges first; please rebase and fold `logoShortcutEnabled` into `PanicWipeSettings` — one type, one `reset()`, one danger-zone control (instant / confirm-first / off).
4. move the enable guard out of the view: post-#1509 the check belongs in `AppChromeModel.requestPanicWipe()` so disable/confirm/instant is decided at one choke point, not at the gesture site (`ContentHeaderView.swift:50`).
5. **safety**: toggling the gesture OFF deserves a confirmation alert — as written, a triple-tap while disabled is a silent no-op, and someone who panics on a disabled gesture gets zero signal that nothing was wiped. that's a worse failure mode than a confirm dialog.

Chessing234:
Rebased onto #1509 and folded per review:

1. Fixed/replaced the broken xcstrings region with real translations for the new keys.
2. One type: \`PanicWipeSettings.LogoShortcutMode\` = instant / confirm / off (migrates legacy confirm/enabled booleans).
3. Gesture enable/confirm/off decided in \`AppChromeModel.requestPanicWipe()\` only.
4. Turning **off** requires a confirmation dialog; a disabled triple-tap surfaces an alert (not a silent no-op).

PR base is now \`fix/panic-wipe-confirm\` (#1509). Ready for another look after that one.

Chessing234:
Codex P1 (missing xcstrings comma) is addressed on tip — `Localizable.xcstrings` validates as JSON again.

Chessing234:
rebased onto main — three-way panic settings still on tip

jackjackbits (changes requested):
The Aug-9 rework fixed everything we asked for July 31 — valid JSON, real translations across all 30 locales (spot-checked, genuinely good), confirm-on-disable dialog, and the explicit "logo wipe is off / nothing was wiped" alert closes the silent-disable trap. The three-way setting itself is well built and panic-wipe-resets-to-instant is the right call for the seizure path.

Two blockers remain:

1. **Hard conflict with merged #1509.** Main's `requestPanicWipe()` now always confirms, with the dialog owned by ContentView. This PR re-implements `requestPanicWipe` as the three-way switch and adds a duplicate dialog on ContentHeaderView. Please rebase onto the #1509 shape: one dialog, owned where main put it, with the mode check folded into the existing path.

2. **The migration silently reverts every existing install to instant wipe.** The legacy-key migration reads `panic.confirmLogoShortcut` / `panic.logoShortcutEnabled` — keys that never shipped (#1509 wrote no key) — so on update every install falls through to the default, which this PR sets to `.instant`. That reverses the confirm-first default that was deliberately merged in #1509. Default must be `.confirm` to preserve current behavior; whether instant should even be the panic-wipe reset target is a separate question we're fine with as you have it.

Also in the rebase: main's catalog copy (`app_info.privacy.panic.description` "…asks to confirm", `panic_note`) becomes wrong under instant mode — those keys need mode-aware or neutral wording.

Chessing234:
rebased onto the #1509 shape. requestPanicWipe is still the one choke point and the dialog is still on ContentView — this just folds the mode switch into that path and puts the disabled-gesture alert next to it. ContentHeaderView is untouched; main already routes the triple-tap through requestPanicWipe so there was nothing left to change there.

the migration was worse than a wrong default, and you called it: neither legacy key was ever written, so every existing install fell through to instant. a pr about giving people more control would have silently moved everyone off confirm-first on update. deleted the migration rather than fixing it — nothing was ever stored, so it's dead code that can only get the default wrong — and an unparseable value now falls back to confirm too. reset goes back to confirm so a wiped device comes back armed.

copy: instead of editing main's panic.description / panic_note, the row and caption pick between main's key and new _instant/_off variants. catalog is +2232/−0, twelve keys, nothing pre-existing modified.

7/7 green. PanicWipeSettings ran standalone, 13/13, including the fresh-install-reads-confirm case. no xcode here so i haven't touched the picker or seen either dialog.

Chessing234:
rebased onto #1509 shape — one dialog on contentview, mode switch in requestPanicWipe, default .confirm (no dead migration). added panic_note_instant for instant-mode caption.

PR #1586, opened 2026-07-31
Mark a DM unread when sent while the app was closed

heyaim:
A DM sent while the app is closed reaches `handlePrivateMessage` through the gift-wrap lookback, so it is minutes or hours old by the time it is handled. The unread decision gated on that age:

```swift
let isRecentMessage = Date().timeIntervalSince(messageTimestamp) < 30
let shouldMarkUnread = !wasReadBefore && !isViewing && isRecentMessage
```

So the message is appended to the conversation but the conversation is not marked unread. Nothing in the UI shows it arrived, and the person finds it only if they happen to open that chat.

The mesh overload of `handlePrivateMessage` in the same file has no recency test at all: if the chat is not on screen, the conversation is marked unread.

```swift
} else {
    context.markPrivateChatUnread(peerID)
    context.notifyPrivateMessage(from: message.sender, message: message.content, peerID: peerID)
}
```

The same message therefore badges when it arrives over the mesh and does not when it arrives over Nostr.

The gate made sense when it was written: it predates the persistent gift-wrap dedup from #1398, and before that store existed, every relaunch replayed the lookback and re-processed old DMs, so recency kept replayed DMs from re-badging and re-alerting on every launch. #1398 now drops replayed events before they reach this code, which leaves the gate suppressing only first-time deliveries.

This change moves the recency test off the badge and onto the notification. Recency still gates the alert, so a reconnect that pulls a batch of old DMs does not fire a burst of them, and nothing new reaches the lock screen. Only the in-app unread state changes.

## Changes

In the Nostr overload of `ChatPrivateConversationCoordinator.handlePrivateMessage`, `isRecentMessage` moves off `shouldMarkUnread` and onto the notification.

## Tests

Three added, one existing assertion changed.

- An old message marks the conversation unread and sends no notification.
- A recent message does both, which keeps the notification path covered.
- A message arriving while that chat is on screen does neither.
- `geoPrivateMessage_sendsDeliveryAckOnceAndDeduplicates` asserted `unreadPrivateMessages.isEmpty`, with a comment explaining that an old timestamp means no unread marking. That assertion depended on the recency gate this change removes, so it now expects the conversation to be unread. Flagging it because it is an existing assertion, not a new one, and the comment above it changed with it.

## Validation

- Full iOS test suite green on the iPhone 17 Pro simulator (Xcode 26.6).
- `swiftlint` adds no violations on the two touched files.

## What this does not change

Notification behavior is untouched: a message older than the 30 second window still does not notify. Read receipts, delivery acks, and dedup are all unchanged.

### DIFF
--- a/bitchat/ViewModels/ChatPrivateConversationCoordinator.swift
+++ b/bitchat/ViewModels/ChatPrivateConversationCoordinator.swift
@@ -548,8 +548,14 @@ final class ChatPrivateConversationCoordinator {
 
         let isViewing = context.selectedPrivateChatPeer == conversationPeerID
         let wasReadBefore = context.sentReadReceipts.contains(messageId)
+        // Recency gates only the notification. A DM sent while the app was
+        // closed reaches this path through the gift-wrap lookback long after
+        // it was sent, so gating the badge on recency left the conversation
+        // looking read and the message findable only by opening that chat.
+        // The mesh overload below marks unread whenever the chat is not on
+        // screen, with no recency test at all.
         let isRecentMessage = Date().timeIntervalSince(messageTimestamp) < 30
-        let shouldMarkUnread = !wasReadBefore && !isViewing && isRecentMessage
+        let shouldMarkUnread = !wasReadBefore && !isViewing
         if shouldMarkUnread {
             context.markPrivateChatUnread(conversationPeerID)
         }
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
Nice separation of concerns: unread badge vs notification recency. The gift-wrap lookback case is exactly where the old `isRecentMessage` gate hurt, and the tests pin both sides (old → badge only, fresh → badge + notify).

Minor: the mesh overload comment says it already marks unread with no recency test — worth a quick assert in an existing mesh private-message test that those two paths stay aligned so a future refactor doesn’t re-couple them. Otherwise LGTM on the Nostr/geo path.

PR #1587, opened 2026-08-01
fix: prevent changing the creator of an existing group during upsert

areebahmeddd:
## Summary

* `applyGroupState` validates incoming group state, but misses an ownership check when updating an existing group
* this allows an attacker to **create** a valid-looking group state that replaces the existing roster and encryption key for a `groupID`, leading to silent group takeover
* fixes the issue in `upsert`, the single entry point for group state updates, without affecting legitimate creator-side key rotation


Chessing234 (commented):
Solid choke-point fix — pinning creator identity in `upsert` is the right place, and the hijack test covers the exact `applyGroupState` miss.

One ask before merge: confirm rotation / member-remove paths that rebuild a `BitchatGroup` always preserve `creatorFingerprint` (the happy-path test covers same-creator epoch bump, but a false reject on a legitimate roster update would be painful). Also worth a one-line log of the existing vs claimed creator fingerprints when refusing, so field reports of “group updates silently fail” are debuggable.

areebahmeddd:
yes, checked. `creatorFingerprint` is immutable, so `rotateKey` can't change it and invite/remove already follow the same path. The existing rotate key test already covers this flow

one thing is that: if the creator rotates their Noise static key, existing peers won't accept future group updates. they would need to leave and re-create the group, which is an intentional tradeoff of pinning the creator key

Chessing234:
nice catch — pinning creator at upsert + before the removal branch closes the silent takeover cleanly.

one edge: if a stored group somehow has a nil/empty creator fingerprint (legacy/corrupt disk), the mismatch check could refuse legitimate creator rotations. worth a quick guard or a test for that path if it can still appear.

areebahmeddd:
nil can't make it through decode since `creatorFingerprint` is required, and an empty value is already rejected by `upsert`. The only gap was `loadFromDisk`, which assigns groups directly, so I added the same creator invariant check there.

good to merge? cc @jackjackbits

areebahmeddd:
closing this due to inactivity..

PR #1588, opened 2026-08-01
Make the panic wipe confirmable and its outcome visible

jackjackbits:
## What

The triple-tap panic wipe was the only unrecoverable accidental-data-loss path in the app: it fired **instantly, unconfirmed**, on the same 18pt logo whose single tap opens App Info — the one control the empty state tells new people to tap. Meanwhile the far less destructive triple-tap-to-clear-chat asks first, and the Settings-pane panic button has always confirmed. The risk ordering was inverted.

- **Triple-tap now raises the same confirmation dialog the Settings pane uses** (`wipe all data?` / `wipe everything`). One extra tap under duress; zero accidental wipes. Reuses the existing localized strings.
- **Successful wipe leaves a system message** — `all data wiped — new identity created`. In duress, "did it work?" must not be a guess; the natural response to uncertainty is tapping again.
- **A wipe that doesn't commit is now visible.** `panicRecoveryBlocked` previously had no UI binding anywhere: the app looked wiped, sat silently offline, and keys could still be on disk. A persistent red banner under the header now says data may remain and that relaunching retries the wipe.
- Settings/privacy copy no longer claims the triple-tap wipes "instantly" (updated in all 30 locales; 2 new keys added with full locale coverage).

## Tests

- `ChatViewModelPanicTests`, `LocalizationCoverageTests`, `AppArchitectureTests` — 32/32 green on iPhone 17 Pro sim (count-verified via xcresulttool).
- New test pins that `requestPanicWipe()` only raises the dialog and never wipes directly.
- `panicClearAllData_delegatesToTransport` updated: the timeline now holds exactly one system message (the wipe confirmation) instead of being empty.
- iOS + macOS builds green.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -125,6 +125,16 @@ final class AppChromeModel: ObservableObject {
     }
 
     func panicClearAllData() {
+        // A wipe invalidates everything on screen, and its outcome must be
+        // visible: the success message and the failed-wipe banner both live
+        // on the root timeline, so a sheet left up (the App Info danger-zone
+        // path keeps its sheet presented) would hide the one signal that says
+        // whether the wipe worked.
+        isAppInfoPresented = false
+        isLocationChannelsSheetPresented = false
+        isNoticesSheetPresented = false
+        showingFingerprintFor = nil
+
         prepareForPanic?()
         onPanicWipe()
         chatViewModel.panicClearAllData()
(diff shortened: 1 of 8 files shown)


Chessing234:
confirm-before-wipe on the logo is the right call — that control was way too easy to hit by accident.

small nit: the confirmation dialog is attached to the logo view itself. if anything ever covers/hides the header chrome while `showPanicConfirmation` is true, the dialog can vanish with it. parking it on `ContentView` (same as the blocked banner) would be a bit more durable.

jackjackbits:
Both addressed in the follow-up commit:

- **Codex P1 (banner under the App Info sheet):** `panicClearAllData` now dismisses all chrome-owned presentation (App Info, location channels, notices, fingerprint) before wiping, so the outcome — success message or failed-wipe banner — is always visible. Pinned by a new test (`panicWipeDismissesChromePresentation`).
- **Dialog host:** moved the confirmation dialog off the logo `Text` onto the ContentView header stack, next to the blocked banner, as suggested.

PR #1589, opened 2026-08-01
feat: search/filter nicknames in the people sheet

Chessing234:
## Summary
- Adds a nickname search field to the people sheet and filters mesh, geohash, bridge, and group rows with a shared `PeopleNameFilter`.
- Catalog strings for all locales; unit tests cover empty/case-insensitive matches.
- Filtering is display-only (does not mutate `PeerListModel` order).

## Test plan
- [ ] Open people sheet on mesh — type a nickname fragment; unrelated rows hide
- [ ] Clear the field — full list returns with sticky order intact
- [ ] Location channel people sheet — same filter behavior
- [ ] `swift test --filter PeopleNameFilterTests`

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -2470,7 +2470,7 @@
       "shouldTranslate" : false
     },
     "%@ active" : {
-      "comment" : "A label at the bottom of the people list sheet showing the number of active users.",
+      "comment" : "A label at the bottom of the people list sheet showing the number of active people.",
       "extractionState" : "stale",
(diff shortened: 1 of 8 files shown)


Chessing234:
Addressed the Codex P2 on sticky order: `orderedIDs` / `onChange` now track the unfiltered geohash roster; the name filter only affects which rows render. Clearing the search no longer reshuffles Alice/Carol after Bob.

Chessing234:

Chessing234:
awaiting-review

PR #1595, opened 2026-08-01
Stop claiming encryption and privacy properties that aren't live

jackjackbits:
## What

Fixes the two worst false-security signals found in the design/UX audit.

### Sticky-optimistic encryption status

`getEncryptionStatus` returned `.noiseSecured`/`.noiseVerified` for **any peer whose fingerprint was ever persisted**, regardless of the live session state. After a cold launch or a handshake **failure**, the DM header showed a solid lock (accessibility: "encrypted"), and the composer caption claimed `private · end-to-end encrypted` — with no secure session in existence. This defeated the caption's own stated principle ("claims end-to-end encryption only once the session is actually secured", `ContentSheetViews.swift`).

The mapping now reflects only the live session:

| session | before (known peer) | after |
|---|---|---|
| established | secured/verified | secured/verified |
| handshaking/queued | **secured/verified** | handshaking (`lock.rotation`) |
| none | **secured/verified** | no handshake |
| failed | **secured/verified** | failure (`lock.slash`, red) |

A remembered fingerprint now only decides what an *established* session upgrades to (verified vs secured). Peers reachable **only over Nostr** keep the encrypted caption — gift-wrapped delivery is E2E without a Noise session (`availability == .nostrAvailable`).

### Privacy copy that wasn't true (all 30 locales)

- **"ephemeral identity — new peer ID generated regularly"** → the peer ID is stable until a panic wipe; rotation is spec-only (`docs/PEER-ID-ROTATION.md`, `IdentityModels.swift` says so explicitly). For the stated audience, a Privacy screen claiming rotating identifiers when the identifier is a stable device fingerprint is copy someone could make a coverage decision on. Now: *"local-only identity — your identity is a keypair created on this device; panic wipe replaces it instantly."*
- **"no servers, accounts, or data collection"** → location channels and internet delivery go through Nostr relays. Now: *"no accounts, no phone numbers, no data collection."*
- **location channels: "your IP address is hidden by routing all traffic over tor"** → tor is a toggle. The description now says tor is the default, and the sheet shows the existing red tor-off warning (`app_info.settings.tor.off_warning`, no new strings) right at the point of joining whenever the toggle is off — the warning belongs where the exposure is about to happen, not only in settings.

## Tests

- New `getEncryptionStatus_reflectsLiveSessionNotHistory` pins the honest mapping (incl. failed → `.none` and verified-fingerprint-no-session → `.noHandshake`).
- `ChatPeerIdentityCoordinatorContextTests` + `AppArchitectureTests` + `LocalizationCoverageTests`: 31/31 green (count-verified).
- iOS + macOS builds green.

### DIFF
--- a/bitchat/App/PrivateConversationModels.swift
+++ b/bitchat/App/PrivateConversationModels.swift
@@ -357,7 +357,10 @@ final class PrivateConversationModel: ObservableObject {
         }
         if let noiseKey = Data(hexString: headerPeerID.id),
            let favoriteStatus = FavoritesPersistenceService.shared.getFavoriteStatus(for: noiseKey),
-           favoriteStatus.isMutual {
+           favoriteStatus.isMutual,
+           // No stored recipient key means no Nostr delivery — and the
+           // "end-to-end encrypted" caption keys off .nostrAvailable.
+           favoriteStatus.peerNostrPublicKey != nil {
             return .nostrAvailable
         }
         if chatViewModel.meshService.isPeerConnected(headerPeerID) || chatViewModel.connectedPeers.contains(headerPeerID) {
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -11560,181 +11560,181 @@
         "ar" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "هوية محلية فقط"
+            "value" : "هوية محفوظة على الجهاز"
           }
         },
         "bn" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "শুধুই স্থানীয় পরিচয়"
+            "value" : "ডিভাইসে থাকা পরিচয়"
           }
         },
         "de" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "nur lokale identität"
+            "value" : "gerätelokale identität"
           }
         },
         "en" : {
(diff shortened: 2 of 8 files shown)


Chessing234:
agree with the live-session lock fix.

one gap: availability == .nostrAvailable can still be true for a mutual favorite with a nil peerNostrPublicKey, so the caption can claim e2e with no noise and no usable nostr key. gate on an actual recipient key (same idea as NostrTransport.refreshReachablePeers), not header availability alone — and a regression for favorite / no session / nil key → unencrypted caption would help.

also "local-only identity" is better, but once someone verifies/shares a fingerprint it isn't only local. something like "device-local keypair" / "not a server account" might age better.

jackjackbits:
Both points addressed:

- **Nostr availability now requires a stored recipient key.** Fixed at the source — `BitchatPeer.connectionState` returns `.nostrAvailable` only when `nostrPublicKey ?? favoriteStatus?.peerNostrPublicKey` is non-nil (same rule as `NostrTransport.refreshReachablePeers`), and the header-state fallback in `PrivateConversationModel` got the same guard. Regression test added: mutual favorite + nil key → `.offline` (caption stays "private conversation"); with a key → `.nostrAvailable`.
- **Copy:** "local-only identity" → "device-local identity" in all 30 locales — agreed that "only" over-claims once a fingerprint/QR has been shared.

PR #1596, opened 2026-08-01
Remove the public-channel screenshot broadcast

jackjackbits:
## What

From the design/UX audit's critical tier: screenshots announced themselves to the whole channel with no opt-out.

`ChatLifecycleCoordinator.handleScreenshotCaptured` sent `* nickname took a screenshot *` to the active public channel — as a mesh broadcast, or on geohash channels as a **mined ephemeral event published to the 5 nearest Nostr relays**, permanently recording that this nickname was present and active at that geohash at that timestamp. Screenshotting police conduct, a threat, or evidence for a lawyer are core protest-app use cases, and every one of them outed the person doing it. The partial suppressions that already existed (location sheet / App Info open) showed the risk was known; this finishes the job.

- **Public channels no longer send anything on screenshot.** The mesh branch and the geohash mined-event helper are gone; the `ChatLifecycleContext` protocol drops its now-unused requirements (`sendMeshMessage`, `isTeleported`, `recordGeoParticipant`, `addSystemMessage`).
- **Geohash timeline screenshots now raise the existing local privacy warning** (`content.alert.screenshot.message`, same alert the channel sheet uses) — warn the person, tell no one. No new strings.
- **DM notices stay, but the local echo is honest now.** `you took a screenshot` used to render unconditionally while the actual notice was silently skipped without an established Noise session — implying the peer was told when they weren't. The echo now only appears when the notice actually went out.

Receive-side rendering of screenshot notices from other clients is unchanged (wire compat with Android).

## Tests

- `handleScreenshotCaptured_publicChannel_staysSilent` (new) pins that no DM selected → nothing sent, nothing echoed.
- `handleScreenshotCaptured_privateChat_echoesOnlyWhenPeerWasNotified` (updated) pins the echo-only-when-sent contract.
- `ChatLifecycleCoordinatorContextTests` + `ChatViewModelServiceLifecycleTests`: 8/8 green (count-verified).
- iOS + macOS builds green.

### DIFF
--- a/bitchat/App/AppRuntime.swift
+++ b/bitchat/App/AppRuntime.swift
@@ -424,26 +424,24 @@ private extension AppRuntime {
     }
 
     func handleScreenshotCaptured() {
-        if appChromeModel.isLocationChannelsSheetPresented {
+        let isLocationChannelActive: Bool = {
+            if case .location = chatViewModel.activeChannel { return true }
+            return false
+        }()
+
+        switch Self.resolveScreenshotResponse(
+            isLocationChannelsSheetPresented: appChromeModel.isLocationChannelsSheetPresented,
+            isAppInfoPresented: appChromeModel.isAppInfoPresented,
+            hasPrivateChatOpen: chatViewModel.selectedPrivateChatPeer != nil,
+            isLocationChannelActive: isLocationChannelActive
+        ) {
+        case .warnLocally:
             appChromeModel.triggerScreenshotPrivacyWarning()
-            return
+        case .ignore:
+            break
+        case .forwardToChat:
+            chatViewModel.handleScreenshotCaptured()
         }
-
-        if appChromeModel.isAppInfoPresented {
-            return
-        }
-
-        // Screenshots are never announced to public channels (see
-        // ChatLifecycleCoordinator.handleScreenshotCaptured). A geohash
-        // timeline screenshot still reveals a place, so warn the person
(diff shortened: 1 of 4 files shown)


Chessing234:
couple thoughts:

mesh screenshots still get no local cue — appruntime only warns on geohash. same protest-screenshot risk; either fire the privacy alert there too or call out why mesh is exempt.

would also help to pin the geohash warning path (location + screenshot → privacy alert, nothing sent). coordinator tests cover the silent send, but the warn lives in appruntime.

android can still broadcast — worth a note/issue that mixed rooms can still leak from the other client.

jackjackbits:
All three points:

- **Geohash warn path pinned:** the routing decision is now a pure table (`AppRuntime.resolveScreenshotResponse`) with tests covering location-sheet → warn, geohash timeline + no DM → warn (and the coordinator tests already pin that nothing is sent), App Info → nothing, mesh/DM → forward.
- **Why mesh is exempt** is now documented on the response enum: a mesh screenshot reveals no place and triggers no send, so there's nothing actionable to warn about — an alert on every screenshot would train people to dismiss the one alert that matters (the location one). Happy to add a cue there too if you'd rather over-warn.
- **Android:** correct — a mixed room can still see `* X took a screenshot *` lines from Android clients until it ships the same change; receive-side rendering is deliberately unchanged for wire compat. Flagging for the Android tracker is probably worth an issue there.

PR #1597, opened 2026-08-01
Add a persistent connectivity banner and stop the radar lying

jackjackbits:
## What

Two "the app looks fine while it isn't" findings from the design/UX audit.

**Bluetooth-off had exactly one signal: a dismissible alert.** Once dismissed, the app looked completely normal — empty timeline, radar sweeping "searching for people" — while the radio was off (`MeshEmptyStateView.isSearchingForPeers` only checked peer counts). Now:

- The radar only sweeps when the radio can actually scan (`bluetoothAvailable` passed from `MessageListView`); an animated search over a dead radio is an actively false status display.
- A persistent red banner under the header carries the state instead: **off** → "bluetooth is off — nobody nearby can reach you. tap to fix." (deep-links to settings), **denied** → tap to open settings, **unsupported** → mesh unavailable, location channels still work (the old alert copy claimed the whole app "requires bluetooth", which isn't true).

**A Tor bootstrap stall was only announced inside geohash timelines** (`addGeohashOnlySystemMessage`), so anyone sitting in `#mesh` or a DM routing over Nostr got silence and hanging `sent` states. The stall handler now also sets a chrome-level `ChatViewModel.torBlocked` (cleared on ready / preference change), mirrored to `AppChromeModel`, driving the same banner: "tor can't connect — internet features are paused. mesh still works."

Design notes:

- The banner renders **nothing** when healthy — no permanent chrome.
- `.unknown`/`.resetting` radio states deliberately stay quiet: a false "bluetooth is off" flash during launch would be the same category of lie this fixes.
- Priority (bluetooth outranks tor) and the quiet-start rule live in a pure `ConnectivityIssue.resolve(bluetoothState:torBlocked:)` so they're testable without a view.
- The existing one-shot Bluetooth alert is unchanged; the banner is the persistent complement.

## Tests

- New `ConnectivityIssueTests` pin priority order and quiet-start behavior.
- `ConnectivityIssueTests` + `LocalizationCoverageTests` + `ViewSmokeTests`: 21/21 green (count-verified).
- 4 new strings with all-30-locale coverage; iOS + macOS builds green.

### DIFF
--- a/bitchat/App/AppChromeModel.swift
+++ b/bitchat/App/AppChromeModel.swift
@@ -17,6 +17,9 @@ final class AppChromeModel: ObservableObject {
     @Published var showBluetoothAlert = false
     @Published var bluetoothAlertMessage = ""
     @Published var bluetoothState: CBManagerState = .unknown
+    /// Tor bootstrap has stalled (network likely blocks it); drives the
+    /// connectivity banner. Mirrored from `ChatViewModel.torBlocked`.
+    @Published private(set) var torBlocked = false
     @Published var showScreenshotPrivacyWarning = false
 
     private let chatViewModel: ChatViewModel
@@ -144,6 +147,10 @@ final class AppChromeModel: ObservableObject {
         chatViewModel.$bluetoothState
             .receive(on: DispatchQueue.main)
             .assign(to: &$bluetoothState)
+
+        chatViewModel.$torBlocked
+            .receive(on: DispatchQueue.main)
+            .assign(to: &$torBlocked)
 
         hasUnreadPrivateMessages = !privateInboxModel.unreadPeerIDs.isEmpty
     }
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -2047,6 +2047,750 @@
         }
       }
     },
+    "content.banner.bluetooth_denied" : {
+      "comment" : "Persistent banner while Bluetooth permission is denied; tapping opens system settings",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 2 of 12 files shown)


Chessing234:
banner + honest radar look right.

a few nits:
- people/dm sheet covers the root header, so the banner disappears there — worth mirroring ConnectivityIssue in ContentSheetViews
- on macos, "tap to fix" for bt-off opens Privacy_Bluetooth, not power controls. maybe split denied → privacy settings vs powered-off → copy that says turn bluetooth on (no dead deep link)
- one test that torBlocked clears when tor is turned off would be nice
- optional: if user is mesh-only, soften/hide the tor stall line so they aren't nagged about internet features they aren't using

Chessing234 (commented):
for .bluetoothOff, tap-to-fix still calls SystemSettings.bluetooth.open(), which on macos opens the privacy pane, not power. powered-off probably needs different copy/action than denied.

also the banner is only on the root headerView, so the people/dm sheet covers it — same place that already mirrors the bluetooth alert.

jackjackbits:
Addressed:

- **Banner now mirrors into the people/DM sheet** via a top safe-area inset on the sheet's navigation content — covers both the people list and the open DM, same `ConnectivityIssue.resolve` source.
- **Powered-off vs denied now route to different places:** new `SystemSettings.bluetoothPower` opens the macOS Bluetooth pane (radio controls); `.bluetoothDenied` keeps the privacy anchor. The one-shot Bluetooth alert's settings button routes by state the same way.
- **Tests added:** `torBlocked` clears on `handleTorDidBecomeReady` and on `handleTorPreferenceChanged` (toggling tor off can't leave a stale banner).
- **Mesh-only softening:** left as-is deliberately — a tor stall also pauses DM delivery to offline favorites, which mesh-first people still rely on, and there's no explicit mesh-only mode to key off. The line only appears when the tor preference is on and bootstrap actually stalled; if that reads as nagging in practice, a dismiss-for-this-session affordance would be my next step.

PR #1598, opened 2026-08-01
Add a recent-chats section so DM conversations can't become unreachable

jackjackbits:
## What

The largest information-architecture hole from the design/UX audit: **a DM conversation could become unreachable while still in memory.**

`UnifiedPeerService` filters the roster to connected / reachable / mutual-favorite; the header envelope only exists while `hasUnreadPrivateMessages`; `/msg` can't resolve offline strangers; and `openMostRelevantPrivateChat()`'s recent-chats fallback was unreachable code (its only caller is gated on unread). Read a DM from a passerby, close the sheet, and the thread had no row anywhere in the UI.

### The fix

- **`PeerListModel.recentChatRows`**: derived from the `ConversationStore`'s `.direct` conversations — people **absent from the rosters above it**, newest activity first, with unread state. Conversations mirrored under both an ephemeral and a stable peer ID collapse to one row per fingerprint (newest wins), matching how the private-chat coordinators consolidate on open. Groups (own section) and blocked peers are excluded.
- **`RecentChatList`**: a "chats" section in the people sheet, same `PeopleSectionHeader` shape as `#mesh`/groups — name, relative timestamp, unread envelope, full VoiceOver description. Renders nothing when empty. Present on **both** mesh and location channels, so a geoDM opened from a channel you've since left stays reachable too (this also plugs part of the geohash-unread blind spot).
- Tap → existing `startConversation(with:)` path (peer-ID migration/consolidation already handled there; sends to unreachable peers route via outbox/courier since #1415).
- Rebuilds subscribe to the store's `changes` **filtered to direct conversations**, so public-timeline traffic (which emits a change per message) doesn't churn the sheet, and rows participate in `renderID`.

Deliberately *not* in this PR: persistence (private chats are RAM-only by design — this list covers the session, which is where the hole was) and any restructuring of the people-sheet/DM modal double duty.

## Tests

- New `peerListModelSurfacesRecentChatsForAbsentPeers` pins: offline stranger's thread gets a row, group threads don't, roster stays clean.
- `AppArchitectureTests` + `LocalizationCoverageTests` + `ViewSmokeTests`: 40/40 green (count-verified).
- 2 new strings with all-30-locale coverage; iOS + macOS builds green.

### DIFF
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -39,12 +39,27 @@ struct GroupChatRow: Identifiable, Equatable {
     var id: String { peerID.id }
 }
 
+/// A direct conversation whose person is NOT in any roster above it. Without
+/// this section, a DM from a passerby became unreachable the moment they went
+/// offline: the roster filters to connected/reachable/mutual-favorite, the
+/// header envelope only exists while unread, and /msg can't resolve offline
+/// strangers — the thread was still in memory with no row anywhere in the UI.
+struct RecentChatRow: Identifiable, Equatable {
+    let peerID: PeerID
+    let displayName: String
+    let hasUnread: Bool
+    let lastActivity: Date
+
+    var id: String { peerID.id }
+}
+
 @MainActor
 final class PeerListModel: ObservableObject {
     @Published private(set) var allPeers: [BitchatPeer] = []
     @Published private(set) var meshRows: [MeshPeerRow] = []
     @Published private(set) var geohashPeople: [GeohashPersonRow] = []
     @Published private(set) var groupRows: [GroupChatRow] = []
+    @Published private(set) var recentChatRows: [RecentChatRow] = []
     @Published private(set) var reachableMeshPeerCount = 0
     @Published private(set) var connectedMeshPeerCount = 0
     @Published private(set) var visibleGeohashPeerCount = 0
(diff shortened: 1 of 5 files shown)


Chessing234:
solid fix for the unreachable offline dm hole.

one gap: `buildRecentChatRows` only treats mesh rows as "already visible". a geo person still in the geohash section can also show under chats — worth adding geohash peer ids (and fingerprints if you have them) to `visibleIdentities` the same way.

Chessing234 (commented):
buildRecentChatRows only subtracts meshRows, so a geo person already in geohashPeople can also land under recent chats. fold current geohash people (and PeerID(nostr_:)) into the visible set.

also geoDM rows go through resolveNickname, which treats nostr_… as hex and falls through to anonnost — geohashDisplayName(for:) is the right resolver there.

jackjackbits:
Both addressed:

- **Geo dedup:** visible geohash participants now join `visibleIdentities` via their `PeerID(nostr_:)` form, so someone in the geohash section can't also appear under chats. (Geo people have no fingerprints, so the pubkey-derived key form is the whole mapping.)
- **GeoDM names:** `isGeoDM` candidates resolve through `geohashDisplayName(for:)` (Nostr key mapping, bare-key fallback) instead of `resolveNickname`'s mesh-only anon fallback.

Test extended to pin both: a geoDM thread gets a properly-named row, and the row collapses while that pubkey is visible in the geohash roster.

Chessing234:
Hi Jack, thank you for making the relevant changes across all PRs. If you could review the large backlog of PRs I have, it would be greatly appreciated. thanks!

PR #1599, opened 2026-08-01
Localize non-redacted notification titles and nearby bodies

Ayush7614:
## Summary
- Localize the preview-on lock-screen titles for mentions and DMs (redacted path was already localized).
- Localize the nearby-peer notification title and singular/plural bodies.

## Test plan
- [x] `NotificationServiceTests` updated to assert via the same localized formatters
- [x] Manual: with previews on, trigger mention/DM/nearby alerts and confirm localized

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -65869,6 +65869,941 @@
         }
       }
     },
+    "notification.mention.title" : {
+      "comment" : "Lock-screen notification title when someone is mentioned; %@ is the sender nickname",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "🫵 you were mentioned by %@"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "🫵 you were mentioned by %@"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "🫵 du wurdest von %@ erwähnt"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "🫵 you were mentioned by %@"
+          }
+        },
+        "es" : {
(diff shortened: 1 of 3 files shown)


Chessing234:
nice, the mention/dm titles are properly localized. one issue: notification.nearby.body splits singular/other by hand in swift instead of using the app's own plural variation support (see content.accessibility.people_count, which already has proper one/few/many/other categories). that flattens languages like arabic/russian/polish into a single bucket. can you switch nearby.body to a real plural variation instead?

Ayush7614:
@Chessing234 

Addressed in the latest commit: merged `notification.nearby.body.one` / `.other` into a single `notification.nearby.body` key using `%#@people@` plural variations (same pattern as `content.accessibility.people_count`, including few/many/one/other/two/zero where that key already had them). Swift no longer branches on `peerCount == 1`; it formats through `NotificationService.nearbyBody(peerCount:)`.

jackjackbits (changes requested):
Half superseded, half genuinely valuable:

- **Superseded:** the nearby notification title/bodies — #1656 (merged Aug 10) added `notification.nearby.title` + `body_one`/`body_many` on main.
- **Still needed:** the non-redacted mention and DM titles — `"🫵 you were mentioned by \(sender)"` and `"🔒 DM from \(sender)"` (NotificationService.swift:232/240) are still hardcoded on main. That's a real blind spot (interpolated literals are invisible to our coverage test), and your translations for those two are good in the 15 major locales.

One thing worth keeping from the superseded half: your `%#@people@` plural-substitution approach for the nearby body is technically better than main's hand-split `body_one`/`body_many` (it handles ru/ar plural categories properly). If you want to include a migration of main's keys to that form in the rework, we'd take it.

**Asks:**
1. Drop the superseded nearby title/body keys (or convert them to the plural-substitution migration above) and rebase onto the current catalog — the PR is currently CONFLICTING.
2. Fix the 14-locale English-paste (ar, bn, fa, fil, he, id, ms, ne, pt, ta, th, uk, ur, vi marked `"translated"` with English values) — every key needs real translations in all 30 locales.

PR #1602, opened 2026-08-01
Localize media-send system messages for mesh-only and failures

Ayush7614:
## Summary
- Localize mesh-only voice/image system messages.
- Localize image prepare-failure and too-large system messages.
- Update coordinator and view-model tests to assert via the same localized strings.

## Test plan
- [x] Updated `ChatMediaTransferCoordinatorContextTests` / `ChatViewModelExtensionsTests`
- [x] Manual: try sending voice/image outside mesh and confirm localized system lines

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -2083,6 +2083,754 @@
         "zh-Hant" : { "stringUnit" : { "state" : "needs_review", "value" : "部分媒體無法刪除。請先嘗試刪除較舊的媒體。" } }
       }
     },
+    "media.system.voice_mesh_only" : {
+      "comment" : "System message when the user tries to send a voice note outside a mesh chat",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Voice notes are only available in mesh chats."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Voice notes are only available in mesh chats."
+          }
+        },
+        "de" : {
+          "stringUnit" : {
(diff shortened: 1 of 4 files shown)


jackjackbits (changes requested):
Half of this was superseded by #1656 (merged Aug 10): the "Voice notes/Images are only available in mesh chats." strings are now localized on main via literal-string keys. The other half is genuinely still needed — `"Failed to prepare image for sending."` (×2 call sites) and `"Image is too large to send."` remain hardcoded in `ChatMediaTransferCoordinator.swift` on main.

**Asks:**
1. Trim the PR to the two surviving strings and rebase onto the current catalog (the xcstrings conflict is real but small at that size).
2. Fix the translation gap: 14 locales (ar, bn, fa, fil, he, id, ms, ne, pt, ta, th, uk, ur, vi) have English pasted with `state:"translated"` — that marks half the catalog as done when it isn't. The 15 locales you did translate are good quality; the other 14 need the same treatment.

Chessing234 (commented):
the catalog is surgical, which is the hard part — but 14 locales are english marked `state: "translated"`: ar, bn, fa, fil, he, id, ms, ne, pt, ta, th, uk, ur, vi. e.g. `media.system.image_too_large` is "Image is too large to send." in vi and th, while de/es/fr are real.

ci won't catch it — `mainCatalogCoversAllLocalesForEveryKey` only checks the value is non-empty, not that it differs from the source. so this is green and still ships english to those 14.

separate thing on the tests: `#expect(context.systemMessages == [String(localized: "media.system.voice_mesh_only", ...)])` compares the lookup against itself. both sides fall back to the same defaultValue if the key is missing, so it can't catch a wrong or empty catalog value the way the literal it replaced could. keeping the literal on the assert side would pin the text and still fail if the key changed.

PR #1604, opened 2026-08-01
Compile mention autocomplete regex through SafeRegex

Ayush7614:
## Summary
- Route `AutocompleteService`’s mention pattern through `SafeRegex.compile` so a bad pattern logs instead of silently disabling `@` autocomplete via `try?`.
- Drop the stale commented command-autocomplete path and clarify that slash suggestions live in `CommandSuggestionsView`.

## Test plan
- [x] Type `@` in the composer with peers present and confirm mention suggestions still appear
- [x] No behavior change expected for a valid pattern

### DIFF
--- a/bitchat/Services/AutocompleteService.swift
+++ b/bitchat/Services/AutocompleteService.swift
@@ -2,7 +2,7 @@
 // AutocompleteService.swift
 // bitchat
 //
-// Handles autocomplete suggestions for mentions and commands
+// Handles autocomplete suggestions for mentions
 // This is free and unencumbered software released into the public domain.
 //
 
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
SafeRegex for the mention pattern is good. anything calling the old NSRegularExpression path for the same autocomplete should go through the same helper so a bad pattern fails closed the same way.

Ayush7614:
Addressed Chessing234 feedback: mention autocomplete now uses MessageFormattingEngine.Patterns.mentionAutocomplete (SafeRegex.compile) alongside the other message patterns, instead of a separate private regex instance in AutocompleteService.

PR #1606, opened 2026-08-02
feat: per-conversation notification mute for DMs and geohash channels

### LINKED ISSUE #1594: Feature: Add per-conversation notification controls
There is current support only for controlling the hiding of notification previews. There is no ready-to-use functionality to mute notifications for individual direct conversations or geohash channels. 
This can be particularly noisy for crowded geohash channels where users wish to remain active but not receive alerts all the time.

### Proposal

Add a persistent mute/unmute control for:

- Individual DMs
- Individual GeoHash channels

This only mutes local notifications, and should not affect message reception, storage, read receipts, or any other state tracking. Existing Notification Preview settings should also remain the same.

My current thinking is:

- Conversations are unmuted by default
- Private-chat Headers should expose a Mute/Unmute Action
- Active Geohash-channel Headers expose the same action
- Muted state should persist across application launches
- Direct conversation preferences should use stable identities where available and not depend solely on changeable keys like nicknames
- Panic wipes remove all muted conversation preferences (Not necessary but seems like expected behaviour)
- UI should provide localized accessibility labels and visibly indicate the muted state

Other ideas that don't have to be included in this PR is timed mutes (e.g 8 hour, 1 week, 1 year similar to Whatsapp), "Mentions-only" mode to receive only notifications that explicitly mention you, per-device proximity notification controls as discussed in #1523 , OS notification scheduling, Mute preferences syncing between devices.

### Implementation

Perhaps a small injectable preference store that exposes conversation-scoped keys and is consulted by `NotificationService` before constructing the notification request. This should attempt to reflect the existing injected preview-policy approach. Happy to implement this in a focused PR

Chessing234:
## Summary
Per-conversation notification mute for DMs and geohash channels (#1594). Muted conversations still receive and store messages; only local notifications are suppressed. Panic wipe clears preferences.

## Review follow-up
Rebased onto current main and reworked for jack's blockers:

1. **Catalog** — five `notification.mute.*` keys added surgically (no full xcstrings rewrite). `en` is `translated`; other locales are `new` with the source string so the coverage guard stays green without marking English as done.
2. **Mute-key asymmetry** — toggle, notification path, and people-sheet badge all resolve through `stableConversationIdentity(for:)`. No peer-ID fallback; pre-handshake DMs disable the mute control.
3. **People-sheet indicator** — muted peers show a small badge (nit from review).

## Test plan
- [x] Unit tests for mute store + notification skip / no-fingerprint delivery
- [ ] Mute a DM after handshake → message stores, no banner
- [ ] Mute a geohash channel → activity notification suppressed
- [ ] Pre-handshake DM mute control disabled
- [ ] Panic wipe clears mute preferences
- [ ] Localization coverage tests on CI

Closes #1594

### DIFF
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -17,6 +17,9 @@ struct MeshPeerRow: Identifiable, Equatable {
     /// Vouched-for by someone I verified, without an explicit verification of
     /// mine — rendered as the unfilled seal (verified gets the filled one).
     let showsVouchedBadge: Bool
+    /// This peer's direct conversation has notifications muted. Surfaced in the
+    /// row so muting is discoverable outside the open thread's bell (#1606).
+    let notificationsMuted: Bool
 
     var id: String { peerID.id }
 }
@@ -231,6 +234,16 @@ final class PeerListModel: ObservableObject {
         let meshRows = allPeers.map { peer in
             let isMe = peer.peerID == myPeerID
             let fingerprint = isMe ? nil : chatViewModel.getFingerprint(for: peer.peerID)
+            // Same resolver the mute toggle and the notification path use, so
+            // the row can never disagree with either (#1606).
+            let muteScope: ConversationNotificationMuteStore.Scope? = isMe
+                ? nil
+                : .direct(fingerprint: chatViewModel.stableConversationIdentity(for: peer.peerID))
+            // Hoisted out of the MeshPeerRow literal: `isMuted` carries a
+            // defaulted UserDefaults parameter, so it cannot be passed as a
+            // bare function value, and inlining it makes the initializer
+            // expression too complex for the type checker.
+            let notificationsMuted = muteScope.map { ConversationNotificationMuteStore.isMuted($0) } ?? false
             let isVerifiedFingerprint = fingerprint.map { peerIdentityStore.isVerified($0) } ?? false
             let verifiedBadge = !peer.isConnected && isVerifiedFingerprint
             // Vouched is subordinate to verified: never show both seals.
@@ -249,7 +262,8 @@ final class PeerListModel: ObservableObject {
                 isMutualFavorite: peer.isMutualFavorite,
                 encryptionStatus: chatViewModel.getEncryptionStatus(for: peer.peerID),
                 showsVerifiedBadgeWhenOffline: verifiedBadge,
-                showsVouchedBadge: vouchedBadge
+                showsVouchedBadge: vouchedBadge,
+                notificationsMuted: notificationsMuted
             )
         }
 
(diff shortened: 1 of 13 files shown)


Chessing234:

jackjackbits (changes requested):
The architecture is right — `ConversationNotificationMuteStore` injected as `conversationMutedProvider` mirrors the existing `hidePreviewsProvider` pattern, panic wipe resets it, and the single DM-notification call site is covered.

**Blockers:**
1. **Catalog rewrite + fake translations** (shared with your other Aug-2 PRs): the branch re-serializes the whole xcstrings via `add_localizable_key.py` and would drop main's 93 Aug-10 keys, breaking the coverage guard; the 4 new keys are English ×29. Rebase onto current main, add keys surgically, translate for real.
2. **Mute-key asymmetry.** The toggle keys the mute with `getFingerprint(for:) ?? peerIDToPublicKeyFingerprint[peerID]` (PrivateConversationModels) while the notification path resolves `getFingerprint(for:) ?? storedFingerprint(for:)` (coordinator). If the fallbacks disagree in availability, the mute is written as `dm:<fingerprint>` on one side and looked up as `dm:<peerid>` on the other — silently ineffective. The peerID fallback also evaporates across peer-ID rotation. Please unify both sides on a single fingerprint-resolution helper, and treat "no fingerprint yet" explicitly (either refuse the toggle pre-handshake or migrate the key on handshake).

**Nit:** nothing outside the toggled bell indicates a conversation is muted — a small indicator in the people sheet row would help.

Chessing234:
rebased onto main and redid the catalog surgically — +930/-0 now, your 93 aug-10 keys untouched, and the 5 keys have real translations across all 30 locales. dropped add_localizable_key.py too, it's what caused the rewrite.

on the key asymmetry: both sides now go through one stableConversationIdentity(for:), and i dropped the peer-ID fallback entirely rather than unifying on it — peer IDs rotate, so a mute keyed on one stops applying while the UI still shows it muted, which felt worse than not muting. scope is nil pre-handshake and the bell is disabled until a fingerprint exists. 11 checks over that, including the exact write-one-key/read-another case you described.

added the people-sheet indicator as well — bell.slash on rows whose DM is muted, same resolver.

no xcode on this machine so i couldn't run the xcode suite; left the device boxes unticked rather than claiming them. CI is green.

Chessing234:
working the catalog batch one pr at a time — next pass rebases this onto tip and adds the mute keys surgically (no full xcstrings rewrite, no english marked translated). will also align the mute-key resolution with the notification path.

Chessing234:
rebased onto tip — mute keys added surgically (no catalog rewrite; non-en marked new, not translated), fingerprint resolution unified via stableConversationIdentity, pre-handshake toggle disabled, people-sheet muted badge added.

Chessing234:
catalog is on current main; mute strings are real translations now (not english paste). fingerprint asymmetry is already unified through `stableConversationIdentity` on both the toggle and notification paths. people-sheet mute indicator still open if you want it.

Issue #1064, opened 2026-03-17
Security UX: Default View Should Not Be the Unencrypted Mesh Channel


ypreiser:
## Security UX: Default View Should Not Be the Unencrypted Mesh Channel

The whitepaper opens by describing BitChat as designed for *"secure, private, and censorship-resistant communication"* and dedicates significant depth to its encryption architecture. Yet the default landing view is `#mesh` — an unencrypted broadcast visible to every device within 7 hops. A new user who opens the app and starts typing is broadcasting in plaintext to dozens of strangers before they understand what they're using. The implementation undermines the stated security goals at the first interaction.

**Proposed change:** Land on a conversations list (P2P chats + encrypted channels) by default. Keep `#mesh` accessible but require a deliberate opt-in with a clear unencrypted/public warning.

This mirrors the principle Signal and WhatsApp both apply: encrypted is the path of least resistance, unencrypted is the explicit choice. Purely a UI/UX change — no protocol modifications required.

Related: #42, #370. Happy to contribute a PR if the direction is agreed upon.

ecgang:
Took a close look at HEAD (6653606) before proposing anything, and the picture is more nuanced than the issue states. Location channels are already restored at launch: `LocationStateManager` persists the selection under `locationChannel.selected` and `GeoChannelCoordinator` re-applies it on start. What still lands users in the public mesh timeline:

- first-ever launch (nothing persisted yet),
- users whose last activity was a DM — DMs aren't a restore target,
- and there's no visual cue that mesh is public and unencrypted.

`ConversationStore.swift:312` hard-defaults `activeChannel` to `.mesh`, and nothing persists the mesh-vs-DM distinction.

So a scoped version of option (b):

1. First-ever launch presents the conversation list instead of dropping straight into the mesh timeline.
2. Persist the last-active conversation in `ConversationStore` (same UserDefaults idiom as `locationChannel.selected`) and restore last-used DMs at launch, with validation — a stale peer falls back to the conversation list. Geohash restore stays where it lives today, so there's no second launch-time writer racing the existing path.
3. A light "public · unencrypted" label on the mesh header, consistent with #1357's direction.

`activeChannel` keeps its non-optional type; no coordinator or send-path signatures change.

Two questions:

1. The conversation list today is a sheet over the timeline, not a navigation destination — is presenting it over the (labeled) mesh timeline acceptable as "lands on the conversation list", or would you want a real list-detail restructure (a much bigger change)?
[...]

PR #1607, opened 2026-08-02
feat: mesh plaintext security guidance without changing default channel

Chessing234:
## Summary

A dismissible banner above the mesh composer saying that `#mesh` is an unencrypted local broadcast (#1064). `#mesh` stays the default channel — this is informed consent, not forced migration.

The body is bridge-aware: with the bridge off it names devices in radio range; with it on it also names Nostr relays in the area. Panic wipe resets the dismissal, so a wiped device gets the guidance again.

## Blockers from your review

**Catalog.** `scripts/add_localizable_key.py` is deleted — it was the whole-file re-serializer. Rebased onto current main and spliced in only the new keys, preserving Xcode's serialization: **+744/−0**, no pre-existing entry touched. Main's 520 keys, including the 93 added Aug 10, are byte-identical.

**Real translations.** All four keys are genuinely translated in all 30 locales at `state: "translated"` — informal register, lowercase, no English left anywhere. `security.mesh_plaintext.dismiss` reuses the exact per-locale wording already shipped for `location_notes.action.dismiss` rather than inventing a second word for the same action.

**The double warning on an empty timeline.** Picked the banner and deleted the empty-state hint, which also drops one key (5 → 4). The banner is the surface that survives because it's present whether or not the timeline has messages, whereas the hint only ever appeared when it was empty — so keeping the hint instead would have lost the warning in the case that matters more. `MeshEmptyStateView` is back to main's version, unmodified.

## Verification

- `MeshPlaintextNoticeSettingsTests` covers the dismissal round-trip and that `reset` posts the change notification, each against an isolated `UserDefaults` suite.
- The settings type is Foundation-only, so I compiled and ran it standalone under `-swift-version 5` — **7/7 assertions pass**, including suite isolation and that `setDismissed(false)` and `reset` are distinct. It also typechecks clean under `-swift-version 6`.
- Catalog re-parsed after the splice with every pre-existing entry compared for equality, and the `everyCodeReferencedKeyExistsInACatalog` logic re-run locally against both catalogs — 0 missing keys.

**Not run locally:** the Xcode suite. Command Line Tools only on this machine, so `xcodebuild` cannot build the app here — CI covers the build, app tests, simulator tests, SwiftLint and periphery. I have not seen the banner render, so its layout is unverified by eye.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -84281,6 +84281,750 @@
         }
       }
     },
+    "security.mesh_plaintext.body" : {
+      "comment" : "Body of the dismissible notice explaining mesh is unencrypted",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "أي شيء تكتبه في ‎#mesh يمكن لكل جهاز ضمن نطاق الراديو قراءته. لرسالة مباشرة مشفّرة، افتح شخصًا من قائمة الأشخاص."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "#mesh-এ আপনি যা লেখেন তা রেডিও পরিসরের প্রতিটি ডিভাইস পড়তে পারে। এনক্রিপ্টেড সরাসরি বার্তার জন্য, লোক তালিকা থেকে কাউকে খুলুন।"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "alles, was du in #mesh schreibst, kann von jedem gerät in funkreichweite gelesen werden. für eine verschlüsselte direktnachricht öffne jemanden aus der personenliste."
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "anything you type in #mesh can be read by every device within radio range. for an encrypted direct message, open someone from the people list."
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "cualquier cosa que escribas en #mesh puede leerla cualquier dispositivo dentro del alcance de radio. para un mensaje directo cifrado, abre a alguien desde la lista de personas."
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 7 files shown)


Chessing234:

jackjackbits (changes requested):
The feature itself is the best-designed of your Aug-2 batch — dismissible banner scoped to the mesh composer, bridge-aware body text, panic wipe restores it, copy avoids "user", sane tests with isolated UserDefaults suites. This is close to mergeable once the branch mechanics are fixed:

**Blockers (shared across the batch):**
1. **Catalog rewrite.** The branch re-serializes the entire `Localizable.xcstrings` (via `scripts/add_localizable_key.py` — Python `json.dump` spacing), and its catalog holds ~430 keys vs main's 520. Merging would drop the 93 keys added Aug 10 (#1654/#1656) and immediately fail `everyCodeReferencedKeyExistsInACatalog`. Please rebase onto current main and add only your new keys, surgically, preserving Xcode's canonical serialization.
2. **Real translations.** All 5 new keys are English pasted into the 29 non-en locales. Every key needs genuine translations — see the rest of the catalog for register conventions (informal, lowercase).

**Minor:** on an empty mesh timeline the banner and the empty-state hint show simultaneously — pick one. Permanent dismissal (no re-show cadence) is fine as a deliberate choice.

Happy to merge after the rebase + real translations.

Chessing234:
add_localizable_key.py is gone and the branch is rebased onto current main — catalog is +744/−0, four keys, your 93 aug-10 keys untouched. real translations in all 30 locales; the dismiss label reuses the per-locale wording already shipped for location_notes.action.dismiss so we don't ship two words for the same action.

on the double warning: kept the banner, deleted the empty-state hint. the banner is there whether or not the timeline has messages, so keeping the hint instead would have dropped the warning in the case that matters more. MeshEmptyStateView is back to main's version and that's one fewer key.

7/7 green. settings type ran standalone, 7/7 assertions, but no xcode here so i haven't seen the banner render.

Chessing234:
catalog rewrite is gone — tip is on current main with only the 4 new keys, and the non-en strings are real translations now (not english paste). minor empty-state overlap still there if you want that tightened before merge.

Chessing234:
hid the banner on an empty mesh so it doesn't sit on top of the empty-state hint.

Chessing234:
rebased onto main, catalog is surgical (+4 keys only, real translations). also hides the banner on an empty mesh timeline so it doesn't stack with the empty-state hint.

Issue #587, opened 2025-09-12
Support Universal Links


qalandarov:
Currently, the app is only supporting url scheme `bitchat://` which doesn't really work on a macOS app (or maybe it's not set up).
Supporting universal links will help with deeplinking as well (and potentially a web app or an app clip in the future)

<img width="372" height="454" alt="Image" src="https://github.com/user-attachments/assets/54e15a50-b03a-48e2-99c4-2d5bb7ed1495" />

SkinnnyJay:
Maybe I am off base, but would allowing universal links introduce a way to phish/track you? vs. creating some anonymous proxy that could mitigate this?

qalandarov:
I think it's actually the other way around, if your app uses a url scheme (e.g. `bitchat://`) then other apps could just check if this app is installed on the device by a simple "can open url" and then passing your url scheme whereas checking if can open url with an http link would always return true so they wouldn't know if the app is actually installed or not besides [other benefits](https://medium.com/@saurabhmishra2290/mastering-universal-links-in-ios-enhancing-security-and-user-experience-1101384e951c)

tommyming:
As I rmb, universal links (aka normal urls e.g. `https://bitchat.com`) are something relatively safer than deeplinks (aka custom URL schemes e.g. `bitchat://`).

tommyming:
btw can help on this if it is not done yet! 
shouldn't be that hard as long as there are already seomthing done on deeplink.

jackjackbits:
Still open and unimplemented — no associated-domains entitlement or applinks handling on main; only the `bitchat://` scheme is wired up. The code side is small, but this needs a decision on a canonical domain plus AASA hosting, and for this project a hosted domain is also a blockable, observable dependency — which cuts against the censorship-resilience goals that now shape the roadmap. Keeping as backlog pending that product decision; the privacy point about canOpenURL probing of custom schemes is well taken.

PR #1608, opened 2026-08-02
feat: copy message deep links for mesh, geohash, and DMs

Chessing234:
## Summary

Copies a `bitchat://` deep link to a specific message from the context menu, and opens those links back to the message (#587).

Reworked after review — all four blockers:

### `mid` is no longer decorative

Opening a link now reveals the message it names. The ID is held in `pendingDeepLinkMessageID` until the destination conversation has rendered, the render window is grown when the target sits behind it, and the row is addressed as `"<contextKey>|<messageID>"` — matching how `MessageDisplayItem.id` is composed.

The window growth is the part that made this real: an unrendered row is not in the view hierarchy, so `scrollTo` would have silently done nothing for anything older than the current window. An unknown `mid` is dropped, since the message may never have reached this device.

### The `dm` path is validated

`MessageDeepLink.directConversationTarget(fromPath:)` shape-checks the path before it reaches conversation state. Only the shapes this file emits links for are accepted: a bare 16-hex mesh peer, a `nostr_` geo DM, or a `group_` conversation. Routing-only prefixes (`noise:`, `name:`, `mesh:`, `bridge:`), geohash *chat* IDs, non-hex and over-length input are all rejected. This mirrors the charset/length gate the geohash handler already applied.

### Location precision

Copying a geohash link is gated behind the same confirmation #1513 put in front of invite sharing, reusing `ChannelShare.shouldWarn(forGeohash:)` so the two thresholds cannot drift apart. Mesh and DM links carry no location and copy without a prompt.

### Catalog

Rebased onto current main with surgical key additions: **+930/−0**, five keys, main's 93 Aug-10 keys untouched, no pre-existing entry modified. Real translations in all 30 locales at `state: "translated"`, `%@` specifiers verified preserved in each. `scripts/add_localizable_key.py` is dropped — it produced the rewrite.

## Known limitation: DM links embed a routing peer ID

You were right that this is a durability and disclosure problem. `bitchat://dm/<peer-id>` addresses the conversation by its current routing peer ID, which rotates — so a copied DM link goes stale, and it discloses the live routing ID at copy time.

Fingerprint addressing would fix both, but the fingerprint is not always known at copy time (it needs a completed handshake) and resolving one back to a conversation needs a lookup that does not exist yet. Documented in `docs/MESSAGE-DEEP-LINKS.md` and left as follow-up rather than shipped half-done. Happy to take it as a separate PR if you'd like it before this lands.

## Verification

- `MessageDeepLinkTests` grew from URL construction only to 15 cases covering the open path: accepted shapes, **12 hostile/malformed `dm` paths**, percent-decoding, `mid` round-trip and absence, row-identity composition, and window growth.
- The deep-link logic was compiled and executed against the **real `BitFoundation` module** (not a stub) — 26 assertions pass — and typechecked under `-swift-version 5` and `-swift-version 6`.
- Catalog integrity and translation coverage checked programmatically; the repo's `everyCodeReferencedKeyExistsInACatalog` logic re-run locally and passes.

**Not run locally:** the Xcode suite — Command Line Tools only on this machine, so `xcodebuild` cannot build the app here. CI covers the build, app tests, iOS simulator tests, SwiftLint and periphery.

I have not opened a link on a device, so the scroll itself is unexercised by hand; the addressing and window arithmetic it depends on are covered by tests.

Refs #587

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -13579,6 +13579,936 @@
         }
       }
     },
+    "message.link.copy" : {
+      "comment" : "Context-menu action that copies a bitchat:// deep link to this message",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "نسخ رابط الرسالة"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "বার্তার লিঙ্ক কপি করো"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "nachrichtenlink kopieren"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "copy message link"
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "copiar enlace del mensaje"
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "کپی پیوند پیام"
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "kopyahin ang link ng mensahe"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "copier le lien du message"
+          }
+        },
+        "he" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "העתקת קישור להודעה"
+          }
+        },
+        "hi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "संदेश लिंक कॉपी करें"
+          }
+        },
+        "id" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "salin tautan pesan"
+          }
+        },
+        "it" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "copia link del messaggio"
+          }
+        },
+        "ja" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "メッセージリンクをコピー"
+          }
+        },
+        "ko" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "메시지 링크 복사"
+          }
+        },
+        "ms" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "salin pautan mesej"
+          }
+        },
+        "ne" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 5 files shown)


Chessing234:
CI red was the known unrelated `SimulatedMeshTests.publicMessageRelaysAcrossLineTopologyWithinTTLBudget` flake — empty-commit retrigger pushed.

Chessing234:

jackjackbits (changes requested):
The mechanism is legitimate — the `bitchat://` scheme is registered and a geohash open-URL handler already existed. But the PR isn't ready:

**Blockers:**
1. **Catalog rewrite + translations** (shared with the batch): whole-file xcstrings re-serialization would clobber main's Aug-10 catalog and break the coverage guard; both new keys are English ×29. Rebase, surgical adds, real translations.
2. **`mid` is never handled on open.** The headline feature — locating a message — is deferred to "Future work" in your own doc; as shipped, these are channel links with a decorative query param. Either implement scroll-to-message on open or retitle/rescope honestly as "copy channel links".
3. **Unvalidated `bitchat://dm/` input.** The new `case "dm"` passes an arbitrary string from any app/webpage straight into `privateConversationModel.openConversation(for:)` — the geohash handler validates charset/length; the dm handler validates nothing. That's a phantom-conversation vector. Validate the ID shape before acting on it.
4. **Location privacy.** Copying a geohash message link puts your geohash (up to full precision) on the clipboard with no warning — house precedent from #1513 invite sharing runs `showSharePrecisionWarning` first. Same gate needed here. Also note DM links embed the current routing peerID, which rotates — links go stale and leak the live routing ID; fingerprint-based addressing would be more durable.

Tests currently cover URL construction only — the open-URL handlers need coverage too, especially the validation added for (3).

Chessing234:
reworked all four.

mid actually works now — the message id is held until the destination conversation has rendered, then the render window is grown if the target sits behind it and the row is addressed the same way MessageDisplayItem builds its id. the window bit is what made it real: an unrendered row isn't in the hierarchy so scrollTo was a no-op for anything older than the window. unknown mid is dropped.

dm paths are shape-checked before they reach conversation state — bare 16-hex mesh peer, nostr_ geo dm, or group_ only. routing-only prefixes, geohash chat ids, non-hex and over-length all rejected. 12 hostile paths in the tests.

geohash link copying goes through the same precision confirmation as #1513, reusing ChannelShare.shouldWarn so the thresholds can't drift apart. mesh and dm links carry no location so they copy straight.

on the rotating peerID: you're right, and i didn't want to half-fix it. fingerprint addressing needs a completed handshake at copy time and a reverse lookup that doesn't exist yet, so i documented the limitation and left it as follow-up. happy to do it as a separate PR if you'd rather it land first.

catalog is +930/-0 with real translations in all 30 locales. no xcode here so i couldn't run the xcode suite or open a link on a device; the addressing and window arithmetic are covered by tests. CI green.

Chessing234:
catalog rewrite is cleared on current main (5 new keys, translated). the mid-handler / “locate message” gap from your review is still the open product question on this one.

Issue #1518, opened 2026-07-27
Images/audio sent from Android not received on iOS


Hermes34-dev:
When sending images or voice messages from Android to iOS in a chat, the iOS device does not receive them. iOS→Android media transfer works fine. Text messages work fine in both directions.

Steps to reproduce:
1. Send an image or audio message from an Android device to an iOS device
2. iOS device does not receive the media (no message appears). The message appears as sent in the Android device.

Environment:
Receiving device: iPhone 13 Pro (iOS 18.5)
Sending devices: tested with 2 different Android devices (Google Pixel 10, Xiaomi Pad 7 Pro - both Android 16)
App version: fill in

Additional notes:
- Confirmed with 2 different Android devices as senders, so likely not specific to one Android device. 
- iOS→Android media transfer works correctly. Android->Android transfer also worked correctly.
- Have not tested iOS→iOS

sergava:
I have same issue with photos in private chats. Audio not tested. 
Android 16 and ios 26.5.2. 
Android app version 1.7.4.
Ios app version 1.7.0.

Chessing234:
triage note: android sends `image/jpeg` / `audio/mp4` which ios already accepts — so this usually isn't a mime allow-list miss. your stack (android 1.7.4 + ios app store 1.7.0) is a real skew: several private-media fixes landed on main *after* v1.7.0 (encrypt-before-fragment, durable receipts, reconnect retry, …). try an ios build from current main / next store cut before digging into codecs. wrote up the matrix in https://github.com/permissionlesstech/bitchat/pull/1581

Hermes34-dev:
I found out this issue is fixed when on flight mode for some reason. When on flight mode with bluetooth turned on, everything is sent and received perfectly between the two devices. Thank you for the dev's colaboration in this :) I am awaiting the app's new update

vinayak509143:
Investigating the cross-platform media reception issue between Android and iOS. Working on reconciling the decompressed payload boundary and adding structured diagnostic logging for dropped frames. Will submit a PR with reproduction tests.

PR #1609, opened 2026-08-02
feat: surface private media decode failures in the DM thread

Chessing234:
## Summary

Surfaces authenticated private-media decode failures in the affected DM instead of dropping them silently (#1518), with a structured reason code in the log. Public mesh transfers keep the existing security-logging path.

Reworked after review:

- **Rebased onto current main, catalog edited surgically.** No re-serialization: +930/−0, five added keys, main's 93 Aug-10 keys untouched, no pre-existing entry modified. `scripts/add_localizable_key.py` is dropped.
- **Real translations** in all 30 locales at `state: "translated"`, with `%@`/`%lld` specifiers verified preserved in every one.
- **No repo path in user-facing copy.** `media.decode.failure.malformed` no longer points at `docs/ANDROID-IOS-MEDIA-INTEROP.md`; it now says the sender may be on an app version that packs media differently.

## Rate limiting

Decode failures are attacker-triggerable — an authenticated peer can send malformed media as fast as the link allows — so `PrivateMediaDecodeFailureThrottle` posts at most one line per peer per five-minute window.

Suppressed failures are counted, not dropped: the next line that surfaces reports how many were withheld. Peers are throttled independently, so a flooder cannot silence a genuine failure from someone else. Closed windows are pruned, so a sender rotating peer IDs cannot grow the throttle's state without bound.

## Why a blocked sender returns `true`

The return value distinguishes "consumed, deliberately silent" from "could not handle this". A blocked sender's payload is not a failure to handle — it is a drop we chose, and blocking should produce no local UI. Marking it `false` would file a deliberate policy drop alongside genuine decode errors.

To be precise about scope: this is currently a statement of intent, not a behavior change. `handlePrivatePayload` is `@discardableResult` and its only production caller — the `.privateFile` branch of `deliverNoisePayload` in `BLEService` — ignores the result. Today the difference is observable only to `BLEFileTransferHandlerTests`. It would matter if a caller ever branches on it.

## Verification

Ran locally:

- Throttle logic compiled and executed standalone — 9 checks, including 500 malformed payloads in 50 seconds producing exactly one line, suppressed counts surfacing on the next line, peer independence, reason pass-through, and 1,000 rotating peer IDs not accumulating state.
- Typechecked against the real `BitFoundation` module (not a stub) under `-swift-version 5`, `-swift-version 6` and `-strict-concurrency=complete`.
- The repo's `everyCodeReferencedKeyExistsInACatalog` logic re-implemented and run — passes. Catalog integrity and translation coverage checked programmatically.

**Not run locally:** the Xcode suite — Command Line Tools only on this machine, so `xcodebuild` cannot build the app here. CI covers the build, app tests, iOS simulator tests, SwiftLint and periphery.

I have not reproduced a malformed payload on a device, so those boxes stay unticked.

Refs #1518

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -13579,6 +13579,936 @@
         }
       }
     },
+    "media.decode.failure.magic_mismatch" : {
+      "comment" : "System message when file magic bytes mismatch MIME; %@ is the MIME",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "تعذّر فتح ملف الوسائط — المحتوى لا يطابق النوع المعلن (%@)."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "মিডিয়া ফাইল খোলা যায়নি — বিষয়বস্তু ঘোষিত ধরনের সাথে মেলে না (%@)।"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "medien-datei ließ sich nicht öffnen — inhalt passt nicht zum angegebenen typ (%@)."
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "couldn't open the media file — contents don't match the declared type (%@)."
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "no se pudo abrir el archivo multimedia — el contenido no coincide con el tipo declarado (%@)."
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "فایل رسانه باز نشد — محتوا با نوع اعلام‌شده هم‌خوانی ندارد (%@)."
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "hindi mabuksan ang media file — hindi tugma ang nilalaman sa idineklarang uri (%@)."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "impossible d'ouvrir le fichier — le contenu ne correspond pas au type déclaré (%@)."
+          }
+        },
+        "he" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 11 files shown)


Chessing234:

jackjackbits (changes requested):
Surfacing decode failures in the thread is a good idea and the delegate + env-closure design is sound, with a tested blocked-peer gate.

**Blockers:**
1. **Catalog rewrite + translations** (shared with the batch): the branch re-serializes the entire xcstrings and would clobber main's Aug-10 catalog (93 keys), breaking `everyCodeReferencedKeyExistsInACatalog`; the 4 new keys are English ×29 (`needs_review` doesn't change that). Rebase onto current main, surgical key adds, real translations.
2. **UI copy points at a repo file.** `media.decode.failure.malformed` tells people to "see docs/ANDROID-IOS-MEDIA-INTEROP.md" — that file doesn't exist in the tree, and a repository path has no place in user-facing copy regardless. Describe the problem in plain words.
3. **Unthrottled DM spam vector.** Every malformed private media payload from an authenticated peer posts a system line via `addLocalPrivateSystemMessage` with no rate limit or dedup — a hostile peer can flood a DM thread with system messages. Debounce/coalesce per peer (e.g. one line per N minutes, or "N media items failed to decode").

**Please also explain** the behavior change where a blocked-peer decode failure now returns `true` ("handled") from `handlePrivatePayload` — previously all failures returned `false`. It's plausible but it's unexplained in the PR body.

Chessing234:
rebased and redid the catalog the same way — +930/-0, your aug-10 keys intact, real translations in all 30 locales with the %@/%lld specifiers checked in each one.

dropped the docs path from the malformed copy — it now just says the sender may be on a version that packs media differently.

for the flood: added PrivateMediaDecodeFailureThrottle, one line per peer per 5 min. suppressed failures are counted and reported on the next line rather than dropped, peers are throttled independently so a flooder can't silence someone else, and closed windows get pruned so rotating peer IDs can't grow the state. 9 checks including 500 malformed payloads in 50s producing exactly one line.

on the blocked-peer return: it's meant to say "consumed, deliberately silent" rather than "couldn't handle it" — blocking is a drop we chose and shouldn't produce local UI. worth being precise though: handlePrivatePayload is @discardableResult and the only production caller ignores it, so today that's intent rather than a behaviour change and only the tests observe it. written up in the PR body.

no xcode here so the xcode suite went unrun; device boxes left unticked. CI is green.

Chessing234:
blocked-peer path returns true because blocking is a deliberate drop, not a decode failure — the handler reports the payload as handled so upstream retry logic doesn't wedge. no system line is posted for blocked senders.

PR #1610, opened 2026-08-02
feat: export public identity verification card

Chessing234:
## Summary

Copies a public identity card — nickname, noise fingerprint, npub — from the verification sheet as plain text. Public fields only; no key material (ref #183).

## Blockers from your review

**`scripts/add_localizable_key.py` is gone.** You were right that it was the root cause across the batch — it round-tripped the catalog through Python's `json.dump` and rewrote all ~107k lines, which is why these PRs conflicted with main and with each other. It is deleted here and not used anywhere.

**Catalog.** Rebased onto current main and spliced in only the five keys, preserving Xcode's serialization: **+930/−0**, no pre-existing entry touched. Main's 520 keys, including the 93 added Aug 10, are byte-identical.

**Translations.** Real translations in all 30 locales at `state: "translated"`, `%@` specifiers verified preserved in each. `identity.card.npub` is the one exception worth naming: it is the bare term `npub` plus a placeholder, so the Latin-script locales are legitimately identical to English and only the CJK locales differ (full-width colon). That is the correct translation rather than an untranslated paste, but I would rather point at it than have you find it.

**Doc/code mismatch — dropped both.** I went to emit the URL and found a reason not to. `VerificationService.verifyScannedQR` rejects any payload older than `TransportConfig.verificationQRMaxAgeSeconds`, which is 5 minutes. A signed URL on a card people copy, keep and forward would stop verifying almost immediately, while reading as the most authoritative line on the card — worse than not being there.

So `identity.card.verify_url` is dropped, the doc explains the reasoning instead of promising the line, and a test asserts the card contains no `bitchat://`.

That also speaks to your "marginal next to the QR" point, and I think it is the honest answer rather than a defence: the QR is the better tool whenever both people are at a screen. The card is for when they are not — the fingerprint and npub are the two fields that do not expire.

## Verification

- `IdentityVerificationCardTests` grew from 1 test to 4: public-fields-only, the URL omission and its reason, npub absent when nil, and empty npub treated as absent.
- The card is Foundation-only by design, so I compiled and ran it standalone under `-swift-version 5` — **15/15 assertions pass**, including line counts for both card shapes and `current` matching the memberwise init. It also typechecks clean under `-swift-version 6`.
- Catalog re-parsed after the splice with every pre-existing entry compared for equality, and the `everyCodeReferencedKeyExistsInACatalog` logic re-run locally against both catalogs — 0 missing keys.

**Not run locally:** the Xcode suite. Command Line Tools only on this machine, so `xcodebuild` cannot build the app here — CI covers the build, app tests, simulator tests, SwiftLint and periphery. I have not tapped the button on a device; the sheet wiring is a small diff over `VerificationSheetView` and is only as verified as CI makes it.

### DIFF
--- a/bitchat/App/VerificationModel.swift
+++ b/bitchat/App/VerificationModel.swift
@@ -63,6 +63,14 @@ final class VerificationModel: ObservableObject {
         return VerificationService.shared.buildMyQRString(nickname: currentNickname, npub: npub) ?? ""
     }
 
+    func identityVerificationCard() -> IdentityVerificationCard {
+        IdentityVerificationCard.current(
+            nickname: currentNickname,
+            npub: try? chatViewModel.idBridge.getCurrentNostrIdentity()?.npub,
+            noiseFingerprint: chatViewModel.getMyFingerprint()
+        )
+    }
+
     func verifyScannedPayload(_ payload: String) -> VerificationScanOutcome {
         guard let qr = VerificationService.shared.verifyScannedQR(payload) else {
             return .invalid
--- a/bitchat/Identity/IdentityVerificationCard.swift
+++ b/bitchat/Identity/IdentityVerificationCard.swift
@@ -1,0 +1,71 @@
+//
+// IdentityVerificationCard.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Foundation
+
+/// Public-only identity card for out-of-band verification (#183).
+///
+/// Contains nickname, npub, and Noise fingerprint — never private key material.
+/// Passphrase-sealed backup export remains future work; this card is the safe
+/// first step for sharing who you are on bitchat.
+struct IdentityVerificationCard: Equatable {
+    let nickname: String
+    let npub: String?
+    let noiseFingerprint: String
+
+    var plainText: String {
+        var lines: [String] = [
+            String(
+                localized: "identity.card.title",
+                defaultValue: "bitchat identity card",
+                comment: "Title line when exporting a public identity verification card"
+            ),
+            "",
+            String(
+                format: String(
+                    localized: "identity.card.nickname",
+                    defaultValue: "nickname: %@",
+                    comment: "Nickname line on the public identity card; %@ is the nickname"
+                ),
+                locale: .current,
+                nickname
+            ),
+            String(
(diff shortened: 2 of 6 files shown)


Chessing234:

jackjackbits (changes requested):
The card itself is correctly scoped — nickname + noise fingerprint + npub, no key material — and the feature is harmless (ref #183), though marginal next to the signed QR the sheet already shows.

**Blockers:**
1. **Remove `scripts/add_localizable_key.py`.** This PR carries the script that's poisoning your whole Aug-2 batch: it round-trips the catalog through Python `json.dump`, rewriting all ~107k lines. That fights Xcode's canonical serialization and our translation workflow, and it's why every PR in the batch conflicts with main and with each other. Keys should be added surgically in Xcode's format.
2. **Catalog rewrite + translations** (consequence of the above): rebase onto current main — your catalog is missing the 93 Aug-10 keys, so merging breaks the coverage guard — and give the 6 new keys real translations (currently English ×29).
3. **Doc/code mismatch.** `docs/IDENTITY-VERIFICATION-CARD.md` promises a "Signed verification QR URL" line and the PR adds `identity.card.verify_url`, but `plainText` never emits it — that's a dead catalog key and an overstated doc. Either emit it or drop both.

Chessing234:
add_localizable_key.py is deleted — you were right that it was the root cause across the batch. rebased onto current main and spliced in only the 5 keys in xcode's format, +930/−0, your 93 aug-10 keys untouched. real translations in all 30 locales.

on the doc mismatch: i went to emit the url and found a reason not to. verifyScannedQR rejects anything older than verificationQRMaxAgeSeconds, which is 5 minutes — so a signed url on a card people copy and forward stops verifying almost immediately while still reading as the most authoritative line on it. dropped identity.card.verify_url, doc says why now, test asserts no bitchat:// on the card.

that's also my honest answer to "marginal next to the qr" — the qr is better whenever both people are at a screen. the card is for when they aren't, and the fingerprint and npub are the two fields that don't expire.

one to flag rather than have you find: identity.card.npub is the bare term plus a placeholder, so the latin-script locales are legitimately identical to english and only the cjk ones differ. noted in the body.

7/7 green. the card is foundation-only so i compiled and ran it standalone, 15/15, but no xcode here so i haven't tapped the button.

PR #1612, opened 2026-08-02
feat: export chat as plaintext with sealing warning

Chessing234:
## Summary
- Add `ChatPlaintextExporter` to format mesh, geohash, and DM timelines as plain text with an explicit unsealed-plaintext warning header.
- Expose an export control on non-empty timelines; confirm before writing plaintext, then open the system share sheet.
- Localize export actions, confirmation copy, scope titles, and accessibility labels (en + needs_review for other locales).

## Test plan
- [ ] Export a mesh conversation and verify the warning line appears at the top
- [ ] Export a DM and confirm scope title uses the peer nickname
- [ ] Cancel the confirmation dialog and confirm no file is shared
- [ ] Run `ChatPlaintextExporterTests`

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -3907,6 +3907,1866 @@
         }
       }
     },
+    "chat.export.accessibility" : {
+      "comment" : "Accessibility label for the export-chat control",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "export chat as plaintext"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "export chat as plaintext"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "export chat as plaintext"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "export chat as plaintext"
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "export chat as plaintext"
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "export chat as plaintext"
+          }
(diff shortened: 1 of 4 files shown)


Chessing234:
Addressed Codex P1 on tip: `@testable import bitchat` in the exporter tests, plus `defaultValue:` so catalog resolution in the test bundle still matches expectations.

jackjackbits (changes requested):
The exporter code is clean (POSIX-locale timestamps, honest confirmation copy) — but this one needs a product/opsec decision before any rework, so please hold on code changes until that's settled:

**The core question:** exporting full sealed-DM history as unsealed plaintext through `ShareActivityView` sends content to arbitrary share targets — cloud drives, clipboard managers, anything with a share extension. For an app whose threat model includes device seizure and protest use, that's a deliberate opening of the sealed-store boundary. If it survives review at all, it likely needs to be settings-gated and/or DM-excluded. We'll make that call and follow up here.

**If it goes ahead, the rework asks are:**
1. Catalog rewrite + translations (shared with the batch): rebase onto current main without the whole-file xcstrings re-serialization (drops main's 93 Aug-10 keys, breaks the coverage guard); 10 new keys need real translations, not English ×29.
2. The permanently visible floating export capsule on every non-empty conversation is heavy chrome for a rare action — move it into a menu/context action.
3. It shares a raw string while the header says "plain text file" — share an actual file (and note a string lands in more leak-prone targets).

Chessing234:
rebased onto main and rebuilt the catalog hunk surgically — it was reformatting the whole file (+55628/-53169) to add 10 keys. now +1860/-0, no existing entry touched, no key removed. whole pr is +2049/-0 and merges clean again.

suite green locally, 2018 → 2021, the delta is the three exporter tests.

Chessing234:
swiftlint was a real failure — trailing comma in the export header literal, fixed in 580a4c8.

the app-test failure is a known main-branch flake, not this diff: `timeoutRestoredSessionDefersQueueDrainUntilConvergence` at BLEServiceCoreTests.swift:1017, same test and same line, failed on main in runs 31418625613 and 31360294420. nothing here touches BLE session state.

Chessing234:
holding on code changes until the product/opsec call is settled.

PR #1613, opened 2026-08-02
test: fail CI when String(localized:) keys are missing from xcstrings

Chessing234:
## Summary
- Add `LocalizationKeyScanner` to walk production Swift and collect every `String(localized:)` key (including `defaultValue:` call sites).
- Extend `LocalizationCoverageTests` so CI fails when a referenced key is missing from the main or share-extension catalog.
- Skip `_PreviewHelpers/` (preview-only keys should not force catalog rows).
- Seed the 17 board / delivery / location keys the scan found missing — without reformatting the whole `.xcstrings` file.

## Why the comments
The scanner and coverage tests now spell out the two failure modes (missing locale vs missing key), what the line parser does and does not handle, and why main vs share catalogs are checked separately.

## Test plan
- [ ] Run `LocalizationCoverageTests` and `LocalizationKeyScannerTests`
- [ ] Confirm an intentionally missing catalog key fails the new test
- [ ] Confirm PR diff for `Localizable.xcstrings` is key insertions only (no full-file reformat)


Chessing234:
Addressed Codex P1s on tip: Foundation import in the scanner tests, and the 17 missing catalog keys the coverage gate found are now seeded (en + needs_review).

Chessing234:
codex notes are on tip — Foundation import is in the scanner tests, and the scanned missing keys were added without reformatting the whole xcstrings file.

jackjackbits:
Closing as superseded: #1654 (merged Aug 10) added `everyCodeReferencedKeyExistsInACatalog` to LocalizationCoverageTests, which does exactly this — scans source for literal `String(localized:)` keys and fails CI on any key absent from both catalogs. The 17 keys your scanner found are the same ones #1654 found, and #1655/#1656 gave them real translations across all 30 locales.

Two ideas from this PR are genuinely better than what we have and may get folded into our test: per-catalog attribution (share-extension keys must live in the share catalog specifically, not just the union) and file:line output on failure. Appreciate the work — this was the right guard to want.

PR #1614, opened 2026-08-02
feat: clear unread badge for a single conversation

Chessing234:
## Summary

Mark a conversation as read without opening it. Adds the action to the mesh, groups and chats rows in the people sheet, and to the header of an open DM.

No read receipt is sent. The unread badge is local state, so dismissing it stays local — that's the whole point of the feature, and it's why this is not just "open the thread for me".

## Alias-aware clearing

`ChatUnreadStateResolver.matchingUnreadPeerIDs` collects every peer ID that resolves to the same conversation badge — the routing peer ID, the noise-key ID, the nostr ID, and geoDM threads matched by nickname — mirroring what `hasUnreadMessages` already does when it decides to *show* the badge.

Clearing only the routing peer ID left the badge lit on the aliases, which is the ghost-badge bug. Show and clear now read the same set.

## Blockers from your review

**Catalog.** Rebased onto current main and spliced in only the three new keys, preserving Xcode's serialization: **+558/−0**, no pre-existing entry touched. Main's 520 keys — including the 93 added Aug 10 — are byte-identical. This branch never carried `add_localizable_key.py`.

**Translations.** All three keys are genuinely translated in all 30 locales at `state: "translated"`, lowercase to match the `mesh_peers.*` / `chats.*` namespace. No English pasted into a non-English locale.

**The chats section (#1598).** Rebasing brought it in, and it now carries the same action — `RecentChatList` gets `onClearUnread`, in both the context menu and the VoiceOver rotor. That section is exactly where a stale badge is most likely to sit, since those rows are the ones with no roster entry above them.

## Verification

- `ChatUnreadStateResolverTests` covers `matchingUnreadPeerIDs` against the noise-key, nostr and nickname-matched geoDM aliases; `ClearUnreadConversationTests` covers the clearing path end to end.
- Catalog integrity checked programmatically: re-parsed after the splice and every pre-existing entry compared for equality, plus the `everyCodeReferencedKeyExistsInACatalog` logic re-run locally against both catalogs — 0 missing keys.

**Not run locally:** the Xcode suite. This machine has Command Line Tools only, so `xcodebuild` cannot build or test the app here — CI covers the build, app tests, simulator tests, SwiftLint and periphery. I have not exercised the menus by hand, so the wiring is only as good as the tests plus CI.

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -67,6 +67,12 @@ final class ConversationUIModel: ObservableObject {
         chatViewModel.sendMessage("/clear")
     }
 
+    /// Dismisses the unread indicator for a conversation the user has not
+    /// opened; does not emit read receipts.
+    func clearUnread(for peerID: PeerID) {
+        chatViewModel.markPrivateChatRead(peerID)
+    }
+
     func sendHug(to sender: String) {
         chatViewModel.sendMessage("/hug @\(sender)")
     }
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -106,6 +106,12 @@ final class PeerListModel: ObservableObject {
         chatViewModel.startPrivateChat(with: peerID)
     }
 
+    /// Clears the unread badge for one direct or group conversation without
+    /// opening it or sending read receipts.
+    func clearUnread(for peerID: PeerID) {
+        chatViewModel.markPrivateChatRead(peerID)
+    }
+
     func toggleFavorite(peerID: PeerID) {
         chatViewModel.toggleFavorite(peerID: peerID)
     }
--- a/bitchat/App/PrivateConversationModels.swift
+++ b/bitchat/App/PrivateConversationModels.swift
@@ -183,6 +183,13 @@ final class PrivateConversationModel: ObservableObject {
         chatViewModel.markPrivateMessagesAsRead(from: peerID)
     }
 
+    /// Clears the unread badge for the selected conversation without sending
+    /// read receipts — for when new messages arrive while the thread is open.
+    func clearUnreadForSelectedConversation() {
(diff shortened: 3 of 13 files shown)


Chessing234:

jackjackbits (changes requested):
This is the best feature of your Aug-2 batch and I want it: "mark as read" without read receipts is the privacy-correct call, and the alias-aware clearing buried inside (`ChatUnreadStateResolver.matchingUnreadPeerIDs` clearing noise-key/nostr/geoDM-nickname aliases together, mirroring `hasUnreadMessages`) fixes real badge ghosts. Tests are solid.

The blockers are purely mechanical, shared with the batch:
1. **Catalog rewrite.** The branch re-serializes the entire xcstrings from an Aug-1 base (~430 keys vs main's 520) — merging drops the 93 keys added Aug 10 and fails `everyCodeReferencedKeyExistsInACatalog`. Rebase onto current main and add only your 3 keys, surgically, preserving Xcode's serialization (no `add_localizable_key.py`).
2. **Real translations** for those 3 keys — currently English ×29.

Heads-up for the rebase: main's people sheet gained a recent-chats section (#1598) since your base, so expect small conflicts in the sheet views — the context-menu action would ideally cover those rows too.

Fix the mechanics and this is a merge.

Chessing234:
mechanics fixed. rebased onto current main and spliced in only the 3 keys in xcode's serialization — +558/−0, your 93 aug-10 keys untouched. real translations in all 30 locales, lowercase to match the mesh_peers/chats namespace.

thanks for the #1598 heads-up — the chats section now gets the same action, context menu and voiceover rotor. that's the section where a stale badge is most likely to sit since those rows have nothing above them.

7/7 green. no xcode on this machine so i haven't driven the menus by hand.

Chessing234:
catalog rewrite is cleared — tip sits on current main with only the 3 new keys and real non-en strings. ready for another look on the mark-as-read feature itself.

PR #1615, opened 2026-08-02
feat: search messages within the current thread

Chessing234:
## Summary

Filters the open conversation by message body or sender nickname.

## The render problem — you were right

`.searchable` is gone. Checking the hierarchy rather than guessing:

- `ContentView.swift` builds `MessageListView` directly at line 539. The only occurrence of `NavigationStack` in that file is inside a **comment** (line 303) — there is no navigation ancestor, so `.searchable` rendered nothing in the public timeline.
- `ContentSheetViews.swift` opens a `NavigationStack` at line 127 and presents `ContentPrivateChatSheetView` inside it at 131/149, so the DM sheet was the only place the field could ever have appeared.

So the feature was dead exactly where most conversation happens. Replaced with a plain search field, which has no navigation dependency and renders identically in both placements — the failure mode is eliminated by construction rather than by hoping.

The field lives inside `MessageListView`, so neither header needed changing and one implementation serves both call sites. It collapses to a magnifier until used, and resets on conversation switch so a stale filter never applies to a different thread.

## Result cap

Search bypasses the message window, so a common term in a long thread rendered every historical match in one pass. Results are now capped at 200, keeping the newest, with a line stating how many matches were withheld — a truncated list says so rather than passing itself off as complete.

## Accessibility

The label moved onto the search field itself. It had been applied to the whole `ScrollView` chain, which relabelled every message in the list for VoiceOver.

## Catalog

Rebased onto current main with surgical additions: **+930/−0**, five keys, main's 93 Aug-10 keys untouched, no pre-existing entry modified. Real translations in all 30 locales at `state: "translated"`, `%lld` specifiers verified preserved in each.

## Verification

- `ThreadMessageSearchTests` covers the cap: capped at the limit, newest kept, under-cap untouched, non-positive limit uncapped, truncation flag boundaries — plus the original filter behaviour, unchanged.
- The filter and cap were compiled and executed standalone — 11 assertions pass, including that blank queries, case-insensitive body matches, sender matches and blank-body exclusion still behave as before.

**Not run locally, and worth being blunt about:** I could not produce the screenshots you asked for. This machine has Command Line Tools only, no Xcode, so I cannot build or run the app to photograph the field in either placement.

What I can offer instead is the static evidence above for *why* `.searchable` was dead, and the fact that a plain `TextField` in the view hierarchy has no ancestor requirement to satisfy. If you want the screenshots before this lands, say so and I will get to a machine with Xcode rather than guess.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -13579,6 +13579,936 @@
         }
       }
     },
+    "thread.search.accessibility.label" : {
+      "comment" : "Accessibility label for in-thread message search",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "ابحث في رسائل هذه المحادثة"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "এই কথোপকথনের বার্তা খোঁজো"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "nachrichten in dieser unterhaltung durchsuchen"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "search messages in this conversation"
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "buscar mensajes en esta conversación"
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "جستجوی پیام‌های این گفتگو"
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "maghanap ng mensahe sa usapang ito"
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "rechercher des messages dans cette conversation"
+          }
+        },
+        "he" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "חיפוש הודעות בשיחה הזו"
+          }
+        },
+        "hi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "इस बातचीत के संदेश खोजें"
+          }
+        },
+        "id" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "cari pesan di percakapan ini"
+          }
+        },
+        "it" : {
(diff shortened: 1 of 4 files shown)


Chessing234:
Addressed Codex P2 on tip: thread-search strings now cover all catalog locales.

jackjackbits (changes requested):
In-thread search is worth having, but this needs render verification before anything else:

**Blockers:**
1. **Likely dead UI where people actually chat.** `.searchable(text:prompt:)` requires a `NavigationStack`/`NavigationSplitView` ancestor to render, and the public-timeline `MessageListView` (ContentView.swift:539 at your merge-base) has none — the modifier will silently display nothing there on iOS and macOS. Only the DM sheet (which has a NavigationStack, with custom header chrome) might show it. Please attach screenshots of the search field visible in (a) the public timeline and (b) a DM thread, on iOS and macOS. If it doesn't render, this needs a manual search-field UI rather than `.searchable`.
2. **Catalog rewrite + translations** (shared with the batch): whole-file xcstrings re-serialization would clobber main's Aug-10 catalog (93 keys) and break the coverage guard; all 3 new keys are English ×29. Rebase onto current main, surgical adds, real translations.

**Also:**
- `.accessibilityLabel` is applied to the whole ScrollView chain — VoiceOver will relabel the entire message list, not the search affordance. Scope it to the right element.
- Search mode disables message windowing and renders every match from full history — unbounded; consider capping or paginating results.

The `ThreadMessageSearch` filter itself and its unit tests are fine.

Chessing234:
you were right about the render, and it's worse than "might not show": ContentView builds MessageListView directly at :539 and the only NavigationStack in that file is inside a comment at :303 — so .searchable rendered nothing in the public timeline. the DM sheet was the only place it could appear, via the NavigationStack ContentPeopleSheetView opens at :127.

so i dropped .searchable for a plain field. no navigation ancestor to satisfy, renders the same in both placements. it lives inside MessageListView so neither header changes and one implementation covers both call sites; collapses to a magnifier until used and resets on conversation switch.

capped results at 200 newest with a line saying how many were withheld — an uncapped common term was rendering every historical match in one pass. moved the accessibility label off the scrollview chain onto the field itself.

catalog rebased and done surgically, +930/-0, real translations in all 30 locales.

the screenshots i can't give you — this machine has command line tools only, no xcode, so i can't build or photograph it. the static evidence above is why i'm confident .searchable was dead, and a plain TextField has no ancestor requirement to fail. if you want the screenshots before it lands say so and i'll get to a machine with xcode rather than guess.

Chessing234:
awaiting-review — searchable dropped for a plain textfield so it renders outside a navigationstack; catalog is surgical on tip.

PR #1620, opened 2026-08-02
fix: raise iOS max decompressed payload size to match Android (10 MiB)

### LINKED ISSUE #1618: iOS/Android disagree on max expanded (decompressed) payload size — 1.13 MiB vs 10 MiB, silent drop on iOS
## Summary

The two clients disagree on the largest message the mesh will carry, which causes silent, unexplained delivery failures for otherwise-legitimate large compressed messages sent from Android to iOS.

## Details

- iOS: `FileTransferLimits.maxFramedFileBytes` (`localPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift`) = **1,179,760 bytes** (~1.13 MiB)
- Android: `AppConstants.Protocol.MAX_PAYLOAD_LENGTH` (`app/src/main/java/com/bitchat/android/util/AppConstants.kt:70`) = **10,485,760 bytes** (10 MiB)

Both platforms enforce the same 50,000:1 compression-ratio guard on decode, but Android allows a legitimately-compressed message to declare an expanded (`originalSize`) payload up to 10 MiB, while iOS caps the same field at ~1.13 MiB (`BinaryProtocol.swift`'s `decodeCore`: `guard originalSize >= 0 && originalSize <= FileTransferLimits.maxFramedFileBytes`).

## Repro scenario

1. Android sender constructs and sends a compressed message with a declared `originalSize` between ~1.13 MiB and 10 MiB (well within its own limit, and with a compressed on-wire size small enough to pass the shared ratio guard — typical for genuinely compressible content).
2. The packet decodes fine on the Android sender's own re-parse and on any Android recipient.
3. An iOS recipient's `BinaryProtocol.decode` rejects it silently at the `originalSize` bounds check — the message is simply dropped, with no error surfaced on either end.

## Suggested fix

Reconcile the two constants (or otherwise agree on a shared, enforced expanded-payload ceiling) so cross-platform delivery of large compressed messages doesn't silently fail depending on which platform sent it. Happy to be pointed at the intended design if this is a known, accepted asymmetry rather than an oversight.

## Scope note

Filing as a plain compatibility bug, not a security report — no attacker control involved, just legitimate content sent between two honest clients on different platforms.

harshjainnn:
iOS's FileTransferLimits.maxFramedFileBytes was 1,179,760 bytes (~1.13 MiB) while Android's AppConstants.Protocol.MAX_PAYLOAD_LENGTH is 10,485,760 bytes (10 MiB). Since both platforms already enforce the same 50,000:1 compression- ratio guard independently, the lower iOS ceiling served no additional protection — it just caused legitimate large compressed messages from Android to be silently dropped on decode with no error surfaced to either side.

Raise iOS's limit to 10 MiB to match Android's intended design ceiling.

Fixes #1618

### DIFF
--- a/localPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift
+++ b/localPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift
@@ -7,6 +7,7 @@ public enum FileTransferLimits {
     /// Compressed images after downscaling should comfortably fit under this budget.
     public static let maxImageBytes: Int = 512 * 1024 // 512 KiB
     /// Worst-case size once TLV metadata and binary packet framing are included for the largest payloads.
+    /// Note: This is now a hard expanded-size ceiling rather than the old tlv+frame formula.
     /// MUST stay in sync with AppConstants.Protocol.MAX_PAYLOAD_LENGTH in the Android repo (issue #1618).
     public static let maxFramedFileBytes: Int = 10_485_760
 
--- a/localPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift
+++ b/localPackages/BitFoundation/Sources/BitFoundation/FileTransferLimits.swift
@@ -7,15 +7,8 @@ public enum FileTransferLimits {
     /// Compressed images after downscaling should comfortably fit under this budget.
     public static let maxImageBytes: Int = 512 * 1024 // 512 KiB
     /// Worst-case size once TLV metadata and binary packet framing are included for the largest payloads.
-    public static let maxFramedFileBytes: Int = {
-        let maxMetadataBytes = Int(UInt16.max) * 2 // fileName + mimeType TLVs
-        let tlvEnvelopeOverhead = 18 + maxMetadataBytes // TLV tags + lengths + metadata bytes
-        let binaryEnvelopeOverhead = BinaryProtocol.v2HeaderSize
-            + BinaryProtocol.senderIDSize
-            + BinaryProtocol.recipientIDSize
-            + BinaryProtocol.signatureSize
-        return maxPayloadBytes + tlvEnvelopeOverhead + binaryEnvelopeOverhead
-    }()
+    /// MUST stay in sync with AppConstants.Protocol.MAX_PAYLOAD_LENGTH in the Android repo (issue #1618).
+    public static let maxFramedFileBytes: Int = 10_485_760
 
     public static func isValidPayload(_ size: Int) -> Bool {
         size <= maxPayloadBytes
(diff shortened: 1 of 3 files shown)


callebtc (approved):

Chessing234 (commented):
blocker-ish: raising maxFramedFileBytes to a flat 10 miB also raises what a peer can force us to decompress into memory — the old ~1.13 miB framed ceiling was intentional for that (see #1618 / the payload-cap docs). if the goal is android interop, keep the inflate budget separate from the on-wire frame budget, or at least call out the ios memory tradeoff in the pr.

Chessing234:
can you link the actual MAX_PAYLOAD_LENGTH value from the android repo? also the old limit was derived deliberately (payload + tlv/binary overhead), this is a ~9x jump on what an untrusted ble peer can push through decompression before rejection, worth a note on the memory tradeoff. test only covers 2mb, would help to also cover near the new 10mb boundary

jackjackbits (changes requested):
The asymmetry you're fixing is real and verified — Android's `AppConstants.kt:70` is 10 MiB, enforced at both encode and decode, so Android will happily send what iOS silently drops (#1618). But raising `maxFramedFileBytes` isn't one check — it moves five enforcement points at once:

- `BinaryProtocol.swift:333` (on-wire frame cap) and `:369` (the decompress-allocation bound — `CompressionUtil.swift:41` does `UnsafeMutablePointer.allocate(capacity: originalSize)` per packet)
- `NostrInboundPipeline.swift:578` — the embedded `bitchat1:` packet cap, **reachable from any Nostr relay over the internet, no BLE proximity needed**, which would now accept ~14 MB of base64
- `NoiseSecurityConstants.swift:28` (`maxPrivateFilePlaintextSize`)
- `BLEFragmentAssemblyBuffer.swift:209` (reassembly cap, multiplied by `maxInFlightAssemblies`)

With the 50,000:1 ratio guard, a ~210-byte compressed body can force a 10 MiB allocation before rejection. Android mitigates its 10 MiB with a `DecompressionResourcePool`; this PR adds no equivalent, and iOS jetsam limits are tighter. Meanwhile the practical payoff is thin: post-decode file payloads are still capped at 1 MiB (`BitchatFilePacket.swift`), and BLE transfers hit the cross-platform 256-fragment limit long before multi-MiB — so the main thing actually enabled is bigger attacker-supplied inflations on the internet path.

**Asks:**
1. Separate the iOS inflate/allocation budget from the interop ceiling rather than raising them together (the earlier review comment on the memory tradeoff still stands unaddressed).
2. Add the 10 MiB-boundary test and the Android source link that were asked for (the only test is at 2 MB).
3. The rewrite also changes the constant's semantics from a derived framed-envelope bound to a flat number while keeping the "framed" name — rename or re-derive.

The right venue for the shared ceiling is #1618 — it's possible the resolution is Android coming *down* rather than iOS going up. #1623 (logging the rejection) is merging first so the drops are at least visible while that's decided.

PR #1622, opened 2026-08-02
Bound the simulated-mesh storm budget to the traffic under test

ecgang:
## What this is now

#1651 fixed the flake this PR was opened for. It drains the mesh to quiet before the baseline is taken, so the wall-clock-driven announce traffic lands in the baseline instead of in the measured delta. Rebased on top of it, the original justification no longer holds — hence the retitle. This is a legibility change, not a flake fix.

Measured on current `main` plus this commit, burning 0.2s of wall clock (the repro from #1630) and changing nothing else:

| burn | placement | baseline total / announce | delta total / announce / message |
|---|---|---|---|
| none | — | 66 / 22 | 4 / 0 / 4 |
| 0.2s | before the settle loop | 80 / 36 | 4 / 0 / 4 |
| 0.2s | after the baseline | 66 / 22 | 4 / 0 / 4 |

The burn moves the baseline by 14 announce frames and leaves the delta at 4. A total-based bound of `<= 12` now reads 4 whatever the runner is doing.

## Why it may still be worth keeping

The assertion counts frames it isn't about. `SimulatedMesh.deliveredFrameCount` sums every packet type, so the message budget's headroom is shared with announce, handshake and prekey traffic. Announce volume in particular tracks real elapsed seconds, because admission runs through `announceThrottle.shouldSend(force:now: Date())` rather than the manual scheduler. Counting per type makes the bound name the traffic it bounds, so a later change on the announce path cannot spend the message budget unnoticed.

The message-only cost is 4, and invariant under the perturbation above, so the budget is 8.

## Scope

Test-only: a per-type counter beside the existing total in `SimulatedMesh.pump`, plus the one assertion. No production code.

Full `SimulatedMeshTests` green locally (macOS scheme, 8/8).

## Note for reviewers

@Chessing234 approved this on the earlier head (`cae78f7f`), which predates #1651 — the wall-clock reasoning was accurate for the diff reviewed then, and isn't for this one. Flagging it so the approval isn't read as covering a claim I've since measured away. Happy to close this if the narrowing isn't worth a review round.

### DIFF
--- a/bitchatTests/Simulation/SimulatedMesh.swift
+++ b/bitchatTests/Simulation/SimulatedMesh.swift
@@ -29,6 +29,29 @@ final class SimulatedMesh {
     private var pendingDeliveries: [(from: Int, packet: BitchatPacket)] = []
     /// Total (packet, receiving-node) deliveries pumped — the storm bound.
     private(set) var deliveredFrameCount = 0
+    /// The same deliveries split by packet type. A whole-mesh frame budget
+    /// silently includes announce traffic, and announce *admission* is the one
+    /// decision the manual scheduler does not own: `sendAnnounceNow` asks
+    /// `announceThrottle.shouldSend(force:now: Date())` (BLEService), so
+    /// whether a scheduled re-announce reaches the wire depends on real
+    /// elapsed seconds, not on `advanceTime`. Burn 0.2s of wall clock in the
+    /// middle of a test — the forced-announce window is 0.15s — and announce
+    /// deliveries rise (measured: 22 → 36) while every other type is
+    /// unchanged. A test that drains the mesh to quiet before it measures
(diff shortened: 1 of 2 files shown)


Chessing234:
this looks right — bounding by message type instead of total frames should kill the wall-clock announce flake without weakening the storm check.

Chessing234 (commented):
perm check — ignore

Chessing234 (commented):
per-type storm bound looks right — announce admission on wall clock was exactly what made the total-frame budget flake under load.

ecgang:
@jackjackbits ready to merge: up to date with main (5e9287f), all checks green. Test-only: it narrows the storm budget in the simulated-mesh test to the traffic under test, so a future failure names the right cause.

PR #1625, opened 2026-08-03
fix(dm): NFC-normalize nicknames when consolidating geo DM threads

Chessing234:
## Summary
- Geo DM temp-thread consolidation compared senders with `lowercased()` only
- Match on `normalizedNickname` so NFD wire senders merge into NFC peer nicknames (#1502 follow-up)
- Test covers the `café` NFD/NFC pair

## Test plan
- [ ] `PrivateChatManagerTests.consolidateMessages_movesTemporaryGeoDMHistoryByCanonicalNickname`
- [ ] Existing geo DM consolidation tests still pass


jackjackbits:
Closing because the proposed NFC/NFD regression is already handled by the existing Swift String equality comparison, including after lowercased(). The added test does not demonstrate a behavior fixed by this change. A concrete example that fails on main would justify reopening this.

PR #1626, opened 2026-08-03
fix(composer): match @mentions across Unicode NFC/NFD forms

Chessing234:
## Summary
- `parseMentions` required an exact token match against stored nicknames
- Resolve captures via NFC so a combining-accent `@café` maps to the stored precomposed token
- Test covers the NFD typed / NFC stored pair

## Test plan
- [ ] `ChatComposerCoordinatorContextTests.parseMentions_matchesUnicodeEquivalentNicknames`
- [ ] Existing mention parsing tests still pass


Chessing234:
addressed — mention regex now includes `\p{M}` and parseMentions NFC-normalizes before matching, so the NFD `@café` case in the test lands.

Chessing234:
awaiting-review

PR #1627, opened 2026-08-04
fix: log silent rejection of oversized compressed payloads

tagadearpit:
When a compressed packet declares an originalSize larger than FileTransferLimits.maxFramedFileBytes, BinaryProtocol.decodeCore returned nil with no log. The suspicious-ratio path already logs a warning, so this mismatch left legitimate Android	o iOS messages invisible on both ends (see #1618).

Add a matching SecureLogger.warning that records the declared size and the ceiling. Behavior is unchanged: the packet is still refused. Existing oversizedPayloadIsRejected test continues to cover the case.

### DIFF
--- a/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
+++ b/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
@@ -366,7 +366,13 @@ public struct BinaryProtocol {
                     guard let rawSize = read16() else { return nil }
                     originalSize = Int(rawSize)
                 }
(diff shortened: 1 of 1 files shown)


Chessing234 (commented):
same change as #1623 — #1623 has the clearer comment tying it to #1618. maybe close this one so they dont race?

jackjackbits:
Closing as a duplicate of #1623, which predates this by two days and carries the same hunk with an explanatory comment and a full CI run. Nothing against the change itself — it's the right fix, which is why it's landing via #1623.

PR #1631, opened 2026-08-04
feat: preserve original bytes during re-encoding of packets with foreign encoders

areebahmeddd:
## Problem

* signed packets could fail signature verification when compressed by a different DEFLATE implementation, since encoding recompressed the payload before signing and verification
* relay forwarding made the issue worse by recompressing payloads when updating the packet, breaking signatures for downstream peers

## Fix

* preserves the original compressed payload bytes during decode and reuses them when re-encoding, avoiding unnecessary recompression
* only reuses the stored bytes while they still match the decoded payload, otherwise falling back to normal compression
* keeps the wire format unchanged with no compatibility impact

## Tests

* adds a regression test using a handcrafted raw DEFLATE stored block to simulate a packet produced by a different encoder
* verifies that decoding and re-encoding preserves the original wire bytes, ensuring signature verification remains valid across implementations

Related PR: https://github.com/permissionlesstech/bitchat-android/pull/863

### DIFF
--- a/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
+++ b/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
@@ -137,7 +137,16 @@ public struct BinaryProtocol {
         var payload = packet.payload
         var isCompressed = false
         var originalPayloadSize: Int?
-        if CompressionUtil.shouldCompress(payload) {
+        // Re-encode of a decoded packet: reuse the originator's bytes (see WirePayload).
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
foreign compressed re-encode test is the right invariant. nothing covers the uncompressed-on-wire branch (skip shouldCompress / only-if-smaller) — thats the other way a cross-client encoder can diverge.

areebahmeddd:
closing this due to inactivity.. (note: the android pr is clean and merged)

PR #1633, opened 2026-08-05
fix(macOS): keep Tor/BLE alive when minimized

### LINKED ISSUE #1593: Feature Request: Persistent Tor connection on macOS when minimized
On macOS, it seems that the app disconnects from Tor when minimized or a fullscreen application takes focus, preventing bluetooth relay.

It would be nice to leave the app running but minimized to the dock and still be able to relay connections from other devices via bluetooth.

Chessing234:
## Summary

Hold a `ProcessInfo` activity so App Nap cannot throttle Tor and the BLE mesh while the macOS window is minimized or fully occluded (#1593).

Reworked after review — the scope is now the minimize/occlusion case only:

- **Dropped** `applicationShouldTerminateAfterLastWindowClosed → false` and the dock-reopen handler. Closing the last window quits exactly as it does on main, so stopping bitchat still stops the radios. This also removes the deprecated `activate(ignoringOtherApps:)` call and the flaky `WindowGroup` + `applicationShouldHandleReopen` path.
- **Gated** the assertion on a transport actually relaying, rather than holding it for process lifetime: Tor switched on (including bootstrap, which needs CPU before `isReady` flips) or Bluetooth powered on. With Tor off and Bluetooth unavailable there is nothing to relay, so the assertion is released.

## Shape

`RelayActivityAssertion` holds the policy and the begin/end bookkeeping and imports only Foundation. `MacRelayActivityController` owns the Combine subscriptions to `NetworkActivationService`, `TorManager.isReady`, and `ChatViewModel.bluetoothState`. Splitting them keeps the part worth testing free of AppKit and Combine.

| torDesired | torReady | bluetoothPoweredOn | assertion |
|---|---|---|---|
| any true | — | — | held |
| false | false | false | released |

## Verification

Ran locally:

- `RelayActivityAssertionTests` — 16 cases / 33 assertions: the full transport truth table, plus the bookkeeping a leak would otherwise only reveal in Instruments (begin exactly once, no stacking when a second transport comes up, hold through losing one transport, release with the last, re-arm after a radio cycles, idempotent release). Compiled and run against the real `RelayActivityAssertion.swift`.
- Typechecked the changed files under `-swift-version 5`, `-swift-version 6`, and `-strict-concurrency=complete`.

**Not run locally:** the Xcode test suite and any on-device check — this machine has Command Line Tools only, no Xcode, so `xcodebuild` cannot build the app here. CI covers the build, the app test target, the iOS simulator tests, SwiftLint, and the periphery scan.

I have not verified the App Nap behavior on macOS hardware, so I have not ticked that box. If you would rather see a measured before/after (minimize, confirm relay traffic continues) before this lands, say so and I will find a machine with Xcode.

Closes #1593

### DIFF
--- a/bitchat/App/MacAppNapRiskMonitor.swift
+++ b/bitchat/App/MacAppNapRiskMonitor.swift
@@ -1,0 +1,65 @@
+//
+// MacAppNapRiskMonitor.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+#if os(macOS)
+
+import AppKit
+import Combine
+import Foundation
+
+/// Tracks when macOS would apply App Nap to this process.
+///
+/// App Nap throttles a backgrounded app whose windows are all minimized or
+/// fully occluded. While a window is visible, the OS does not nap us and an
+/// activity assertion would be wasted battery for no benefit (#1593).
+@MainActor
+final class MacAppNapRiskMonitor: ObservableObject {
+    @Published private(set) var appNapWouldApply = false
+
+    private var observers: [NSObjectProtocol] = []
+
+    func start() {
+        guard observers.isEmpty else { return }
+        let center = NotificationCenter.default
+        let names: [Notification.Name] = [
+            NSWindow.didMiniaturizeNotification,
+            NSWindow.didDeminiaturizeNotification,
+            NSWindow.didChangeOcclusionStateNotification,
+            NSApplication.didBecomeActiveNotification,
+            NSApplication.didResignActiveNotification
+        ]
+        for name in names {
+            observers.append(
+                center.addObserver(forName: name, object: nil, queue: .main) { [weak self] _ in
+                    self?.refresh()
+                }
+            )
+        }
+        refresh()
+    }
+
+    func stop() {
+        observers.forEach { NotificationCenter.default.removeObserver($0) }
+        observers.removeAll()
+    }
+
+    func refresh() {
+        appNapWouldApply = Self.computeAppNapWouldApply()
+    }
+
+    /// True when every window is minimized or not visibly on screen.
+    static func computeAppNapWouldApply() -> Bool {
+        let windows = NSApplication.shared.windows.filter { $0.isVisible || $0.isMiniaturized }
+        guard !windows.isEmpty else { return true }
+        return !windows.contains { window in
+            !window.isMiniaturized && window.occlusionState.contains(.visible)
+        }
+    }
+}
+
+#endif
--- a/bitchat/App/MacKeepRunningSettings.swift
+++ b/bitchat/App/MacKeepRunningSettings.swift
@@ -1,0 +1,47 @@
+//
+// MacKeepRunningSettings.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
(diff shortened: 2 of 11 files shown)


jackjackbits (changes requested):
The diagnosis is credible — main's dormancy handling is iOS-only, and nothing stops App Nap from freezing Tor/BLE when the macOS window is minimized or occluded (#1593). `beginActivity(.userInitiatedAllowingIdleSystemSleep)` is the standard mitigation, and everything is correctly fenced under `#if os(macOS)`.

**Blockers:**
1. **`applicationShouldTerminateAfterLastWindowClosed → false` needs to be opt-in.** As written, closing the window silently keeps Tor running and BLE broadcasting your nickname — beyond what #1593 asked for ("minimized to dock"), and against this app's privacy posture (panic gestures, privacy screen). Someone who closes bitchat will reasonably believe it stopped. Either scope this PR to the minimize/occlusion case only, or add an explicit "keep running when window is closed" setting that defaults off.
2. **The reopen path needs device verification.** Reopen-after-last-window-close via SwiftUI `WindowGroup` + `applicationShouldHandleReopen` is a known-flaky combination, and `activate(ignoringOtherApps:)` is deprecated on macOS 14+. Your own test plan is entirely unchecked — please run it on hardware and report results in the PR.

**Nit:** the activity is held unconditionally for process lifetime — gate it on Tor/BLE actually being active so an idle configuration doesn't hold the assertion.

Chessing234:
reworked — dropped the last-window-close and dock-reopen changes, so closing the window quits exactly like main does and the deprecated activate(ignoringOtherApps:) path goes with them. the assertion is now gated on a transport actually relaying (tor on including bootstrap, or bluetooth powered on) rather than held for process lifetime, and the policy plus begin/end bookkeeping moved into RelayActivityAssertion so it's testable without appkit — 16 cases over the truth table and the lifecycle (no double-begin, release only with the last transport, re-arm after a radio cycles).

no xcode on this machine, so i couldn't run the xcode suite or check hardware — left those boxes unticked rather than claiming them. if you want a measured minimize/relay run before this lands i'll get to a machine that has it.

the red app-test check is timeoutRestoredSessionDefersQueueDrainUntilConvergence, which also failed on main in run 31418625613 on the 10th. can't rerun it without admin.

Chessing234:
ci is green on tip and the last-window/reopen overreach is already dropped — ready for another look when you have a slot.

Chessing234:
both blockers should already be on the tip: last-window-close still returns true (close means quit), and the nap exemption is gated on tor/ble actually being active via RelayActivityAssertion.shouldHold. reopen-on-hardware still unchecked on my side — if you want that verified before merge i can leave it open for a mac pass.

Chessing234:
scoped the assertion to minimized/occluded windows only and tightened ble gating to activation+powered-on. last-window-close still terminates (returns true). haven't been able to run the reopen path on mac hardware here — will need device verification before merge.

Chessing234:
last-window-close is default-off quit again — `applicationShouldTerminateAfterLastWindowClosed` returns true unless you flip the new macos privacy toggle (settings → keep running when window is closed). app nap assertion unchanged: only while minimized/occluded and tor/ble actually relaying.

Chessing234:
update() tests now set appnapwouldapply; they were starting from the nap=false default.

Issue #1629, opened 2026-08-04
iOS/Android disagree on max expanded (decompressed) payload size — 1.13 MiB vs 10 MiB, silent drop on iOS


tagadearpit:
iOS: FileTransferLimits.maxFramedFileBytes ≈ 1.13 MiB
Android: AppConstants.Protocol.MAX_PAYLOAD_LENGTH = 10 MiB
Both platforms share the 50 000:1 compression-ratio guard, but Android can legitimately produce a compressed message whose declared originalSize is larger than what iOS accepts. iOS drops it silently.

Steps to reproduce:

From Android send a well-compressed payload whose expanded size is e.g. 2–5 MiB.
On iOS the packet never appears in the chat.
No error is shown to either user and (before Issue 1 fix) no log either.

jackjackbits:
Closing as a duplicate of #1618, which holds the design discussion about the iOS/Android decompressed-size ceiling. Please keep follow-up there.

Issue #1628, opened 2026-08-04
Log silent rejection of oversized compressed payloads instead of dropping with no trace


tagadearpit:
When a compressed packet declares an originalSize larger than FileTransferLimits.maxFramedFileBytes (~1.13 MiB), BinaryProtocol.decodeCore returns nil with no log.

The suspicious-compression-ratio guard (right next to it) already logs a warning. Because of this inconsistency, legitimate large compressed messages sent from Android (which allows up to 10 MiB expanded size) are silently dropped on iOS with nothing visible on either side.

This makes #1618 much harder to diagnose in the field.

Steps to reproduce:

On Android, construct/send a compressed message whose declared expanded size is between ~1.13 MiB and 10 MiB (still within Android’s limit and the shared 50 000:1 ratio guard).
Receive it on iOS.
Observe: the message never appears and there is no log entry explaining why it was rejected.
(Alternatively) Unit test path: the existing oversizedPayloadIsRejected test already proves the packet is refused; before this change there was simply no log.
Fix status:
I pushed the fix to your fork on branch:

fix/log-oversized-compressed-payload

Commit: adds a SecureLogger.warning that records the declared size and the ceiling. Behavior is unchanged (packet still rejected).

PR #1634, opened 2026-08-05
docs: why iOS keeps the ~1.13 MiB framed payload cap

Chessing234:
## Summary
- documents why iOS should **not** raise `maxFramedFileBytes` to Android's 10 MiB expanded size (#1618 / #1629)
- decompress allocates from the wire-declared `originalSize` *before* the 50 000:1 ratio guard, so matching Android would grow a cheap remote allocation on the tighter platform
- logging the silent reject stays #1628 / open log PRs — this only records the safe direction

## Test plan
- [ ] skim `FileTransferLimits.maxFramedFileBytes` comment + new PRIVATE-MEDIA section
- [ ] no behavior change

Refs #1618 #1629


Chessing234:
closing docs-only pr per house policy

PR #1636, opened 2026-08-05
ci: prevent manual source manifests from overwriting releases

vekovius:
## What changed

I restricted the release-attachment step in the source-manifest workflow to tag-push events. Manual dispatches still build, verify, attest, and upload the manifest as a workflow artifact.

## Why

A manual run can use a tagged workflow revision while its `ref` input points to a different tag, branch, or commit. The workflow builds the manifest from that input, but the previous condition still attached it to `github.ref_name` with `--clobber`. That could replace a release asset with a manifest for a different source tree.

The updated condition prevents manual runs from mutating release assets. Tag-push reruns retain their existing behavior and are outside this change:

| Event | Attach to release |
| --- | --- |
| Tag push | Yes |
| Manual dispatch from a tag | No |
| Manual dispatch from a branch | No |
| Manual dispatch whose input names a tag | No |

## Validation

- `git diff --check`
- Reviewed the event matrix above against the exact workflow condition
- `actionlint` v1.7.12

This is one of four independent findings from an audit of `main`. It does not depend on the other changes and can be reviewed or merged separately.

<details>
<summary>Related independent findings</summary>

- #1635 — enforce `MessageDeduplicator` count bounds for marked IDs
- #1637 — exclude blocked people from geohash participant counts
- #1638 — make dedup cache eviction generation-aware

</details>

### DIFF
--- a/.github/workflows/source-manifest.yml
+++ b/.github/workflows/source-manifest.yml
@@ -98,7 +98,7 @@ jobs:
           path: SOURCE-MANIFEST.txt
 
       - name: Attach to release
-        if: startsWith(github.ref, 'refs/tags/')
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
         env:
           GH_TOKEN: ${{ github.token }}
         run: |


Chessing234 (commented):
condition's right — `github.ref` is `refs/tags/x` for a dispatch launched from a tag, so the old guard genuinely let that through, and gating on `event_name == 'push'` is the narrowest fix. the `startsWith` is now redundant given `on.push.tags`, but keeping it is the right call.

one adjacent case the same reasoning covers and this doesn't: re-running the tag-push workflow. `event_name` is still `push` on a re-run, checkout resolves `github.ref` to wherever the tag points *now*, and the upload is still `--clobber`. so if a release tag ever gets moved, a re-run replaces the published manifest with one for a different tree — which is the outcome your table is built to prevent, just reached by a different route. not blocking, and arguably out of scope for the title, but if you want release mutation tied to the event that created the tag, a re-run isn't that event either.

vekovius:
@Chessing234 I’m keeping this PR scoped to manual dispatches as titled. Handling reruns requires a separate policy decision around pinning the original event SHA versus preventing later release mutation. I’ll narrow the description so it doesn’t imply this condition also covers tag-push reruns.

PR #1638, opened 2026-08-05
fix: make dedup cache eviction generation-aware

vekovius:
## Summary

Removing and then reinserting a deduplication key left an older queue node pointing at the same map key. When that stale node reached eviction, it could delete the newly inserted value. Removed queue nodes also made `count` and capacity enforcement describe queue storage rather than live entries.

I made map entries and order nodes generation-aware, so eviction only removes the live entry for the matching generation. Updates preserve their original generation and insertion position, while reinsertion receives a new generation. Live count and capacity enforcement now use `map.count`.

Because live-count enforcement no longer consumes removed order nodes, I also compact stale unconsumed nodes independently of eviction pressure. Compaction runs once there are at least 32 stale nodes and they occupy at least half of the unconsumed order storage. The stale count is derived from `order.count - head - map.count`, and rebuilding preserves the order of generation-matching live nodes.

## Tests

- `swift test --filter LRUDeduplicationCacheTests` — 15 tests passed
- `swift test --filter MessageDeduplicationServiceTests` — 53 tests passed
- `swift test --quiet` — 2,008 tests passed

The regressions cover reinsertion, replacement with a different key, non-refreshing updates, bounded storage across 1,001 remove/reinsert cycles, eviction after compaction, and service-level `forgetContent` followed by re-recording.

This is one of four independent findings from an audit of `main`. It does not depend on the other changes and can be reviewed or merged separately.

<details>
<summary>Related independent findings</summary>

- #1635 — enforce the `MessageDeduplicator` count bound for marked IDs
- #1636 — prevent manual source manifests from overwriting releases
- #1637 — exclude blocked people from geohash participant counts

</details>

### DIFF
--- a/bitchat/Services/MessageDeduplicationService.swift
+++ b/bitchat/Services/MessageDeduplicationService.swift
@@ -28,7 +28,6 @@ final class LRUDeduplicationCache<Value> {
     private var map: [String: StoredEntry] = [:]
     private var order: [OrderEntry] = []
     private var head: Int = 0
-    private var staleNodeCount: Int = 0
     private var nextGeneration: UInt64 = 0
     private let capacity: Int
 
@@ -43,11 +42,6 @@ final class LRUDeduplicationCache<Value> {
     var count: Int {
         map.count
     }
-
-    #if DEBUG
-    /// Order-node storage exposed only for bounded-storage regression tests.
-    var _orderStorageCountForTesting: Int { order.count }
-    #endif
 
     /// Checks if a key exists in the cache
     func contains(_ key: String) -> Bool {
@@ -71,14 +65,11 @@ final class LRUDeduplicationCache<Value> {
         }
 
         trimIfNeeded()
-        compactIfNeeded()
     }
 
     /// Removes a specific key from the cache
     func remove(_ key: String) {
-        guard map.removeValue(forKey: key) != nil else { return }
-        staleNodeCount += 1
-        compactIfNeeded()
+        map.removeValue(forKey: key)
     }
 
     /// Clears all entries from the cache
@@ -86,7 +77,6 @@ final class LRUDeduplicationCache<Value> {
         map.removeAll()
         order.removeAll()
         head = 0
-        staleNodeCount = 0
         nextGeneration = 0
     }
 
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
the generation fix is right — remove-then-reinsert really does let the stale node evict the live value on main.

two asks. the test hook diverges from house style: existing seams are plain `bucketCountsForTesting` (MessageRateLimiter.swift:110) with no `#if DEBUG`, and gating it means the `<= 32` assertion — the only thing pinning the storage bound — silently disappears outside a debug run.

and the compaction half looks speculative: `remove` has one caller, `forgetContent` from ChatViewModel.swift:1061 on message deletion, so nothing in the app does 1001 remove/reinsert cycles. worth landing the generation fix alone and justifying `staleNodeCount` separately.

vekovius:
@Chessing234 Addressed in `4816a13`. I removed the stale-node compaction, its debug-only storage hook, and the 1,001-cycle test, leaving this PR focused on generation-aware eviction. The existing consumed-prefix compaction remains. The 14 cache tests, 52 service-filter tests, and full 2,007-test suite pass.

Chessing234 (commented):
generation fix looks right and the trim is much clearer now. but dropping the compaction has a consequence i didn't think through when i asked for it, so this one's partly on me.

`trimIfNeeded` now gates on `map.count`, and stale order nodes aren't in `map`. so a `remove` followed by a `record` leaves `map.count` at or below capacity, no pop happens, `head` never advances, and the `head >= 32 && head * 2 >= order.count` compaction never fires. `order` then grows by one node per remove/record cycle for the life of the process.

main didn't have that: `activeCount = order.count - head` counted stale nodes, so they pushed the trim threshold, forced pops, advanced `head`, and got compacted away. it over-evicted live entries doing it — which is the bug you're fixing — but it did bound the array.

concretely, capacity 3: record a,b,c; remove("a"); record("d") → map 3, order 4, head 0, no trim. repeat and order climbs without limit while map stays at 3.

so the two halves weren't independent the way i claimed. the reclamation needs to come back in some form that isn't tied to eviction pressure — maybe compact when stale nodes (`order.count - head - map.count`) exceed some fraction, which is a live-entry-aware version of what main got for free.

vekovius:
@Chessing234 Addressed in c28dc11. I restored compaction outside eviction pressure and derive stale storage as `order.count - head - map.count`. Once at least 32 stale nodes make up half of the unconsumed order storage, the cache rebuilds the queue from generation-matching live nodes while preserving order.

The storage metric now follows the existing plain `bucketCountsForTesting` style, and the regression verifies that 1,001 remove/reinsert cycles stay at or below 32 stored order nodes before checking subsequent eviction.

Tests:
- `swift test --filter LRUDeduplicationCacheTests` — 15 passed
- `swift test --filter MessageDeduplicationServiceTests` — 53 passed
- `swift test --quiet` — 2,008 passed

PR #1640, opened 2026-08-07
fix: return after peer color cache hit

Chessing234:
peer color cache checked for a hit but never returned, so every call recomputed the color and rewrote the same cache entry. added a debug-only counter so a test can actually assert the cache short-circuits.


Chessing234:
good catch — the seeds are unique per test but the counter isn't, so under --parallel a sibling computing a colour between the two reads would fail an assertion that has nothing wrong with it.

83acfc4 marks the suite @Suite(.serialized). checked that nothing outside this file reads _peerColorComputeCountForTesting, so serializing just this suite covers it.

Chessing234:
awaiting-review

PR #1641, opened 2026-08-07
l10n: localize the voice-recording cancel button

Chessing234:
the icon-only cancel button in the recording HUD used a raw Cancel literal, so voiceover announced english regardless of locale. switched it to the common.cancel key already used everywhere else.

### DIFF
--- a/bitchat/Views/ContentComposerView.swift
+++ b/bitchat/Views/ContentComposerView.swift
@@ -238,7 +238,7 @@ private extension ContentComposerView {
             }
             Spacer()
             Button(action: voiceRecordingVM.cancel) {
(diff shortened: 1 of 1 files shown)


jackjackbits:
Thanks for catching this — it was a real gap when you opened the PR. It's since been fixed on main by #1656 (merged Aug 10): `ContentComposerView.swift:241` now uses `String(localized: "common.cancel", ...)`, reusing the existing translated key. That form is also what our new coverage guard (`everyCodeReferencedKeyExistsInACatalog`, from #1654) scans for, which a bare `LocalizedStringKey` wouldn't satisfy. Closing as superseded.

PR #1642, opened 2026-08-07
fix: evict fully-aged-out geohashes from activity tracker

Chessing234:
messageTimes/lastMessages kept an entry for every geohash ever seen since prune() only emptied the timestamp array, never dropped the dictionary entry. roaming across regions over a long session leaked one entry per geohash for the life of the process. now sweeps and drops fully-aged-out geohashes on each recorded message.


Chessing234:
awaiting-review

PR #1643, opened 2026-08-07
test: add coverage for BoundedIDSet eviction

Chessing234:
the gateway/bridge loop-prevention caches all sit on this type but it had no direct unit tests. covers insert/contains, fifo eviction at capacity, re-insertion of an evicted id, and duplicate inserts not disturbing eviction order.

### DIFF
--- a/bitchatTests/Services/BoundedIDSetTests.swift
+++ b/bitchatTests/Services/BoundedIDSetTests.swift
@@ -1,0 +1,87 @@
+//
+// BoundedIDSetTests.swift
+// bitchatTests
+//
+// Tests for BoundedIDSet insertion-ordered eviction
+//
+
+import Testing
+@testable import bitchat
+
+struct BoundedIDSetTests {
+
+    @Test func insertReturnsTrueForNewID() {
+        var set = BoundedIDSet(capacity: 3)
+        #expect(set.insert("a") == true)
+        #expect(set.contains("a"))
+    }
+
+    @Test func insertReturnsFalseForAlreadyPresentID() {
+        var set = BoundedIDSet(capacity: 3)
+        set.insert("a")
+        #expect(set.insert("a") == false)
+    }
+
+    @Test func containsIsFalseForUnknownID() {
+        let set = BoundedIDSet(capacity: 3)
+        #expect(set.contains("missing") == false)
+    }
+
+    @Test func evictsOldestWhenOverCapacity() {
+        var set = BoundedIDSet(capacity: 2)
+        set.insert("a")
+        set.insert("b")
+        set.insert("c")
+
+        #expect(set.contains("a") == false)
+        #expect(set.contains("b"))
+        #expect(set.contains("c"))
+    }
+
+    @Test func evictedIDCanBeReinserted() {
+        var set = BoundedIDSet(capacity: 1)
+        set.insert("a")
+        set.insert("b") // evicts "a"
+
+        #expect(set.insert("a") == true)
+        #expect(set.contains("a"))
+        #expect(set.contains("b") == false)
+    }
+
+    @Test func stayingUnderCapacityKeepsEverything() {
+        var set = BoundedIDSet(capacity: 5)
+        for id in ["a", "b", "c"] {
+            set.insert(id)
+        }
+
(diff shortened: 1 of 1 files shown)


Chessing234:
awaiting-review

bitchatTests/Services/BoundedIDSetTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        var set = BoundedIDSet(capacity: 1)
+        set.insert("a")
+        set.insert("b") // evicts "a"
+
+        #expect(set.insert("a") == true)
+        #expect(set.contains("a"))
+        #expect(set.contains("b") == false)
+    }
+
+    @Test func stayingUnderCapacityKeepsEverything() {
```
jozanek, line 51: These read as subsumed to me: `evictsOldestWhenOverCapacity` already requires two survivors at capacity, and `evictedIDCanBeReinserted` already runs at capacity 1. Same for `insertReturnsTrueForNewID` and `containsIsFalseForUnknownID` — I can't construct an edit that fails those and nothing else.

The sibling type's tests cover this concern in three (`BridgeDropDedupStoreTests.swift:18-60`). Would you be open to trimming to the ones that can fail alone?

jozanek (commented):
Thanks for this. I read it as: `BoundedIDSet` backs nine loop-prevention caches across `BridgeService`, `GatewayService` and `BLEService`, nothing in the suite names it, and you're closing that gap. That matches the diff, and the gap is real — the only existing test crossing the 512 boundary (`BridgeServiceTests.swift:471`) asserts on radio-alias rows, never on which ID was evicted.

What I'd find most useful is a note on which tests bite. That seems to be the line here: #1550 landed in two days with "changing `case nil: continue` to `return nil` fails both new tests and nothing else"; #1625 was closed for lacking it. I count three that catch a real edit — `insertReturnsFalseForAlreadyPresentID` (five sites branch on that Bool: `GatewayService.swift:390`, `BridgeService.swift:453`, `:726`, `:742`), `evictsOldestWhenOverCapacity`, and `evictedIDCanBeReinserted`, the only one proving `members.remove` accompanies `order.removeFirst()`. The realistic edit they guard is someone replacing that O(n) `removeFirst()` in a mesh hot path.

One correction, to save you a round-trip: I expected to argue that `duplicateInsertDoesNotDisturbEvictionOrder` freezes FIFO against a documented LRU policy (`WHITEPAPER.md:56`). It doesn't — `LRUDeduplicationCache.record` appends to `order` only for new keys, and `MessageDeduplicator.isDuplicate` returns without refreshing. Both are FIFO under an LRU name, so your test pins what every dedup cache here actually does. My ask on it is just the comment, inline.

Asides, not requests: `BoundedIDSet(capacity: 0)` silently disables the cache while `insert` still returns `true` (unreachable today, and `LRUDeduplicationCache` guards the same edge with a `precondition`) — I think you were right not to test that. And `maxEventAgeSeconds` is 900s while `receivedEventIDs` fills in ~43s at the 720/min ceiling, which may deserve its own issue.

What would move me to approve: the note on which tests bite, and test 8's comment giving the reason rather than the mechanic.

bitchatTests/Services/BoundedIDSetTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        set.insert("b")
+        set.insert("c")
+
+        #expect(set.contains("a") == false)
+        #expect(set.contains("b") == false)
+        #expect(set.contains("c"))
+    }
+
+    @Test func duplicateInsertDoesNotDisturbEvictionOrder() {
+        // Re-inserting an already-present ID is a no-op (returns false) and
```
jozanek, line 74: This says what the code does; I couldn't find why. `BoundedIDSet.swift:9-11`, #1384/#1412/#1425 and `docs/` are all silent, so I can't tell whether FIFO was chosen or defaulted into — and a "must" in a test reads as a spec to whoever comes next.

There's a good argument for it: a replay being a free no-op keeps eviction pressure proportional to distinct *signed* events, which is what the 720/min gate at `BridgeService.swift:95` bounds. Would you be open to putting that here instead?

PR #1644, opened 2026-08-07
test: add coverage for PeerDisplayNameResolver

Chessing234:
pure display-name/collision-suffix logic with a couple non-obvious rules (disconnected peers exempt from suffixing and from collision counting, own nickname folded into the count) that had no direct tests.


Chessing234:
awaiting-review

PR #1646, opened 2026-08-08
Make SwiftLint blocking in CI now that the violation backlog is zero

jackjackbits:
Closes out the "advisory-only lint gate" finding from the repo evaluation. The advisory job from #1361 existed because of the violation backlog (#1088) — that backlog turned out to be down to just **28 real violations** (an earlier 23k count was entirely untracked local `.device-lab/` clutter that CI never sees). At zero violations, a baseline file is the wrong tool; the job simply becomes enforcing.

## What changed

**Violations → 0**
- 24× `closure_parameter_position`: capture lists / parameters joined onto the opening-brace line (`BLEService`, `NoiseSessionManager`, `ChatMediaTransferCoordinator`, `BLENoisePacketHandlerTests`).
- 1× `trailing_comma`, 1× `opening_brace` — via `swiftlint --fix`.
- 2× `orphaned_doc_comment`: the `periphery:ignore` comments in `PeerIDRotation` / `AnnounceV2Packet` sat between the doc block and the declaration; moved above the doc block. Verified with a local `periphery scan --strict` against the committed baseline: **no new findings**, so the ignore commands still register.
- 1× `todo`: rule disabled in config instead — the single TODO cites tracked issue #1434 and is a deliberate marker. (This keeps the repo's zero-inline-`swiftlint:disable` record intact.)

**Config**
- `.device-lab` added to `excluded` so local runs match CI.
- `todo` added to `disabled_rules` with rationale.

**CI**
- `continue-on-error` dropped, "(advisory)" label removed, `--strict` added so any new warning fails the job. Same pinned 0.65.0 container, same least-privilege setup.

Not changed: the Periphery job stays advisory (its macOS-only scan has documented false-positive classes), and the "to be enabled as we fix" disabled rules (complexity/length) remain disabled — enabling those is real refactoring work, not a CI flip.

## Testing
- `swiftlint lint --strict` exits 0 locally on SwiftLint 0.65.0 (same version the CI container pins).
- Full SwiftPM suite green after the formatting changes: 2,004 tests / 216 suites.
- `periphery scan --strict`: "No unused code detected."

### DIFF
--- a/.github/workflows/swift-tests.yml
+++ b/.github/workflows/swift-tests.yml
@@ -264,11 +264,13 @@ jobs:
             CODE_SIGNING_ALLOWED=NO \
             test
 
-  # Advisory only: SwiftLint reports style violations without ever failing the
-  # build. Runs in a pinned container (no Xcode plugin, no pbxproj changes) so
-  # it can never break the documented xcodebuild path or block a merge.
+  # Blocking since the backlog reached zero (all violations fixed; rules the
+  # codebase deliberately breaks are disabled in .swiftlint.yml). --strict
+  # promotes warnings to errors so new violations fail the job. Runs in a
+  # pinned container (no Xcode plugin, no pbxproj changes) so it can never
+  # break the documented xcodebuild path.
   lint:
-    name: SwiftLint (advisory)
+    name: SwiftLint
     runs-on: ubuntu-latest
     timeout-minutes: 15
     # This job runs a third-party container image, so give it the least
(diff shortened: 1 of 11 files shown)


jackjackbits:
The strict gate earned its keep on its very first CI run: it failed on the merge commit because #1542 (merged to main after this branch was cut) introduced a new `is_disjoint` violation in `ContentComposerView.handleKeyDown`. Merged main in and fixed it (`modifierFlags.intersection([...]).isEmpty` → `isDisjoint(with:)`). Exactly the class of drift the advisory job was silently accumulating.

Issue #1549, opened 2026-07-29
Arti: No pluggable transports, unsafe as enabled default


joshqou:
Direct connections to the tor network are trivially detected, by even consumer-grade DPI systems (see #1347). In countries that use deep packet inspection such as China and Iran, directly connecting to the tor network by default could put users in danger. This is the primary reason why bridges and pluggable transports exist.

As mentioned in [docs/TOR-INTEGRATION.md](https://github.com/permissionlesstech/bitchat/blob/main/docs/TOR-INTEGRATION.md#known-gap-no-bridges-or-pluggable-transports), vendored Arti is not built with bridges support despite upstream supporting bridges and pluggables.

Unless there's some severe issue with Arti's implementation of pluggables, this is a pretty glaring safety issue that is straightforward to resolve. If Arti can't use pluggables on iOS for whatever reason, tor routing should really be disabled by default.

Chessing234:
covered by open https://github.com/permissionlesstech/bitchat/pull/1534 (obfs4 + snowflake via arti). duplicating that work here would just fight the stacked branch — tracking against that pr.

PR #1650, opened 2026-08-09
Say in the app that tor here has no bridges or pluggable transports

krishrathi1:
Refs #1549. **This does not close it** — see "What this is not" below.

## The gap

`docs/TOR-INTEGRATION.md` has documented since the integration landed that the vendored Arti has no bridges and no pluggable transports:

> `arti-client` is built with `default-features = false` and features `["tokio", "rustls"]` only — no `pt-client`, no `bridge-client`

I checked the source and it is exactly that — `localPackages/Arti/arti-bitchat/Cargo.toml:12-15`, and `arti-bitchat/src/lib.rs:262-270` bootstraps from `TorClientConfigBuilder::from_directories(...).build()` with no bridge lines and no configurable directory authorities. A repo-wide search for `pluggable|obfs4|snowflake|pt_client|bridge-client` across Swift, Rust, TOML and shell returns nothing. (Every `bridge` identifier in the Swift tree is the unrelated BLE↔Nostr mesh bridge.)

Nothing in the app says it. The toggle reads:

> sends internet traffic through tor, so relay operators see tor's address instead of yours. covers location channels and private messages delivered over the internet. **recommended: on.**

and the only caption underneath (`AppInfoView.swift:454-459`) appears when tor is **off**. So the screen is candid about the cost of switching tor off and silent about the cost of leaving it on: a direct connection to the public Tor network is recognisable as Tor to anyone inspecting the traffic. The nearest the UI ever comes is `system.tor.blocked` — "this network may be blocking it" — which is about reachability, after the fact, and offers no remedy.

That is the asymmetry this fixes. Where using Tor is itself the risk — the case that motivates bridges existing at all — "recommended: on" is advice the app can't back up, and the person acting on it has no way to know.

## The change

A caption under the toggle, shown while tor is **on**, in the app's voice:

> no bridges or pluggable transports yet: a network operator can see that you use tor, though not what you send. take care where using tor is itself risky.

Deliberately **not** red like the off-warning: this is a standing limitation of the feature as shipped, not a fault condition. It uses `secondaryTextColor`, matching the mesh-bridge caption directly above it.

Three files, +202/−0: the string and the `else` branch in `AppInfoView.swift`, the catalog entry, and two sentences in `docs/TOR-INTEGRATION.md` recording that the gap is now stated in-app and not only in docs.

## Localization

New key `app_info.settings.tor.no_bridges_note`, populated for **all 30 locales** — `LocalizationCoverageTests.mainCatalogCoversAllLocalesForEveryKey` asserts every key covers the union of locales in the catalog, so an English-only entry would fail CI.

I wrote those 30 translations, so **please have native speakers check them** — fa, zh-Hans, ru and ar especially, since those are the audiences the string is for and a mistranslated security caveat is worse than none. Happy to take corrections in review or drop any locale back to English.

## What this is not

This is not a fix for #1549 and should not close it. Real circumvention needs the pluggable-transport cargo features enabled, bridge configuration plumbed through an FFI that currently exports seven functions (`arti_start`, `arti_stop`, `arti_is_running`, `arti_bootstrap_progress`, `arti_bootstrap_summary`, `arti_go_dormant`, `arti_wake`) and none of which accept config, plus a rebuild under the pinned Rust 1.96.0 toolchain and a provenance-manifest update — all in one PR, per `arti-provenance.yml`. That is a build-side change this cannot substitute for.

I also did not take the issue's other suggestion of defaulting tor off. That trades a detectability risk for an unconditional one — every relay seeing every user's IP, including relays carrying private messages — and it is a product decision for maintainers, not a drive-by change. This closes the disclosure gap in the meantime, which costs nothing and helps someone today.

## Verification status

I have no Xcode/Swift toolchain on this machine, so I have **not** compiled the project or run the suite — CI is the first real check on the Swift.

The catalog edit I did verify locally: the file parses as JSON (428 keys), and I re-implemented `LocalizationCoverageTests`' logic against it — every key, including the new one, covers all 30 locales. The diff is +186 lines in the catalog with no reformatting of existing entries.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -3535,6 +3535,192 @@
         }
       }
     },
+    "app_info.settings.tor.no_bridges_note" : {
+      "comment" : "Caption shown under the tor toggle while tor is on, disclosing that the bundled tor client ships without bridges or pluggable transports, so a network operator can identify the connection as tor",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "لا توجد جسور أو وسائل نقل قابلة للتوصيل بعد: يمكن لمشغّل الشبكة أن يرى أنك تستخدم tor، وإن لم يرَ ما ترسله. توخَّ الحذر حيث يكون استخدام tor بحد ذاته خطرًا."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "এখনও ব্রিজ বা প্লাগেবল ট্রান্সপোর্ট নেই: নেটওয়ার্ক অপারেটর দেখতে পায় যে আপনি tor ব্যবহার করছেন, যদিও আপনি কী পাঠাচ্ছেন তা নয়। যেখানে tor ব্যবহার করাই ঝুঁকিপূর্ণ, সেখানে সতর্ক থাকুন।"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "noch keine bridges oder pluggable transports: ein netzbetreiber kann sehen, dass du tor benutzt, aber nicht, was du sendest. vorsicht dort, wo schon die nutzung von tor riskant ist."
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "no bridges or pluggable transports yet: a network operator can see that you use tor, though not what you send. take care where using tor is itself risky."
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "todavía sin puentes ni transportes conectables: un operador de red puede ver que usas tor, aunque no lo que envías. ten cuidado donde usar tor ya sea arriesgado."
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "هنوز پل یا ترابری اتصال‌پذیر وجود ندارد: اپراتور شبکه می‌تواند ببیند که از tor استفاده می‌کنید، هرچند نه آنچه می‌فرستید. جایی که خودِ استفاده از tor خطرناک است، احتیاط کنید."
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "wala pang bridges o pluggable transports: makikita ng operator ng network na gumagamit ka ng tor, kahit hindi kung ano ang ipinapadala mo. mag-ingat kung saan mapanganib na ang paggamit mismo ng tor."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "pas encore de bridges ni de pluggable transports : un opérateur réseau peut voir que tu utilises tor, mais pas ce que tu envoies. prudence là où utiliser tor est déjà risqué."
+          }
+        },
+        "he" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "עדיין אין גשרים או תעבורות מתחלפות: מפעיל הרשת יכול לראות שאתה משתמש ב-tor, אך לא מה שאתה שולח. היזהר במקומות שבהם עצם השימוש ב-tor מסוכן."
+          }
+        },
+        "hi" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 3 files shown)


krishrathi1:
@ecgang  please review my changes ...

krishrathi1:
@ecgang  please review my changes ...

ecgang:
Reviewed. This looks right to me, and the catalog edit is the part I checked hardest — `Localizable.xcstrings` is easy to wreck by re-serialising it, and you didn't.

Parsed both sides and compared against `origin/main`:

- exactly one key added, `app_info.settings.tor.no_bridges_note`, none removed
- the order of every pre-existing key is unchanged, and no pre-existing entry's content differs
- 30 locales on the new key, which is the same set all 426 other keys carry — nothing missing, nothing extra
- every `stringUnit` is `translated` with a non-empty value, and no locale is an untranslated copy of the English string
- entry shape matches its sibling `app_info.settings.tor.off_warning` exactly: `comment`, `extractionState: manual`, `localizations`

The claim itself checks out against the repo: `docs/TOR-INTEGRATION.md:45` records `arti-client` built with `default-features = false` and `["tokio", "rustls"]` only — no `pt-client`, no `bridge-client`. So the caption states a real property of what ships, not a guess.

Placement reads correctly: the `else` hangs off `if !locationChannelsModel.userTorEnabled`, so the note appears while Tor is **on**, and secondary rather than `alertRed` — a standing limitation of the feature, not a fault. That matches how the off-state warning already behaves.

One thing I considered and decided is fine: the binding is the user's preference, not bootstrap state, so the note shows even where Tor is blocked and never connects. That is still true and still worth saying — a failed bootstrap to the public relays is exactly as identifiable as a successful one.

What I did not check: translation quality across 30 languages. I read the English, Spanish, French and Italian and they carry the same meaning; the rest I verified structurally only.

Full CI green.

Chessing234 (commented):
the disclosure gap is real and the caption is the right place for it. two things on the strings.

the 30 locales use three different policies for tor's feature names, and four of them switch mid-sentence: pl "mostków i pluggable transports", sv "bryggor eller pluggable transports", th "บริดจ์หรือ pluggable transport", vi "cầu nối hay pluggable transport". de/fr/nl/it/id/ms/fil keep both terms in english, es translates both ("puentes ni transportes conectables"). "pluggable transports" is tor's own documented name and it's what someone would search for a remedy — worth picking one policy and applying it, probably latin everywhere like you did for tor itself.

second: torSubtitle two lines above still ends "recommended: on", which is the exact advice your PR body says the app can't back up where using tor is itself the risk. right now the card recommends it unconditionally and then qualifies it in smaller grey text. i know touching that string is another 30-locale edit, so maybe a separate one, but leaving them contradicting each other reads worse than either alone.

krishrathi1:
@Chessing234 both addressed.

**Terminology:** you're right about pl/sv/th/vi -- checked the other 26 locales and there isn't actually a clean three-way split, it's two consistent policies (7 keep both terms in Latin: de/fr/nl/it/id/ms/fil; the other 19 fully translate or transliterate both) plus these 4 that switch mid-sentence. Since "pluggable transports" is the term someone would search for a remedy, like you said, I reverted the translated word (mostków/bryggor/บริดจ์/cầu nối) back to English "bridge(s)" in just those four, so each string now reads consistently in Latin script for that term -- matching the 7 that already did, and matching how `tor` itself sits untranslated in all 30. I did not touch the 19 locales that fully translate both terms consistently within themselves (ar, bn, es, fa, he, hi, ja, ko, ne, pt, pt-BR, ru, ta, tr, uk, ur, zh-Hans, zh-Hant) -- making "pluggable transports" Latin-only *everywhere* would mean rewriting entire established phrases in 18 more languages I can't verify grammatically, which felt like a bigger and riskier call than the actual bug you found (a string contradicting itself mid-sentence). Happy to revisit if you think the wider policy change is worth it regardless.

**torSubtitle:** agreed it's not worth leaving for later -- dropped the trailing "recommended: on." from the subtitle across all 30 locales and the English defaultValue in AppInfoView.swift, so the toggle just states what tor does and lets the two conditional captions underneath (off warning in red, no-bridges note in grey while on) carry the actual guidance instead of a blanket recommendation contradicting one of them.

Both were done as exact-substring replacements against the existing catalog text (find each locale's known trailing/embedded clause, verify it matches byte-for-byte, replace, done), not a re-parse-and-reserialize of the whole file -- diffed old vs new structurally afterward: same 521 keys, same 30 locales per changed key, same `state`/`comment`/`extractionState` on every entry, only the two `value` fields actually differ. Rebased onto current upstream/main while I was in there (12 commits, none touching these exact lines, rebased clean).

Still no Xcode toolchain on this end, so as before this is verified by structural diffing and careful reading, not a build+run -- flagging that again rather than letting it go unsaid twice.

krishrathi1:
@Chessing234 @ecgang All terminology and catalog points addressed, with 7/7 passing CI checks. Please take a re-look when you have time!

PR #1651, opened 2026-08-09
Deflake gift-wrap tests: settle deadlines, not latency budgets

jackjackbits:
## What

Three CI runs this week failed on gift-wrap round-trip tests (`handleGiftWrap_privateMessageStoresConversationAndMapping`, `handleGiftWrap_deliveredAckUpdatesExistingMessage` on the swift-test job; `handleGiftWrap_routesEmbeddedPrivateMessageAndDeduplicates` on a rerun) — always the same signature: the async NIP-17 unwrap hop missed the wait window, so `didStore`/`routed` asserted false.

**Diagnosis:** 40 local iterations of both suites (`-run-tests-until-failure`, ~2000 test cases) pass clean, ruling out a real crypto bug. The failing waits used explicit `timeout: 5.0`, `longTimeout` (10s), or `TestHelpers.waitUntil`'s 5s default — all below what `TestConstants.settleTimeout`'s own doc block calls the rule: *a wait deadline is not a latency budget; size it for the worst-case scheduler.*

**Fix:** every positive (expected-true) wait in `ChatViewModelExtensionsTests` and `ChatNostrCoordinatorContextTests` now passes `TestConstants.settleTimeout` (30s): 5× explicit `5.0`, 7× `longTimeout`, 2× bare default, +2 in the coordinator tests. All return as soon as the condition holds, so green runs pay nothing. No negative waits in these files (verified per call site). Stale comment referencing "the 5s default" updated.

Not touched here: `TestHelpers.waitUntil`'s 5s *default* itself — ~28 call sites across other files rely on it and a few are negative-polarity waits that would each burn 30s; that's a follow-up with per-site polarity triage. Also seen once: `SimulatedMeshTests.publicMessageRelaysAcrossLineTopologyWithinTTLBudget` (delivered 18 > budget 12) — different class (fanout determinism assumption), left for its own investigation if it recurs.

## Tests

`ChatViewModelNostrExtensionTests` + `ChatViewModelPrivateChatExtensionTests` + `ChatViewModelMediaTransferTests` + `ChatNostrCoordinatorContextTests` + `TestTimingHygieneTests`: 49/49 green (count-verified).

### DIFF
--- a/bitchatTests/ChatNostrCoordinatorContextTests.swift
+++ b/bitchatTests/ChatNostrCoordinatorContextTests.swift
@@ -356,7 +356,7 @@ struct ChatNostrCoordinatorContextTests {
 
         // The NIP-17 unwrap runs off the main actor; wait for the hop back.
         let convKey = PeerID(nostr_: sender.publicKeyHex)
-        let routed = await TestHelpers.waitUntil({ context.handledPrivateMessages.count == 1 })
+        let routed = await TestHelpers.waitUntil({ context.handledPrivateMessages.count == 1 }, timeout: TestConstants.settleTimeout)
         #expect(routed)
         #expect(context.recordedNostrEventIDs == [giftWrap.id])
         #expect(context.nostrKeyMapping[convKey] == sender.publicKeyHex)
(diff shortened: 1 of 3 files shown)


jackjackbits:
Added the second flake from this week's CI runs to this PR: `SimulatedMeshTests.publicMessageRelaysAcrossLineTopologyWithinTTLBudget` (failed at 14 and 18 frames vs the 12 budget on two runs).

Root cause found: the frame **baseline** was snapshotted after a single 2s advance, but discovery isn't quiet yet — every first-seen peer schedules an afterglow re-announce at a **random 0.3–0.6s** delay (`BLEAnnounceHandler:287`), and those cascade further relay rounds. Whether that traffic lands before or after the snapshot depends on the draw. The test now settles until a full window passes with no new frames before taking the baseline, so the budget measures only the message under test. The 12-frame storm bound itself is unchanged. 30 local iterations green.

(That afterglow delay is also the one remaining random timer breaking the deterministic-sim promise from #1548 — making it injectable like `connectedAnnounceJitterOffset` would be a nice follow-up.)

PR #1652, opened 2026-08-10
i18n: review and refine Korean localization

promptprobe:
## Summary

* review the Korean main-app and share-extension catalogs against the current English source
* correct inaccurate, machine-translated, and unnatural Korean wording
* normalize recurring Korean terminology across mesh, relays, geohash features, private messaging, identity verification, and Nostr
* preserve plural metadata and positional format arguments
* add focused regression coverage for a Korean positional-placeholder mismatch

## Testing

* jq empty bitchat/Localizable.xcstrings bitchatShareExtension/Localization/Localizable.xcstrings
* swift test --filter KoreanLocalizationFormatTests
* swift test --filter LocalizationCoverageTests
* swift test
* macOS Debug build with code signing disabled
* iOS Simulator Debug build with code signing disabled


jackjackbits (changes requested):
Mechanically this is exemplary — only `ko` values touched, zero key adds/deletes/renames, plural and substitution metadata preserved, the `%@ / %@` → `%1$@ / %2$@` positional fix is correct, and there's no collision with the 93 keys main gained Aug 10. The translation quality itself is genuinely good. Four asks before merge:

1. **Drop 사용자.** Seven keys introduce it (e.g. `mesh_peers.tooltip.vouched` actively swaps 사람 → 사용자; also `content.input.placeholder.mesh`, `system.mesh.block_failed`, …). House rule: no "user"-equivalent in any locale — use 사람, as main's own newest ko strings do.
2. **Revert 노스트르 → Nostr.** Four strings transliterate Nostr while the PR (correctly) keeps Bluetooth/Tor/Noise in Latin — be consistent with your own policy.
3. **Restore `state:"translated"`.** 87 entries were flipped to `needs_review` — backwards for strings you just reviewed; harmless at runtime but it wrecks the catalog's progress signal.
4. **Restore the "over the internet" clause** in `content.empty.location_intro` — the current rewrite loses the meaning that location channels work without proximity.

On register: the PR normalizes toward 합니다체 while main's newest ko strings use 해요체, which would leave the catalog split roughly half/half. We'll make an explicit register call and follow up here — the four asks above stand regardless of which way that goes.

promptprobe:
Thanks for the detailed review. Addressed the four requested items, and made one additional house-rule alignment:

* removed the seven 사용자 usages introduced by this PR
* restored canonical Nostr naming in all four strings
* restored all 87 reviewed entries to translated
* restored the “over the internet” meaning in content.empty.location_intro
* additionally changed the remaining 사용자 in system.mesh.blocked to 사람 so the modified PR scope consistently follows the house rule

I left the Korean register unchanged pending your call on 합니다체 vs 해요체.

The full localization test suite and both macOS and iOS Simulator builds passed after the requested fixes. The final one-string house-rule adjustment also passes catalog parsing, diff/semantic checks, and the focused Korean localization format test.

Chessing234 (commented):
i ran the placeholder check across the whole catalog rather than just ko, and the result says something about the test rather than about your translations.

your ko values are clean — i compared every ko string unit and plural/device variation on this branch against its en counterpart and the format-specifier signatures match on all 146 changed entries, zero mismatches. so the `%@ / %@` → `%1$@ / %2$@` fix is the only one ko needed.

but the same check over all 30 locales on this branch says `location_channels.subtitle_prefix` is still non-positional in fifteen of them:

```
ar de es fr he id it ja ms ne pt-BR ru uk zh-Hans zh-Hant   ->  "#%@ • %@"
en, ko (this branch)                                        ->  "#%1$@ • %2$@"
```

`KoreanLocalizationFormatTests` asserts en and ko only, so it goes green on exactly the state above — the one locale it covers is the one locale that can't regress, and the fifteen that currently differ never get looked at. that's a lot of test for one string in one language.

the check is three lines wider if it walks `localizations` instead of naming `"ko"`, and it stops being a korean test:

```swift
for (locale, localization) in localizations {
    #expect(Self.placeholders(in: try value(for: locale)) == expected, "\(locale)")
}
```

worth going further and asserting the signature of *every* key against its en value rather than one key against a literal — that's the shape that would have caught the ko bug before anyone read the strings, and it generalises to the whole catalog. jack has an open invitation on #1613 for file:line output from the localization guard, so a general format check would sit naturally next to it.

one thing i deliberately did not flag: `fingerprint.message.vouched_by` uses `%d` directly in bn / hi / ne / ta where en goes through the `%#@people@` substitution. that's legal xcstrings and reads deliberate to me, so i've left it out of the count above — mentioning it only so you know the difference between 15 and 19 if you run the same scan.

no comment on register or terminology; that's jack's call and he's already made the asks.

PR #1653, opened 2026-08-10
Deflake AppArchitectureTests: raise the local waitUntil to settleTimeout

jackjackbits:
Post-merge follow-up: the swift-test job on #1588's merge commit failed in `peerListModelSurfacesRecentChatsForAbsentPeers` — the geoDM row-collapse wait missed `AppArchitectureTests`' file-local `waitUntil` default of **3 seconds** (below the `minimumSettleTimeout` floor `TestConstants` documents; same class as #1651). Default raised to `settleTimeout` (30s). All callers in the file are expected-true waits, so passing runs return as soon as the condition holds.

### DIFF
--- a/bitchatTests/AppArchitectureTests.swift
+++ b/bitchatTests/AppArchitectureTests.swift
@@ -584,11 +584,19 @@ struct AppArchitectureTests {
 
         // While that person is visible in the geohash roster, the chat row
         // collapses — same absent-from-rosters contract as mesh.
-        viewModel.participantTracker.setActiveGeohash("u4pruy")
-        viewModel.participantTracker.recordParticipant(pubkeyHex: pubkeyHex, geohash: "u4pruy")
-
+        //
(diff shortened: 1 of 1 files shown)


jackjackbits:
The timeout bump alone wasn't the fix here — CI failed again with 60s available. Reproduced locally with `swift test --parallel`: the geo-dedup wait wasn't slow, it was **permanently broken** under parallel load. The view model's own channel binding delivers its initial `.mesh` selection asynchronously, and when it lands it resets the active participant geohash (`GeohashSubscriptionManager` → `setActiveParticipantGeohash(nil)`), wiping the test's tracker setup — so the row could never collapse.

The wait now re-asserts `setActiveGeohash` + `recordParticipant` on every poll (both idempotent), the same self-healing pattern `ChatViewModelExtensionsTests` already uses for singleton interference. 6× full `swift test --parallel` runs green locally. The settleTimeout bump stays — it's still the right floor for the rest of the file.

PR #1655, opened 2026-08-10
Localize CommandProcessor: all 48 command strings, 30 locales

jackjackbits:
Part 2 of the unlocalized-strings cleanup — stacked on #1654 (base branch; will retarget to main when it merges).

`CommandProcessor.swift` had **zero** `String(localized:)` calls: every command error, usage hint, and the entire `/help` reference text was hardcoded English, while the autocomplete panel (`CommandInfo`) is fully localized — localized suggestions while typing, English the moment anything goes wrong.

- All 47 literal sites converted (verified: zero raw message literals remain); interpolations use `String(format:)` with positional specifiers (`%1$@`, `%2$lld`) so languages can reorder.
- `/trace` pluralization done properly (`path_one`/`path_many`) instead of the `hop\(hops == 1 ? "" : "s")` hack; the `"you"` label at the start of the path chain is localized too.
- `helpText`/`groupUsage` become computed vars so they resolve in the current locale.
- 49 `command.*` keys × 30 locales; command syntax (`/msg`, `@name`, `cashuA…`, `#mesh`) stays verbatim, only prose translates; `/help` keeps its 16-line structure in every locale (validated programmatically along with specifier parity).
- One audit copy fix riding along: `blocked X. you will no longer receive messages from them` → `…no longer see their messages` (blocking filters at display time; packets still arrive and relay). Test assertion updated.

## Tests
`CommandProcessorTests` (21) + `LocalizationCoverageTests` (4, incl. the new code-reference guard): 25/25 green. iOS build green.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -4279,6 +4279,9120 @@
         }
       }
     },
+    "command.action.not_found" : {
+      "comment" : "Error when an action command can't resolve its target; placeholders are the command and the nickname",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "تعذّر %1$@ %2$@: غير موجود"
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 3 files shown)


jackjackbits:
Auto-closed by GitHub when the stacked base branch was deleted during the #1654 merge — recreated unchanged against main.

PR #1659, opened 2026-08-11
Add a read-receipt toggle

jackjackbits:
Tier-1 UX-audit batch 1/6. Read receipts were unconditional — a presence oracle ("awake, holding the phone, opened the app at 3:14") with no setting; every mainstream messenger ships this switch. Default stays ON. All four origination paths gated (mesh/nostr routing, direct mesh, geohash, PrivateChatManager); withheld receipts are still recorded locally as sent so re-enabling never fires a retroactive burst disclosing past reads. Only outbound receipts are affected. Toggle in the privacy section; reset on panic wipe; 2 strings × 30 locales; tests pin default, reset, and nothing-reaches-transport-while-off.

Stacked series: this → #ptt-consent → #identity-changed → #spoofing → #favorites-consent → #app-lock. Merge in order WITHOUT --delete-branch until the child retargets.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -15253,6 +15253,378 @@
         }
       }
     },
+    "app_info.settings.read_receipts.subtitle" : {
+      "comment" : "Subtitle explaining what the read-receipt setting shares and what turning it off withholds",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "يتيح للآخرين معرفة متى قرأت رسائلهم الخاصة. الإيصال يكشف أيضًا أنك كنت مستيقظًا وفتحت التطبيق في تلك اللحظة — أوقف هذا الخيار لتُبقي نشاط قراءتك لنفسك. ستظل ترى الإيصالات التي يرسلها الآخرون."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "অন্যরা দেখতে পাবে আপনি কখন তাদের ব্যক্তিগত মেসেজ পড়েছেন। রসিদটি এটাও জানিয়ে দেয় যে সেই মুহূর্তে আপনি জেগে ছিলেন এবং অ্যাপটি খুলেছিলেন — আপনার পড়ার তথ্য নিজের কাছে রাখতে এটি বন্ধ করুন। অন্যরা যে রসিদ পাঠায় তা আপনি তবুও দেখতে পাবেন।"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "lässt andere sehen, wann du ihre privaten nachrichten gelesen hast. eine bestätigung verrät auch, dass du in dem moment wach warst und die app geöffnet hast — schalte das aus, um deine leseaktivität für dich zu behalten. bestätigungen von anderen siehst du weiterhin."
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "lets people see when you've read their private messages. a receipt also says you were awake and opened the app at that moment — turn this off to keep your reading activity to yourself. you'll still see receipts others send."
+          }
+        },
+        "es" : {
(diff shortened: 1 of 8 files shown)


jackjackbits:
Codex P1 addressed: withheld receipts now land in **both** tracking sets — the manager's withheld branch also records into the owner's persisted set via a bootstrapper-wired `markReceiptHandled` hook, so the lifecycle pass can never retroactively send a receipt for a message read while the setting was off. Test strengthened to require both sets.

RodsF1 (commented):
The send-gating is complete — I traced every read-receipt send site and all three transports are guarded (`routeReadReceipt:174`, `sendMeshReadReceipt:188`, `sendGeohashReadReceipt:206`), including the geohash path that's easy to miss. And the withhold-records-as-sent trick is the right instinct:

```swift
// re-enabling the setting must never fire a retroactive burst
// disclosing past reads
guard sendsReadReceipts() else { return true }   // routeReadReceipt
```

But I think that invariant — the headline promise of the PR — is **only half-enforced, and there's a concrete path that breaks it**. This is subtle, so let me lay out the whole chain; I verified each link on the branch.

There are **two separate** "already sent" sets:
- `ChatViewModel.sentReadReceipts` (persisted), written by `markReadReceiptSent` (`ChatViewModel.swift:536` — inserts into this set *only*).
- `PrivateChatManager.sentReadReceipts` (in-memory, `PrivateChatManager.swift:31`).

The review-fix (cfa87545) bridges them **one way only**: when the *manager* withholds, it also marks the viewModel set (`markReceiptHandled`, `PrivateChatManager.swift:270`). Its own docstring says why — because the lifecycle read pass dedups against the viewModel set. But nothing bridges the other direction, and one of the withhold paths lives on the other side:

- Incoming message **while you're viewing the chat** is handled by the coordinator: `ChatPrivateConversationCoordinator.handlePrivateMessage` → `sendMeshReadReceipt` (gated, so with the toggle OFF it sends nothing) → then `markReadReceiptSent(message.id)` (`:700`). So the message is recorded in the **viewModel set only** — never in the manager set.
- `PrivateChatManager.markAsRead` runs on **every chat open** (`startChat` → `markAsRead`, `:227`) and re-scans the conversation, sending a receipt for any message **not in the manager's own set** (`:243`, `!sentReadReceipts.contains`) — it never consults the viewModel/persisted set — *before* its `guard sendsReadReceipts()`.

So:

1. Toggle **OFF**. You're viewing chat P; peer sends M. Receipt withheld; M recorded in the viewModel set, **not** the manager set.
2. You leave P, turn the toggle **ON** (maybe for a different conversation).
3. You reopen P → `startChat` → `markAsRead` → M isn't in the manager set → `sendsReadReceipts()` is now true → **a read receipt for M is sent.**

The peer gets a read receipt, at the moment you re-enabled, for a message you read while receipts were off — the exact retroactive disclosure the code comment promises can't happen. It's bounded to a single session (the in-memory `ConversationStore` is wiped on relaunch, so the re-scan finds no old messages after restart), but within a session it's a real delayed presence-oracle leak, firing precisely when the user believes they've just turned protection *on*.

Fix options, any one of which closes it: have `markReadReceiptSent` also insert into the manager's set; have `markAsRead`'s re-scan dedup against the persisted set; or unify the two sets. And a regression test that drives the *coordinator* path — toggle off, deliver a PM while the chat is selected, toggle on, reopen, assert nothing leaves — would have caught it (the current test only exercises the manager-withhold path the fix already patched).

Separately, a product question rather than a bug: **the default is `?? true`** (receipts on). Status-quo-preserving and matches mainstream messengers, so defensible — but this is an opt-out for a signal the subtitle itself describes as "you were awake and opened the app at that moment," and opt-outs are found only by people already looking. #1660 in this same stack flips live voice to off-by-default on essentially identical reasoning; worth deciding the two consistently. (Scope is honest, at least: delivery acks stay on, correctly, since delivery is a transport fact rather than a statement about your attention, and the copy says "read receipts.")

Chessing234 (commented):

RodsF1:
Merge-order note: #1701 targets this branch rather than main, so it'd disappear when this lands — on a squash with the head branch deleted GitHub closes a dependent PR instead of retargeting it (#1655 and #1404 both went that way).

PR #1660, opened 2026-08-11
PTT consent: live voice off by default; slide-away-to-cancel

jackjackbits:
Tier-1 batch 2/6 — the last unfixed item from the audit's critical tier. Live PTT defaulted ON in both directions: holding the mic in a public channel streamed voice to radio range before release, and inbound bursts auto-played out the speaker; a buried toggle defaulting on was not consent. Now opt-in (default false) with settings copy saying exactly what on means. And recording could not be aborted one-handed — release always sent, and the HUD cancel button required lifting the finger, which sent. Sliding off the button (>60pt) arms cancel (HUD flips to "release to cancel"); sliding back re-arms send; the armed flag resets on every exit path (pinned by test). 2 new strings + 1 updated × 30 locales.

### DIFF
--- a/bitchat/Features/voice/PTTSettings.swift
+++ b/bitchat/Features/voice/PTTSettings.swift
@@ -16,11 +16,16 @@ import AppKit
 /// User preference for live push-to-talk voice. One switch controls both
 /// directions: streaming your holds live, and auto-playing inbound bursts.
 /// Off means voice messages behave exactly like classic voice notes.
+///
+/// Off by default: live mode sends audio to other people BEFORE the hold is
+/// released and plays strangers' voices out of the speaker unprompted —
+/// both are consequences someone must opt into, not discover. A single
+/// buried toggle defaulting on was not consent.
 enum PTTSettings {
     private static let liveVoiceEnabledKey = "ptt.liveVoiceEnabled"
 
     static var liveVoiceEnabled: Bool {
-        get { UserDefaults.standard.object(forKey: liveVoiceEnabledKey) as? Bool ?? true }
+        get { UserDefaults.standard.object(forKey: liveVoiceEnabledKey) as? Bool ?? false }
         set { UserDefaults.standard.set(newValue, forKey: liveVoiceEnabledKey) }
     }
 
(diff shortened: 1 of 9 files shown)


jackjackbits:
Codex P1 addressed: the live-voice reads are now injectable (`ChatLiveVoiceCoordinator.liveVoiceEnabled`), all 18 fixtures opt in per instance, and the toggle-off test drives the provider instead of mutating the shared preference (which would race parallel suites). `ChatLiveVoiceCoordinatorTests` 20/20 locally.

RodsF1 (commented):
Traced the safety-critical change and it holds: the default-off flip is real (`PTTSettings.swift` now `object(forKey:) as? Bool ?? false`) and gated at all three sites — outbound session choice, inbound frame drop, and receiver-player creation. With the default in place the classic record-then-send path is the only behaviour, and there slide-away-to-cancel genuinely aborts with zero transmission: cancel discards the local file and nothing streams during the hold. Mic teardown on cancel / background / interruption is clean (no hot mic), and `PTTLiveVoiceSession` has no ungated constructor. Good.

One thing the copy overpromises, in the *opt-in* live mode only:

**"slide away to cancel" can't recall audio already streamed.** In live mode `capture.onFrames` sends voice packets **as you speak** (`VoiceCaptureSession.swift`). On cancel, the session stops further capture and sends a `.canceled` control packet — a *retraction*. The receiver's handler stops the player, removes the bubble, and deletes the partial file, so the trailing audio and the saved note are suppressed. But every frame spoken *before* the slide was already transmitted, and if the recipient had live voice on and the chat open, already **played aloud in real time**. Cancel can't un-hear it.

Concrete: a user with live voice enabled holds the mic, says "the meeting point is —", panics and slides off. The HUD swaps to "release to cancel"; they believe nothing was sent. The recipient already heard "the meeting point is" live.

Not a regression — before this PR live was on by default *and* had no cancel at all, so release always sent the whole note; this strictly improves things. But the "cancel" / "release to cancel" wording implies a protection it can't provide once streaming has started. Worth wording live-mode cancel as "stop / retract" rather than "cancel," or having the HUD distinguish "won't send the rest" from "un-sends what you said." The default classic path is unaffected and genuinely safe.

Chessing234 (commented):

PR #1661, opened 2026-08-11
Warn in the chat when a peer's identity key changes

jackjackbits:
Tier-1 batch 3/6. migrateNoiseKeyUpdate silently merged the DM thread onto a peer's NEW noise key: whoever holds the new key inherited the conversation's earned trust under the same nickname, and any earlier verification (bound to the old fingerprint) vanished without a word — Signal treats this as a full-width banner; bitchat treated it as a debug log. The migration now appends a system line to the affected conversation naming the person and saying earlier verification no longer applies. Only when there's a conversation to protect (selected chat or migrated messages). This is also the prerequisite UX for the peer-ID rotation project, where identity changes become routine. 1 string × 30 locales; pinned both ways by coordinator tests.

### DIFF
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -15997,6 +15997,192 @@
         }
       }
     },
+    "system.identity.key_changed" : {
+      "comment" : "Private-chat system warning after a peer's Noise identity key changed; placeholder is the peer's name",
+      "extractionState" : "manual",
+      "localizations" : {
+        "ar" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "تغيّر مفتاح هوية %@ — قد يعني هذا جهازًا جديدًا أو إعادة ضبط. لم يعد التحقق السابق ساريًا؛ تحقّق من هويته مجددًا قبل الوثوق بهذه المحادثة."
+          }
+        },
+        "bn" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@-এর পরিচয় চাবি বদলে গেছে — এর মানে নতুন ডিভাইস বা রিসেট হতে পারে। আগের যাচাই আর প্রযোজ্য নয়; এই চ্যাটে ভরসা করার আগে তাকে আবার যাচাই করে নাও।"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "der identitätsschlüssel von %@ hat sich geändert — das kann ein neues gerät oder ein reset bedeuten. frühere verifizierung gilt nicht mehr; verifiziere die person erneut, bevor du diesem chat vertraust."
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@'s identity key changed — this can mean a new device or a reset. earlier verification no longer applies; verify them again before trusting this chat."
+          }
+        },
+        "es" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "la clave de identidad de %@ cambió — puede significar un dispositivo nuevo o un restablecimiento. la verificación anterior ya no vale; verifica de nuevo su identidad antes de confiar en este chat."
+          }
+        },
+        "fa" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "کلید هویت %@ تغییر کرده — این می‌تواند به معنی دستگاه جدید یا بازنشانی باشد. تأیید قبلی دیگر معتبر نیست؛ پیش از اعتماد به این گفتگو دوباره هویتش را تأیید کن."
+          }
+        },
+        "fil" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "nagbago ang identity key ni %@ — puwedeng ibig sabihin nito ay bagong device o reset. hindi na umiiral ang dating beripikasyon; i-verify siya ulit bago pagkatiwalaan ang chat na ito."
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "la clé d'identité de %@ a changé — cela peut vouloir dire un nouvel appareil ou une réinitialisation. la vérification précédente ne vaut plus ; vérifie à nouveau son identité avant de faire confiance à cette conversation."
+          }
+        },
+        "he" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "מפתח הזהות של %@ השתנה — זה יכול להעיד על מכשיר חדש או איפוס. האימות הקודם כבר לא תקף; כדאי לאמת שוב לפני שסומכים על הצ'אט הזה."
+          }
+        },
+        "hi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@ की पहचान कुंजी बदल गई है — इसका मतलब नया डिवाइस या रीसेट हो सकता है। पहले किया गया सत्यापन अब मान्य नहीं; इस चैट पर भरोसा करने से पहले दोबारा सत्यापित कर लो।"
+          }
+        },
+        "id" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "kunci identitas %@ berubah — bisa berarti perangkat baru atau reset. verifikasi sebelumnya tidak berlaku lagi; verifikasi dia lagi sebelum memercayai chat ini."
+          }
+        },
+        "it" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "la chiave d'identità di %@ è cambiata — può voler dire un nuovo dispositivo o un ripristino. la verifica precedente non vale più; verifica di nuovo la sua identità prima di fidarti di questa chat."
+          }
+        },
+        "ja" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@のアイデンティティキーが変わりました — 新しい端末かリセットの可能性があります。以前の確認はもう有効ではありません。このチャットを信頼する前に、もう一度相手を確認してください。"
+          }
+        },
+        "ko" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@의 신원 키가 변경되었어요 — 새 기기나 초기화일 수 있어요. 이전 확인은 더 이상 유효하지 않으니, 이 채팅을 신뢰하기 전에 상대를 다시 확인해 주세요."
+          }
+        },
+        "ms" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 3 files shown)


jackjackbits:
Both Codex findings addressed: the warning now marks the migrated chat **unread** (unless it's the open conversation), and names offline favorites from the persisted relationship's `peerNickname` before falling back to `resolveNickname` — the anon-prefix fallback hit exactly the offline-key-rotation case this warning exists for. Both pinned by new tests.

RodsF1 (commented):
I think this warning **can never fire in production**, and I'd rather flag that than let it merge looking like protection.

The only non-test caller of `migrateNoiseKeyUpdate` is `handleFavoriteStatusChanged` (`ChatPeerIdentityCoordinator.swift:407`), which opens with:

```swift
guard let peerPublicKey = notification.userInfo?["peerPublicKey"] as? Data else { return }
```

and then gates the warning behind `userInfo["isKeyUpdate"] == true` + `userInfo["oldPeerPublicKey"]`.

But on this branch every production post of that notification is bare:

```
FavoritesPersistenceService.swift:211   post(name: .favoriteStatusChanged, object: nil)
FavoritesPersistenceService.swift:298   post(name: .favoriteStatusChanged, object: nil)
```

No `userInfo` at all — so the **first** guard returns before the `isKeyUpdate` check is even reached. `git grep isKeyUpdate` on the branch returns only two *readers* (`MessageRouter.swift:171` and this coordinator) and **zero writers**. The historical writer, `FavoritesPersistenceService.updateNoisePublicKey(from:to:)`, was deleted in `b2214e30` ("Remove unused favorites and notification helpers", #693) — which is an ancestor of this branch *and* of `main`, so the trigger has been gone since well before this work started.

So the concerning scenario is the one this PR is meant to cover: a favourited contact's key changes (reinstall, or someone announcing their nickname under a new Noise static key), the peer reconnects, and **no warning appears anywhere** — because nothing posts the signal that would produce one.

The tests don't catch it because they call `migrateNoiseKeyUpdate` directly (`ChatPeerIdentityCoordinatorContextTests.swift:385/422/433`), which exercises the message composition and the new unread flag but never the path from a real event. That's why CI is green on an inert feature. A test that posts a real `.favoriteStatusChanged` and asserts a warning appears would have surfaced this.

Worth saying clearly: **the logic in this PR is good.** The favourites-aware naming, the unread marking so the notice isn't silently scrolled past, the message copy — all fine, and worth keeping. The problem is purely that nothing invokes it.

Two things I'd want before merge:

1. **A real trigger.** Either re-establish key-change detection that posts the signal, or — probably better — hang the warning off the transport layer, which already detects this correctly today: `BLEAnnounceTrustPolicy` returns `.keyMismatch` / `.signingKeyMismatch` / `.authenticatedSigningKeyMismatch` and `BLEAnnounceHandler.swift:160-169` handles it, but only writes to `SecureLogger`. The detection exists and is sound; it just has no user-visible surface. Wiring this PR's notice to *that* would make the feature real and reuse a baseline that's already pinned rather than self-asserted.

2. **If the old trigger is revived as-was, don't revive it naively.** The removed detector matched favourites by **nickname** — attacker-controlled, self-asserted data — which would make the warning trigger on the wrong signal. And `migrateNoiseKeyUpdate` migrates chat history and the fingerprint mapping to the new key *before* appending the "verify again" notice, so trust earned by the old key silently transfers to an unverified new one. Whatever re-wires this should pin against the verified key and think about whether migration should happen before the user has acknowledged.

Happy to be wrong if there's a trigger I've missed — but I grepped the branch, its base, and `main` for writers of `isKeyUpdate` / `oldPeerPublicKey` and for `updateNoisePublicKey`, and came up empty on all three.

RodsF1:
Following up on my review — here's a concrete way to make this warning real, but it's an architecture call, so I'd rather propose it than push a rewire onto your branch. Deferring the decision to you.

## Two routes

**Route 1 — resurrect the favorites-layer trigger.** Re-add `updateNoisePublicKey(from:to:)` and post `.favoriteStatusChanged` with `isKeyUpdate`/`oldPeerPublicKey`. I'd argue against this: the removed detector matched a favorite by **nickname** (`favorites.values.first { $0.peerNickname == nickname }`) — attacker-controlled, self-asserted data — and `migrateNoiseKeyUpdate` migrates chat history and the fingerprint mapping to the new key *before* the user acknowledges. So it would warn based on a spoofable signal and silently move earned trust onto the new key first.

**Route 2 — wire the notice to the detection you already have (recommended).** The transport layer already catches this correctly and only logs it. `BLEAnnounceHandler.swift:160-171` handles three rejections from `BLEAnnounceTrustPolicy` — `.keyMismatch`, `.signingKeyMismatch`, `.authenticatedSigningKeyMismatch` — each with a `SecureLogger.warning` and nothing user-visible. Crucially, `BLEAnnounceHandlingPolicy.evaluate` (`:82-102`) only returns a mismatch when a key is **already pinned** (`if let existingNoisePublicKey…`, `if let existingSigningPublicKey…`, `if let authenticatedSigningPublicKey…`), so it **cannot fire on TOFU first contact** — the exact false-positive that would train users to ignore it. And `authenticatedSigningKeyMismatch` compares against an Ed25519 key bound inside a completed Noise session, which is as strong a baseline as exists here. This is the signal the user-facing notice should key off.

## Concrete wiring (Route 2)

1. **Emit at the reject sites.** Add a `notifyIdentityMismatch` closure to `BLEAnnounceHandlerEnvironment` and call it in the three `if case .reject(...)` blocks (`BLEAnnounceHandler.swift:160-171`), keeping the existing log lines.
2. **Hop to the UI actor.** In `BLEService.makeAnnounceHandlerEnvironment`, wire it through the same `notifyUI { deliverTransportEvent(...) }` path the peer-connected/disconnected events already use (`BLEService.swift:2739`, `:3579`), emitting a new `TransportEvent.peerIdentityMismatch(peerID:reason:)`.
3. **Dispatch it.** Add the case to `TransportEvent` (`Transport.swift:68`) and its switch (`:291`, beside `.peerConnected`), plus a delegate method, routing to a new `ChatPeerIdentityCoordinator.handleIdentityMismatch(for:reason:)`. Use a small UI-safe `reason` enum so the BLE policy type doesn't leak across the transport boundary.
4. **Reuse the notice.** Factor the notice-building block out of `migrateNoiseKeyUpdate` (`ChatPeerIdentityCoordinator.swift:626-656`) into a shared `warnIdentityChanged(for:)` — it already appends the `system.identity.key_changed` message (no new strings), marks the thread unread when it isn't selected, and notifies the UI. `handleIdentityMismatch` calls that helper directly.
5. **De-dup per peer** so a replayed mismatched announce warns at most once per offending key, not once per packet.

## Decisions that are yours to make

- **Route 1 vs 2.** I recommend 2 for the reasons above.
- **No migration on this path.** A signing-key mismatch keeps the *same* `peerID` (it derives from the noise key), so `handleIdentityMismatch` should warn into the existing thread and migrate nothing — unlike `migrateNoiseKeyUpdate`, whose noise-key-rotation case moves chat to a new `peerID`. Worth confirming you agree the two cases stay separate. (Note that `migrateNoiseKeyUpdate`'s own trigger is separately dead — re-establishing the offline-favorite-rotation *migration* is a distinct question from making the *warning* fire, and I'd keep them separate.)
- **Which of the three rejections warrant a user warning.** `signingKeyMismatch` / `authenticatedSigningKeyMismatch` are the clear impersonation signals. `keyMismatch` is a noise-key conflict under an existing peerID; the `reason` enum lets you choose whether to surface it or keep it log-only.

## Test

Drive the real path, not the method directly (that's how the current test masks the dead code): model on `signingKeyMismatchWithPinnedKeySkipsUpsertAndIdentityPersistence` (`BLEAnnounceHandlerTests.swift:519`) — pin a signing key, feed an announce with a different one, and assert `notifyIdentityMismatch` fires; then a coordinator-level test asserting `handleIdentityMismatch` appends the system notice and marks unread. This fails on the current branch (nothing reaches the notice) and passes once wired.

If Route 2 sounds right, I'm happy to open a PR against this branch with the wiring + tests. Wanted your call on the route and the migration question first, since it touches the transport event path.

Chessing234 (commented):

RodsF1:
@jackjackbits while you're triaging — the finding from my review above is still open and I don't think the approval reaches it: on this branch the warning can't fire in production.

`migrateNoiseKeyUpdate`'s only non-test caller is `handleFavoriteStatusChanged`, gated on `userInfo["isKeyUpdate"]` and `["oldPeerPublicKey"]`. Nothing in the tree posts either key — the writer `updateNoisePublicKey(from:to:)` was removed in #693 and no branch re-adds it. The three tests pass because they call the method directly, which is also why it had to move out of the private extension.

Route 2 in my comment above is still what I'd recommend: wire the notice to the pinned-key mismatches `BLEAnnounceTrustPolicy` already detects and only logs. It's TOFU-safe by construction — `evaluate` only returns a mismatch when a key is already pinned, so it can't fire on first contact. Roughly five production files.

Still your architecture call and I don't want to push a rewire at you — just say which route and I'll open it against this branch.

shubhambhandari126:
Adjacent case with the same broken binding — and I think the registry already carries what's needed to catch it.

This issue is **same nickname, new key**. There's a mirror: **same key, new nickname.**

A peer that has earned `verified` or `vouched` can change its self-chosen nickname to match someone the user trusts, and the row keeps rendering the badge beside the new name:

- `VouchAttestation.signableBytes` covers `voucheeFingerprint | voucheeSigningKey | timestampMs` — correctly, nothing about a name.
- `PeerListModel` (~L234–242) derives `verifiedBadge` / `vouchedBadge` from the fingerprint, then sets `displayName: peer.displayName`.
- `BitchatPeer.displayName` (~L66) falls back to the peer-claimed `nickname` whenever `localPetname` is empty.

So the badge is truthful about the key while the label next to it is attacker-chosen, and a person reading the row binds the two.

`localPetname` does defuse this — but only when it's set. It's the "local alias" field on the fingerprint sheet, and `verifyFingerprint` (`FingerprintView` ~L248) doesn't populate it, so the label right after verifying is still the peer's own nickname. I also couldn't find any duplicate-nickname disambiguation; the `isDuplicate` hits are all `MessageDeduplicator`.

The detection already exists, it just goes to a log — `BLEAnnounceHandler` ~L237:

```swift
} else if let previousNickname = update.previousNickname, previousNickname != announcement.nickname {
    SecureLogger.debug("🔄 Peer \(peerID.id.prefix(8))… changed nickname: \(previousNickname) -> \(announcement.nickname)", category: .session)
}
```

`BLEPeerAnnounceUpdate.previousNickname` is right there. Options, cheapest first:

1. When a **verified or vouched** peer's claimed nickname changes, surface it wherever this issue's key-change warning lands, instead of only logging it.
2. Seed `localPetname` from the claimed nickname at the moment of verification, so the label the user sees is pinned to what they actually verified.
3. Flag when two peers currently in view claim the same nickname and either one is badged.

Happy to send a PR for (1) or (2) if you'd take one — say which shape you'd prefer.

Read from `main`; I haven't run it on device, so the severity call is yours.

shubhambhandari126:
Patched the mirror case in #1708 (same key, new nickname) — no wire change, binds the seal to the name it was earned under on the receiver side. It deliberately stops at suppressing the seal rather than adding copy, since explaining it to the user belongs with the notice plumbing here.

PR #1662, opened 2026-08-11
Close the system-message spoofing hole in the action parser

jackjackbits:
Tier-1 batch 4/6. processActionMessage rewrote the sender to "system" — the formatter's trusted styling — for ANY `* … *` content containing 🫂, 🐟, or the substring "took a screenshot", so a hostile peer could render arbitrary text as a system-authored line (`* SECURITY: your session key expired, re-verify at evil.example — bob took a screenshot *`). Only the exact locally-generatable action shapes qualify now, and the actor slot must equal the actual wire sender — a peer can hug/slap/screenshot only as themselves; the target slot is bounded, single-line, nickname-shaped. Same templates iOS and Android emit, so legit actions render unchanged; everything else falls through as an ordinary peer message under the sender's real name. Pinned by tests incl. the original spoof payload and actor-mismatch cases. No new strings.

### DIFF
--- a/bitchat/ViewModels/ChatPrivateConversationCoordinator.swift
+++ b/bitchat/ViewModels/ChatPrivateConversationCoordinator.swift
@@ -823,11 +823,21 @@ final class ChatPrivateConversationCoordinator {
     }
 
     func processActionMessage(_ message: BitchatMessage) -> BitchatMessage {
-        let isActionMessage = message.content.hasPrefix("* ")
-            && message.content.hasSuffix(" *")
-            && (message.content.contains("🫂")
-                || message.content.contains("🐟")
-                || message.content.contains("took a screenshot"))
+        // This rewrite hands the message to the formatter as sender "system",
+        // which styles it as trusted, non-peer content. Substring sniffing
+        // ("contains 🫂") let ANY sender put arbitrary text in that styling:
+        // `* SECURITY: session expired — bob took a screenshot *` rendered as
+        // a system-authored line. Only the exact, locally-generatable action
+        // shapes qualify, and the actor slot must be the actual wire sender —
+        // a peer can only "hug"/"slap"/"screenshot" as themselves.
+        guard message.content.hasPrefix("* "), message.content.hasSuffix(" *") else { return message }
+        let inner = String(message.content.dropFirst(2).dropLast(2))
+        let sender = message.sender
+
+        let isActionMessage =
+            Self.matchesActionTemplate(inner, prefix: "🫂 \(sender) hugs ", suffix: "")
+            || Self.matchesActionTemplate(inner, prefix: "🐟 \(sender) slaps ", suffix: " around a bit with a large trout")
+            || inner == "\(sender) took a screenshot"
 
         guard isActionMessage else { return message }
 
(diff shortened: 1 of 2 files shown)


jackjackbits:
Both Codex findings addressed:

- **Regression (suffixed senders):** the actor is now matched by base name via `splitSuffix`, so a location-channel action from `bob#ab12` with unsuffixed content `bob hugs alice` renders as a system action again. Test added.
- **Residual spoof (target preamble):** the target slot must now be a single name token — "you" or a whitespace-free ≤32-char nickname (optionally `#abcd`) — so `bob hugs SECURITY: reset your keys at evil.example` no longer qualifies (spaces). A space-containing nickname degrades to a plain message, the safe direction. Test added.

Chessing234 (commented):
the actor-anchoring is the right shape and the template match kills the preamble case. but the target slot still admits the payload your own example is built around.

`isNameToken` accepts any whitespace-free token up to 32 chars. `https://evil.tld` is 16 chars with no whitespace, so:

```
* 🫂 <attacker's own nick> hugs https://evil.tld *
```

passes `matchesActionTemplate`, the actor equals the wire sender, and it renders as system. and the content body isn't skipped for system senders — `formatMessage`'s `sender != "system"` guard only skips the `<@name>` prefix, the match loop below still runs — so `hasURLHint` fires on `://` and the token comes out blue, underlined, with `matchStyle.link` set. attacker-chosen tappable link under trusted styling, one packet, no preamble needed.

`www.evil.tld` works the same way via the `www.` hint.

the bound that would hold is the nickname charset rather than "no spaces": handleEmote only ever puts a resolved nickname or `you` there, and nicknames don't contain `:` or `/`. rejecting those two characters closes it; matching the actual `#abcd` suffix shape closes it properly.

RodsF1 (commented):
The target-slot hardening in the follow-up is good, but the same class of bypass survives one slot over: **the actor slot is still exploitable, because it's a self-chosen nickname.**

`isNameToken` is applied to the *target*, but the *actor* (`senderBase`) is compared raw. Anchoring the actor to the wire sender stops a peer acting as *someone else* — but `InputValidator.validateUserString` accepts any non-control characters up to 50, **spaces included**, so the preamble the target slot now rejects just moves into the name:

```
nickname:  SECURITY: session expired, re-verify at evil        (44 chars, ≤ 50 limit)
sends:     * SECURITY: session expired, re-verify at evil took a screenshot *
```

That's a peer speaking as itself, so `inner == "\(senderBase) took a screenshot"` matches and the line is handed to the formatter as sender `"system"` — 44 attacker-chosen characters in the trusted, non-peer styling. The `🫂 … hugs you` and `🐟 … slaps you …` templates carry it too.

I confirmed it against this branch's HEAD (`d5309219`) with a test written first — all three templates render as `"system"`, while every existing case here stays green, so this is a gap in the fix rather than a regression in it:

```swift
let hostileNick = "SECURITY: session expired, re-verify at evil"  // 44 ≤ 50
#expect(processed("* \(hostileNick) took a screenshot *", sender: hostileNick).sender == hostileNick)
#expect(processed("* 🫂 \(hostileNick) hugs you *",        sender: hostileNick).sender == hostileNick)
#expect(processed("* 🐟 \(hostileNick) slaps you around a bit with a large trout *", sender: hostileNick).sender == hostileNick)
```

The fix is small — give the actor the same name-token constraint, just at the nickname length bound so a legitimately long name doesn't silently stop rendering:

```swift
let senderBase = message.sender.splitSuffix().0
// The actor is a self-chosen nickname, so it needs the same constraint as
// the target slot — otherwise the rejected preamble just rides in the name.
guard Self.isNameToken(senderBase, maxLength: InputValidator.Limits.maxNicknameLength) else {
    return message
}
```

```swift
// `maxLength` defaults to the target-slot bound; the actor passes the full
// nickname limit so a legitimate long name still renders.
static func isNameToken(_ token: String, maxLength: Int = 32) -> Bool {
    if token == "you" { return true }
    return !token.isEmpty
        && token.count <= maxLength
        && !token.contains(where: { $0.isWhitespace })
}
```

A space-containing nickname then degrades to a plain message under its own name — the same trade already accepted for the target slot. Happy to open a PR against this branch with the fix + the test if that's easier than folding it in here.

RodsF1:
Opened #1699 against this branch with the fix + the repro test, so it can fold in before this merges.

RodsF1:
Merge-order note: #1699 targets this branch rather than main, so it'd disappear when this lands — on a squash with the head branch deleted GitHub closes a dependent PR instead of retargeting it (#1655 and #1404 both went that way).

PR #1663, opened 2026-08-11
Favorites consent, visible mutuality, and honest copy

jackjackbits:
Tier-1 batch 5/6. The star read like a private bookmark but behaved like a friend request + identity handshake: it immediately notified the peer and transmitted your durable Nostr key, while the tapper saw nothing — and mutuality (the state that actually gates offline delivery) was invisible at the point of decision. Now: a one-time consent dialog before the first favorite (both star surfaces; /fav as an explicit typed command proceeds without a dialog; acknowledged flag resets on panic wipe); half star until reciprocated with tooltips + VoiceOver labels; FEATURES copy leads with the real mechanism instead of notifications; and geohash block copy says the block is per-channel (identities are derived per-geohash) on both the context-menu and /block paths. 7 new strings + 3 updated values × 30 locales.

### DIFF
--- a/bitchat/App/PrivateConversationModels.swift
+++ b/bitchat/App/PrivateConversationModels.swift
@@ -105,6 +105,9 @@ struct PrivateConversationHeaderState: Equatable {
     let displayName: String
     let availability: PrivateConversationAvailability
     let isFavorite: Bool
+    /// Whether the favorite is reciprocated — the load-bearing state:
+    /// one-sided favorites do NOT enable offline delivery.
+    let isMutualFavorite: Bool
     let encryptionStatus: EncryptionStatus?
 
     var supportsFavoriteToggle: Bool {
@@ -258,6 +261,7 @@ final class PrivateConversationModel: ObservableObject {
                 displayName: displayName,
                 availability: .meshReachable,
                 isFavorite: false,
+                isMutualFavorite: false,
                 encryptionStatus: nil
             )
         }
@@ -282,6 +286,7 @@ final class PrivateConversationModel: ObservableObject {
             displayName: displayName,
             availability: availability,
             isFavorite: chatViewModel.isFavorite(peerID: headerPeerID),
+            isMutualFavorite: peer?.isMutualFavorite ?? false,
             encryptionStatus: encryptionStatus
         )
     }
--- a/bitchat/Localizable.xcstrings
+++ b/bitchat/Localizable.xcstrings
@@ -5216,181 +5216,181 @@
         "ar" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "تم حظر %@ في محادثات geohash"
+            "value" : "تم حظر %@ في قناة geohash هذه — في قنوات geohash الأخرى يظهر كشخص جديد، فاحظره هناك أيضًا إذا لزم الأمر"
           }
         },
         "bn" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "geohash চ্যাটে %@-কে ব্লক করা হলো"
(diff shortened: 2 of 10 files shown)


jackjackbits:
Both Codex findings addressed: every add-favorite entry point (star button, context menu, VoiceOver custom action) now routes through `requestFavoriteToggle`, which shows the one-time consent dialog before disclosing the nostr key; removals stay immediate. And the empty-star tooltip no longer claims a favorite that doesn't exist — the mutual/pending copy applies only once `isFavorite`, otherwise it reads "add favorite".

RodsF1 (commented):
The UI consent flow here is well built — I checked the pre-consent leak question carefully and it's clean: in both `MeshPeerList.requestFavoriteToggle` and the private-chat header, the dialog is shown *before* any toggle, and `acknowledge()` + the toggle run only inside the confirm branch, so nothing is published or notified before the user agrees. Cancel leaves no half-favourited state. Mutuality display is derived from local state only (`isFavorite && theyFavoritedUs`), so the half-star doesn't leak anything to an observer. Incoming favourite notifications only record `theyFavoritedUs` — they never auto-add — so there's no hidden path that shares your key. And panic wipe resets the flag so a wiped device asks again. All good.

One gap, against the review-fix commit's own claim ("consent on **every** add-favorite path"):

### `/fav` bypasses the gate entirely

`FavoriteConsent` appears **zero times** in `CommandProcessor.swift`. `/fav` dispatches to `handleFavorite(add:)`, which calls straight through:

```swift
// toggleFavorite persists by the real noise key and notifies the peer.
contextProvider?.toggleFavorite(peerID: peerID)
```

— with the comment stating exactly what makes the gate necessary. Downstream that reaches `UnifiedPeerService.toggleFavorite`, which persists the relationship (pinning the Nostr key) and then unconditionally calls `router.sendFavoriteNotification(to:isFavorite:true)`. The gate exists only in the two view layers (`ContentSheetViews.swift:525`, `MeshPeerList.swift:288`).

Concretely: fresh install, user has never favourited anyone, types `/fav alice`. Alice is notified and receives the durable Nostr pubkey immediately — no dialog, no disclosure, no chance to decline. The one-time education screen this PR is built around never runs.

There's a second-order oddity too: `/fav` never calls `acknowledge()`, so the flag stays `false` and the user's *next* star tap still shows the consent dialog — i.e. the disclosure arrives after they've already blind-favourited by command. Conversely, one star tap retroactively silences the dialog for all future `/fav` use. The two entry classes are out of sync in both directions.

A slash command has no view to present a confirmation in, so the fix isn't quite the same shape as the UI gate — options are to print the disclosure and require an explicit confirm step, or to treat an un-acknowledged `/fav` as an error that points at the UI. Either way, worth closing or narrowing the claim, since "every path" is the property the PR is selling.

### Smaller note

The consent is a **single global one-time flag**, not per-peer — after the first acknowledgement every subsequent favourite, including brand-new peers and therefore brand-new durable identity linkages, is silent. The docstring says this is deliberate ("asks once, then never again") and that's defensible for an education screen. But since the thing being consented to is *per-peer* identity linkage, it's worth a conscious decision rather than an inherited one. Not a blocker.

Also: unfavouriting notifies the peer too (`sendFavoriteNotification(isFavorite:false)`) and is intentionally ungated. That's reasonable — it discloses no *new* identifier — but no copy anywhere mentions it, and "removals stay immediate" is only visible in a code comment.

Chessing234 (commented):

RodsF1:
Merge-order note: #1702 targets this branch rather than main, so it'd disappear when this lands — on a squash with the head branch deleted GitHub closes a dependent PR instead of retargeting it (#1655 and #1404 both went that way).

PR #1664, opened 2026-08-11
Add an optional app lock (Face ID / Touch ID / passcode)

jackjackbits:
Tier-1 batch 6/6. The keychain is AfterFirstUnlock and private chats live in memory, so an unlocked, seized, or borrowed phone handed over everything — the threat model's primary event had no defense beyond the panic wipe (which requires still holding the phone). AppLockModel gates the whole UI: locked from the first frame when enabled, re-locks on iOS backgrounding (macOS locks at launch only; PrivacyScreen already covers snapshots), auto-triggers the system prompt with a manual retry. Off by default; the toggle refuses to arm without a device passcode; deliberate fail-open if the passcode is later removed (removing it requires knowing it) — stated in the settings copy. Reset by panic wipe. NSFaceIDUsageDescription added; 5 strings × 30 locales; provider-injected model fully unit-tested.

### DIFF
--- a/bitchat/App/AppLockModel.swift
+++ b/bitchat/App/AppLockModel.swift
@@ -1,0 +1,103 @@
+//
+// AppLockModel.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Combine
+import Foundation
+import LocalAuthentication
+
+/// Optional biometric/passcode gate on the whole app.
+///
+/// The keychain is AfterFirstUnlock and private chats live in memory, so an
+/// unlocked, seized, or borrowed phone hands over everything — the panic
+/// wipe only helps while the phone is still in your hand. This is the
+/// missing complement: with the lock enabled, returning from the background
+/// requires Face ID / Touch ID / the device passcode.
+enum AppLockSettings {
+    private static let enabledKey = "privacy.appLockEnabled"
+
+    static var isEnabled: Bool {
+        isEnabled(in: .standard)
+    }
+
+    static func isEnabled(in defaults: UserDefaults) -> Bool {
+        defaults.bool(forKey: enabledKey)
+    }
+
+    static func setEnabled(_ enabled: Bool, in defaults: UserDefaults = .standard) {
+        defaults.set(enabled, forKey: enabledKey)
+    }
+
+    /// Panic-wipe hook: a wiped device behaves like a fresh install.
+    static func reset(in defaults: UserDefaults = .standard) {
+        defaults.removeObject(forKey: enabledKey)
+    }
+}
+
+@MainActor
+final class AppLockModel: ObservableObject {
+    @Published private(set) var isLocked: Bool
+    @Published private(set) var isAuthenticating = false
+
+    private let isEnabledProvider: () -> Bool
+    private let authenticate: (@escaping @MainActor (Bool) -> Void) -> Void
+
+    /// Providers are injectable so tests drive the lock without LAContext
+    /// or the shared UserDefaults.
+    init(
+        isEnabledProvider: @escaping () -> Bool = { AppLockSettings.isEnabled },
+        authenticate: ((@escaping @MainActor (Bool) -> Void) -> Void)? = nil
+    ) {
+        self.isEnabledProvider = isEnabledProvider
+        self.authenticate = authenticate ?? Self.systemAuthenticate
+        // Locked from the first frame when enabled: the gate must cover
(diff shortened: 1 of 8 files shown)


jackjackbits:
Both Codex findings addressed:

- **Stale callback:** each `lockIfEnabled` bumps a `lockGeneration` and an unlock callback whose generation no longer matches is ignored — a Face ID success that resolves after a background re-lock can't bypass the gate. New test pins it.
- **Fail-open leaves setting on:** the passcode-removed path now calls `AppLockSettings.setEnabled(false)`, so the toggle stops claiming the lock is on and re-adding a passcode later doesn't silently reactivate it — matching the copy's "turns itself off" promise.

RodsF1 (commented):
Reviewed this one closely since it's the seizure-defense gate and the thing duress features would eventually hang off. **The core design is careful** — `lockGeneration` correctly defeats a Face ID callback resolving after a re-lock, `isLocked` is seeded from the provider so the gate covers *launch* and not just background returns, the fail-open is narrowly scoped to "no device passcode" (a state you can only reach by knowing the passcode), and the toggle refuses to arm without one. That's better than most app-lock implementations I've read.

One real gap, and it's the seizure case the feature exists for:

### Sheets render above the lock

`AppLockScreen` is a ZStack sibling of `ContentView` (`BitchatApp.swift:66-76`), but the six `.sheet`/`.fullScreenCover` presentations are attached *inside* `ContentView` (`ContentView.swift:265, 341, 348, 358, 374, 388`). UIKit presents those at the window level — above the entire SwiftUI root, including the ZStack. `lockIfEnabled()` only flips `isLocked` (`AppLockModel.swift:72-77`); it doesn't dismiss anything, and iOS doesn't dismiss modals on background. I also checked whether `ContentView` closes sheets on `scenePhase` — it has phase logic for alerts, but nothing dismisses a presented sheet.

So: **a sheet open at background time is still open, and on top, when the app returns locked.**

Concrete: settings sheet open (`ContentView.swift:341`) → phone taken / backgrounded → attacker foregrounds it → `AppLockScreen` is behind the still-presented sheet, which exposes hand-added **relay hostnames** (names an operator someone chose to route through — the privacy assessment treats these as sensitive local state), the location/geohash surfaces, and the **panic-wipe button**. The chat timelines themselves are correctly covered, since they live inside `ContentView` — it's specifically the modal layer that escapes.

Two ways to close it, depending on taste:
- dismiss active presentations in `lockIfEnabled()` (cheap, but needs every presentation binding routed somewhere resettable), or
- host the gate above the modal layer — a dedicated `UIWindow` at a higher window level, or a `fullScreenCover` on the scene root rather than a ZStack sibling.

High confidence from the structure, but this is exactly the kind of thing worth confirming on a device before picking a fix — open settings, background, foreground, see what's on top.

### Two honesty notes, both minor

**macOS is materially weaker than the copy implies.** The lock engages at launch only (the `scenePhase` re-lock is inside `#if os(iOS)`), so a Mac left unlocked and walked away from stays readable — `PrivacyScreen` covers window snapshots, not a live window. The setting copy ("after it's been in the background") describes iOS behaviour and is shown on both platforms. Worth a platform-conditional line, or saying plainly that macOS locks only at launch.

**`.deviceOwnerAuthentication` means the device passcode opens it.** Right call for the grabbed-while-unlocked case, and the fail-open reasoning in the comment is sound. But it means the lock offers nothing against a *coerced* unlock, which is a large part of the threat model for the people this app is aimed at. The subtitle is honest about the passcode requirement; I'd avoid any framing that implies protection against someone who can compel the passcode, so the feature doesn't get oversold before a duress mode exists.

None of this is a blocker for the concept — the gate is a real improvement over having none. The sheet layering is the one I'd want fixed before merge, since it defeats the lock in precisely the scenario it's for.

RodsF1:
Opened #1703 against this branch with the fix for the sheet-over-lock gap — a dedicated UIWindow at .alert+1 hosting AppLockScreen, so a sheet open at background time (settings/relay names/panic-wipe) can't sit on top of the lock. Kept the ZStack sibling as the cold-launch cover. It's not unit-testable (no UIWindowScene in the test host), but it compiles for iOS and the macOS suite is green; the device check is in the PR body.

Chessing234 (commented):

RodsF1:
Merge-order note: #1703 targets this branch rather than main, so it'd disappear when this lands — on a squash with the head branch deleted GitHub closes a dependent PR instead of retargeting it (#1655 and #1404 both went that way). This branch is the stack tip, so it's likely the last one you merge.

PR #1665, opened 2026-08-12
fix(ble): reject negative and overlapping offsets in inbound write buffer

kappaone-eth:
## Summary

- Fixes a remotely triggerable crash: a BLE central can send ATT writes with a **negative offset**, which flows through `BLEInboundWriteBuffer.append(chunks:for:capBytes:)` into `Data.replaceSubrange` and traps at runtime, crashing any BitChat peripheral that receives the write (denial of service).
- `BLEInboundWriteBuffer` now validates each chunk before mutating the buffer: rejects `offset < 0`, non-monotonic sequences, and overlapping chunks, and drops the offending per-central buffer (`AppendResult.invalid`).
- The peripheral-role write handler logs and discards malformed writes instead of crashing.

## Details

In `didReceiveWrite`, `CBATTRequest.offset` is attacker-controlled by the remote central. Requests are sorted ascending, but nothing rejects a negative offset:

```swift
combined.replaceSubrange(chunk.offset..<end, with: chunk.data)
```

With `chunk.offset < 0`, the range `chunk.offset..<end` is invalid (and if `end` lands below the buffer start, `replaceSubrange` also traps), so a single crafted write crashes the app. Duplicate or overlapping offsets additionally let a sender silently rewrite previously buffered bytes.

The fix validates each non-empty chunk against a running `lastEnd` cursor before touching the buffer. On violation it clears the per-central buffer and returns a new `.invalid(metadata:)` result; the caller logs the offending offsets and keeps operating. Legitimate CoreBluetooth long writes (ascending, non-overlapping offsets, including zero-padded gaps) are unaffected.

## Tests

Added to `BLEInboundWriteBufferTests`:

- negative offset is rejected and the cleared buffer still decodes a later clean frame
- overlapping chunks are rejected
- non-monotonic chunks are rejected

Note: tests were authored on a machine without an Xcode toolchain, so they have not been executed locally — relying on CI.

### DIFF
--- a/bitchat/Services/BLE/BLEInboundWriteBuffer.swift
+++ b/bitchat/Services/BLE/BLEInboundWriteBuffer.swift
@@ -18,6 +18,7 @@ struct BLEInboundWriteBuffer {
         case decoded(packet: BitchatPacket, metadata: BLEInboundWriteAppendMetadata)
         case waiting(metadata: BLEInboundWriteAppendMetadata)
         case oversized(metadata: BLEInboundWriteAppendMetadata)
+        case invalid(metadata: BLEInboundWriteAppendMetadata)
     }
 
     private var buffersByCentralID: [String: Data] = [:]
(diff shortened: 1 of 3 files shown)


kappaone-eth:
Superseded by a new PR (branch history rewritten for authorship).

PR #1666, opened 2026-08-12
fix(ble): reject negative and overlapping offsets in inbound write buffer

kappaone-eth:
## Summary

- Hardens `BLEInboundWriteBuffer.append(chunks:for:capBytes:)` against malformed chunk sequences: rejects negative offsets, non-monotonic sequences, and overlapping chunks (including overlaps with bytes buffered by an earlier callback), and drops the offending per-central buffer (`AppendResult.invalid`).
- An offset-0 chunk arriving on a non-empty buffer is treated as a **fresh frame replacing stale partial bytes** (clear, then apply). This is intentional: our central role writes `.withoutResponse`, so legit writes always land at offset 0 and a restart must overwrite, not be rejected.
- The peripheral write handler now responds to malformed writes with `CBATTError.Code.invalidOffset` instead of silently ACKing them, so the sending central aborts the transfer instead of retrying into a cleared buffer.

## Details

`CBATTRequest.offset` is attacker-controlled input from the remote central, fed straight into:

```swift
combined.replaceSubrange(chunk.offset..<end, with: chunk.data)
```

Note on reachability: the ATT wire field is 16-bit unsigned, so a negative offset cannot be produced over the air; the `offset < 0` guard is defense-in-depth for the `Int`-based buffer API. The reachable hardening is the overlap/rewrite protection: duplicate or out-of-order offsets previously let a sender silently rewrite already-buffered bytes of a pending frame.

The fix validates each non-empty chunk against a running `lastEnd` cursor (seeded with the already-buffered byte count) before touching the buffer. On violation it clears the per-central buffer and returns `.invalid(metadata:)`; the caller NACKs the requests and keeps operating. Legitimate CoreBluetooth long writes (ascending, non-overlapping offsets, including zero-padded gaps) are unaffected.

Known limitation: CoreBluetooth does not distinguish write-with-response from write-without-response in `didReceiveWrite`, so an offset-0 restart from a `withResponse` multi-chunk transfer retry is indistinguishable from a fresh `withoutResponse` frame; both are treated as a restart (replace).

## Tests

Added to `BLEInboundWriteBufferTests`:

- negative offset is rejected and the cleared buffer still decodes a later clean frame
- overlapping chunks are rejected (within one call and against previously buffered bytes)
- non-monotonic chunks are rejected
- offset-0 write on a stale buffer restarts cleanly

Note: tests were authored on a machine without an Xcode toolchain, so they have not been executed locally â€” relying on CI.

Supersedes #1665 (same change; branch history was rewritten for authorship).

### DIFF
--- a/bitchat/Services/BLE/BLEInboundWriteBuffer.swift
+++ b/bitchat/Services/BLE/BLEInboundWriteBuffer.swift
@@ -18,6 +18,7 @@ struct BLEInboundWriteBuffer {
         case decoded(packet: BitchatPacket, metadata: BLEInboundWriteAppendMetadata)
         case waiting(metadata: BLEInboundWriteAppendMetadata)
         case oversized(metadata: BLEInboundWriteAppendMetadata)
+        case invalid(metadata: BLEInboundWriteAppendMetadata)
     }
 
     private var buffersByCentralID: [String: Data] = [:]
@@ -34,10 +35,28 @@ struct BLEInboundWriteBuffer {
         var combined = buffersByCentralID[centralID] ?? Data()
         var appendedBytes = 0
         var offsets: [Int] = []
+        // New chunks must not overlap bytes already buffered for this central.
+        var lastEnd = combined.count
 
         for chunk in chunks where !chunk.data.isEmpty {
             offsets.append(chunk.offset)
+
+            // Reject malformed writes before touching the buffer: a negative
+            // offset traps `Data.replaceSubrange`, and non-monotonic or
+            // overlapping offsets corrupt previously written bytes.
+            guard chunk.offset >= 0, chunk.offset >= lastEnd else {
+                let metadata = BLEInboundWriteAppendMetadata(
+                    accumulatedBytes: combined.count,
+                    appendedBytes: appendedBytes,
+                    offsets: offsets,
+                    packetType: combined.count >= 2 ? combined[1] : nil
+                )
+                buffersByCentralID.removeValue(forKey: centralID)
+                return .invalid(metadata: metadata)
+            }
+
             let end = chunk.offset + chunk.data.count
+            lastEnd = end
 
             if combined.count < end {
                 combined.append(Data(repeating: 0, count: end - combined.count))
--- a/bitchat/Services/BLE/BLEInboundWriteBuffer.swift
+++ b/bitchat/Services/BLE/BLEInboundWriteBuffer.swift
@@ -41,6 +41,14 @@ struct BLEInboundWriteBuffer {
         for chunk in chunks where !chunk.data.isEmpty {
             offsets.append(chunk.offset)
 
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
the overlap/rewrite hardening is worth having, but two things first.

the negative offset: `CBATTRequest.offset` comes straight off the ATT wire and that field is 16-bit unsigned, so i don't think a central can produce offset < 0 at all. can you show a repro or drop the DoS framing?

bigger one: our central role writes with `.withoutResponse` (BLEService.swift:5073, 5126, 5167), so legit inbound writes land at offset 0. with a stale partial buffer pending, the next offset-0 write now trips `chunk.offset >= lastEnd`, goes `.invalid`, and its bytes are dropped — main applied them in place. can you treat offset 0 with a non-empty buffer as a restart (clear, then apply) instead of a reject?

Chessing234 (commented):
the negative/overlap rejection before mutate is the right place, and clearing the buffer on `.invalid` so a later clean frame can decode is good. the offset-0 reset on a non-empty buffer matches withoutResponse rewrite semantics — worth a one-line note in the PR body that this is intentional so reviewers don't treat it as a silent data drop.

Chessing234 (commented):
offset-0 restart on a non-empty buffer treats every fresh withoutResponse write as a replace — good — but a withResponse multi-chunk transfer that legitimately reuses offset 0 after a partial failure isn't distinguished. more important: `.invalid` clears the buffer and returns, yet the peripheral write handler still needs to respond to the ATT request (success vs error). check the call site sends the right CBATT error instead of silently acking a poisoned write; otherwise the central keeps sending into a cleared buffer and you only log.

kappaone-eth:
Thanks for the careful review — both points addressed in the latest commits.

**Negative offset framing:** agreed, the ATT wire field is 16-bit unsigned, so `offset < 0` can't arrive over the air — I have no repro and have dropped the DoS framing from the PR body. The guard stays as defense-in-depth for the `Int`-based buffer API; the reachable part of this PR is the overlap/rewrite hardening, which is how the body now frames it.

**ATT response on poisoned writes:** fixed in `101db18`. `didReceiveWrite` now responds per central group after validation: `.invalidOffset` for requests that fail offset checks, `.success` otherwise. Responses still happen inside the same callback, so the sub-millisecond ATT timing requirement is preserved. A central sending bad offsets now gets an explicit abort instead of a silent ACK into a cleared buffer.

**Offset-0 restart:** implemented in `b74e929` and noted as intentional in the PR body, as you suggested. One caveat you flagged stands: CoreBluetooth doesn't tell us whether a write was with or without response, so a `withResponse` multi-chunk retry at offset 0 is indistinguishable from a fresh `withoutResponse` frame — both take the restart path. That's documented in the PR body's limitations section.

Issue #1658, opened 2026-08-11
Dragging on a playing voice note in a DM can leave the conversation


heyaim:
A playing voice note's waveform accepts a seek: `WaveformView` overlays the bars with `DragGesture(minimumDistance: 0)` and seeks to the release point (`WaveformView.swift:55`), active while the note plays (`VoiceNoteView.swift:71`). In a private chat that waveform renders inside the message list, and the sheet attaches its swipe-right-to-leave gesture to the same list with `.highPriorityGesture` (`ContentSheetViews.swift:575`). A high-priority ancestor drag starves descendant gestures. A drag across the waveform never reaches the seek, and a rightward drag that ends past the leave thresholds (horizontal over 80, vertical under 60, `ContentSheetViews.swift:613`) leaves the conversation while the note is playing.

The same mechanism shows up in two merged PRs:

- #1405 fixed it as the cause: the composer mic's press-and-hold was cancelled 3 to 10 ms after touch-down, and the fix moved the swipe off the whole sheet and onto the message list.
- #1402 hardened against it: the grey-box root cause was a status default, and the PR moved the reveal tap onto a Button as defense against this same gesture; its description notes it starves descendant TapGestures while Button actions survive.

The comment at the attachment site records the mechanism:

```swift
// Swipe-right-to-leave lives on the message list only. On the
// whole sheet it preempted the composer's press-and-hold mic
// gesture (a high-priority ancestor drag cancels child gestures
// within milliseconds — same starvation as the image-reveal bug).
```

After #1405 the app has two drag gestures attached with plain `.gesture`: the composer mic hold (`ContentComposerView.swift:347`), which that fix moved out of the swipe's reach, and this seek, which sits inside it.

### Measurements

**On the real DM screen, in the simulator, with an injected conversation.** A mesh DM with a voice note normally needs two devices. To exercise the real view stack with one, I used a local debug build whose only change seeds a received voice note from a fake peer at launch, through the same handler the mesh receive path uses (`ChatPrivateConversationCoordinator.handlePrivateMessage`), and opens the conversation. Every view under test is stock. With the note playing, a rightward drag of about 130 points across the waveform closed the conversation. Simulator audio would not start on later runs; the drag-never-seeks half is covered by the harness below.

[...]

PR #1668, opened 2026-08-14
fix: stand down swipe-to-leave while a voice note is playing

Chessing234:
Refs #1658.

## What happens

`WaveformView` overlays the bars with a seek `DragGesture(minimumDistance: 0)`
(`WaveformView.swift:55`), live while the note plays — `VoiceNoteView.swift:71`
passes `isInteractive: playback.isPlaying`. In a private chat that waveform
renders inside the message list, and the sheet attaches swipe-right-to-leave to
that same list with `.highPriorityGesture` (`ContentSheetViews.swift:575`).

A high-priority ancestor drag starves descendant gestures, so two things follow:
a scrub across the waveform never reaches the seek, and a scrub that happens to
travel past `horizontal > 80, vertical < 60` ends the conversation while the note
is still playing.

The issue notes this is the third appearance of the same starvation: #1405 fixed
it as a cause (the mic press-and-hold was cancelled 3-10 ms after touch-down, and
the swipe moved off the whole sheet onto the message list), and #1402 hardened
against it (the reveal moved onto a `Button`, because Button actions survive
starvation while descendant gestures do not). The comment at the attachment site
already records the mechanism.

## Fix

While audio is audible the scrub is far likelier to be the reader's intent than
leaving, so the leave gesture stands down for the duration of playback.

Two supporting changes keep that decision honest rather than buried in a view:

- `PrivateChatSwipeToLeavePolicy` holds the 80/60 thresholds and the playback
  guard, so the decision is testable without a gesture recogniser.
- `VoiceNotePlaybackCoordinator` — which already exists to enforce
  single-playback and already exposes an `internal init()` "so tests can isolate
  their own exclusivity slot" — gains `hasActivePlayback`. No new state, no new
  ownership; it is the object that already knows.

**Deliberate behaviour change, stated plainly:** while a voice note plays, the
swipe gesture will not leave the conversation. Leaving stays available from the
sidebar. That is the trade this fix makes, and it is exactly the reported harm
("leaves the conversation while the note is playing") — but it is a change, not
a pure bug fix, so please push back if you'd rather the seek be rescued a
different way (e.g. moving the seek onto a control that survives starvation, the
#1402 route).

## Test plan

- [x] `swift build --build-tests` — clean; no new warnings (the four
      `no calls to throwing functions` warnings in `BLEAnnounceHandlerTests` and
      the unhandled-resource warning are pre-existing on `main`).
- [x] `swift test --skip-build --parallel` — **2032 tests in 221 suites passed**.
      Baseline on `main` before this change: 2018 tests in 219 suites, also green.
      The delta is exactly the 14 new tests in 2 new suites.
- [x] Verified the regression tests have teeth: deleting the
      `guard !isVoiceNotePlaying` line fails exactly the two suppression tests
      and nothing else.
- [ ] Gesture composition itself is **not** covered. `swift test` cannot exercise
      SwiftUI gesture arbitration, so "the seek now actually receives the drag"
      is verified by reading `.highPriorityGesture` semantics and the two prior
      PRs, not by execution. A simulator UI test would be needed to prove it, and
      I have not written one.

New coverage, 14 tests:

- thresholds preserved — decisive drag leaves; sub-threshold, leftward, and
  steep drags do not; vertical is judged on magnitude so a shallow upward drag
  still leaves
- the regression — a qualifying drag is suppressed while playing, resumes once
  stopped, and playback never *invents* a leave from a drag that never qualified
- `hasActivePlayback` — idle is false, activate/deactivate flip it, a bare
  `reserve` does **not** count as playing (an async starter that never becomes
  audible must not suppress the gesture), a foreign `deactivate` is ignored, and
  a takeover pauses the previous holder while staying active

### DIFF
--- a/bitchat/Features/voice/VoiceNotePlaybackController.swift
+++ b/bitchat/Features/voice/VoiceNotePlaybackController.swift
@@ -342,6 +342,15 @@ final class VoiceNotePlaybackCoordinator {
         return true
     }
 
+    /// True while some controller holds the playback slot.
+    ///
+    /// Read by the private-chat swipe-to-leave gesture, which stands down while
+    /// a voice note is audible so a waveform seek is not starved by the
+    /// high-priority ancestor drag. See `PrivateChatSwipeToLeavePolicy`.
+    var hasActivePlayback: Bool {
+        activeController != nil
+    }
+
     func isCurrent(_ reservation: Reservation, for controller: any ExclusivePlayback) -> Bool {
         latestReservation == reservation && latestReservedController === controller
     }
--- a/bitchat/Features/voice/VoiceNotePlaybackController.swift
+++ b/bitchat/Features/voice/VoiceNotePlaybackController.swift
@@ -113,6 +113,11 @@ final class VoiceNotePlaybackController: NSObject, ObservableObject, AVAudioPlay
         updateProgress()
         isPlaying = false
         releaseSession()
+        // The slot means "audio is audible", not "a row is selected": a paused
+        // note has nothing to yield to the next playback, and readers of
+        // `hasActivePlayback` would otherwise see playback that stopped being
+        // heard when the user paused it.
+        exclusivity.deactivate(self)
     }
 
     func stop() {
@@ -296,7 +301,7 @@ extension VoiceNotePlaybackController: ExclusivePlayback {
 }
 
 /// Ensures only one voice playback (note or live burst) runs at a time.
-final class VoiceNotePlaybackCoordinator {
+final class VoiceNotePlaybackCoordinator: ObservableObject {
     static let shared = VoiceNotePlaybackCoordinator()
 
     struct Reservation: Equatable {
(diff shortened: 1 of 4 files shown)


Chessing234:
both findings are right, and the first one is the important half — the onEnded guard stopped the conversation from being left but did nothing for the scrub, because an armed highPriorityGesture starves its descendants at recognition time.

2da25a9: the gesture is attached with `including: playbackCoordinator.hasActivePlayback ? .subviews : .all`, so while a note is audible recognition belongs to the waveform's seek drag. the onEnded guard stays for a drag already in flight when playback starts.

ca15f66: pause() now releases the slot. it was leaving activeController set, so hasActivePlayback meant "a row was played at some point" and the gesture stayed suppressed after a pause until the row went away. a paused note has nothing to yield to the next playback, so the slot is the wrong thing for it to hold. hasActivePlayback is stored and published now, which is also what lets the view react rather than sample it once.

verification, plainly: xcode is not on this machine any more (command line tools only), so swift-testing does not build and i could not run the suite locally this time — the app target does type-check with both changes. the added coordinator test covers pause releasing the slot; CI runs it. gesture arbitration itself is still not unit-testable, as noted in the PR body.

Chessing234:
awaiting-review

PR #1669, opened 2026-08-14
fix: let a resubscribing central recover from the announce rate limit

Chessing234:
Found by reading `BLESubscriptionAnnounceLimiter`, not from an issue.

## The bug

`decision(for:now:)` rebuilds the entry on a **rejected** attempt with
`lastAnnounceTime: now`:

```swift
states[centralID] = State(
    lastAnnounceTime: now,          // <- refreshed even though nothing was announced
    attemptCount: newAttemptCount,
    currentBackoffSeconds: newBackoff
)
```

That field is the reference for *both* comparisons in this type — the backoff
check (`now.timeIntervalSince(existing.lastAnnounceTime)`) and
`pruneStaleEntries`. Refreshing it on rejection means the elapsed time is reset
by the very attempt that was refused, so for a central that keeps resubscribing:

- elapsed never grows past `currentBackoffSeconds`, which has meanwhile been
  doubled up to the 30 s cap, so every subsequent attempt is refused;
- `attemptCount` climbs past `bleSubscriptionRateLimitMaxAttempts` (5), so
  `suppressAnnounce` latches true;
- and `pruneStaleEntries` never fires either, because it measures staleness from
  the same refreshed timestamp.

The result is that the central is suppressed for as long as it keeps trying, with
no path back.

## Why it matters beyond the attack case

The limiter is the BCH-01-004 enumeration defence, and against a hammering
attacker the behaviour looks like a feature. But a legitimate central with a
flapping BLE link resubscribes exactly the same way — `didSubscribeTo` fires on
every reconnect. Once it crosses 5 attempts we stop announcing to it
(`BLEService+LinkLayerPeripheralRole.swift:151-156` returns early), so **we stay
invisible to that peer until the app restarts or it stops reconnecting for a
full window** — and it cannot stop reconnecting, because the link is flapping.

## The fix

Keep `lastAnnounceTime` at the last *allowed* announce. That is what
`TransportConfig` already documents it as — "Minimum interval between announces
per central" — and it makes the backoff mean what it says.

Escalation is deliberately unchanged: `attemptCount` still climbs, the backoff
still doubles to the 30 s cap, and `suppressAnnounce` still engages at 5
attempts. The only change is that elapsed time is now measured from a fixed
point, so a central is re-admitted once the capped backoff has genuinely passed
— i.e. at most one announce per 30 s for someone hammering, instead of none ever.

## Test plan

- [x] `swift test --parallel` — **2020 tests in 219 suites passed**. Baseline on
      `main`: 2018 in 219, also green. The delta is exactly the 2 tests added here.
- [x] Both new tests were written **before** the fix and confirmed to fail against
      unmodified `main`:
      `Expectation failed: (admittedAt → nil) != nil` — after 60 s of one
      resubscribe per second the central was still
      `.rateLimited(backoffSeconds: 30.0, attemptCount: 120, suppressAnnounce: true)`.
- [x] Teeth re-checked after the fix by reverting only the source file and
      re-running: the recovery test fails, the hold-off test still passes, and the
      three pre-existing tests are unaffected either way.
- [x] No new warnings.

New coverage:

- *a central that keeps resubscribing recovers once the backoff elapses* — the
  regression itself, and it asserts re-admission happens **at or after** the
  capped backoff, not merely that it happens.
- *a hammering central is still held off for the whole backoff* — the
  counterpart, so the fix cannot be "corrected" into admitting early and
  weakening the enumeration defence.

The three existing tests (first-allowed/then-limited, suppression threshold,
stale pruning) are untouched and still pass.

### DIFF
--- a/bitchat/Services/BLE/BLESubscriptionAnnounceLimiter.swift
+++ b/bitchat/Services/BLE/BLESubscriptionAnnounceLimiter.swift
@@ -41,8 +41,14 @@ struct BLESubscriptionAnnounceLimiter {
             existing.currentBackoffSeconds * TransportConfig.bleSubscriptionRateLimitBackoffFactor,
             TransportConfig.bleSubscriptionRateLimitMaxBackoffSeconds
         )
+        // `lastAnnounceTime` deliberately keeps the time of the last *allowed*
+        // announce. It is the reference for both the backoff and
+        // `pruneStaleEntries`, so refreshing it on a rejection restarts the
+        // tracking window on every attempt — a central resubscribing faster
+        // than the maximum backoff could then never reach the window and
+        // stayed suppressed for as long as it kept trying.
         states[centralID] = State(
-            lastAnnounceTime: now,
+            lastAnnounceTime: existing.lastAnnounceTime,
             attemptCount: newAttemptCount,
             currentBackoffSeconds: newBackoff
         )
--- a/bitchatTests/Services/BLESubscriptionAnnounceLimiterTests.swift
+++ b/bitchatTests/Services/BLESubscriptionAnnounceLimiterTests.swift
@@ -58,4 +58,50 @@ struct BLESubscriptionAnnounceLimiterTests {
         #expect(limiter.decision(for: freshCentralID, now: afterWindow) == .allowed)
         #expect(limiter.trackedCentralCount == 1)
     }
+
+    @Test("a central that keeps resubscribing recovers once the backoff elapses")
+    func flappingCentralRecoversAfterBackoff() {
+        // A legitimate central whose BLE link flaps resubscribes faster than the
+        // maximum backoff. Every rejection used to refresh `lastAnnounceTime`,
+        // which is the reference for both the backoff and `pruneStaleEntries`,
+        // so the elapsed time never grew: the central stayed suppressed for as
+        // long as it kept trying. It must instead be re-admitted once the
+        // capped backoff has passed since its last *allowed* announce.
+        var limiter = BLESubscriptionAnnounceLimiter()
+        let centralID = "central-flapping"
+        let start = Date(timeIntervalSince1970: 1_000)
+
+        #expect(limiter.decision(for: centralID, now: start) == .allowed)
+
+        let maxBackoff = TransportConfig.bleSubscriptionRateLimitMaxBackoffSeconds
+        var admittedAt: Double?
+        for step in stride(from: 1.0, through: maxBackoff * 2, by: 1.0) {
+            if limiter.decision(for: centralID, now: start.addingTimeInterval(step)) == .allowed {
+                admittedAt = step
+                break
+            }
+        }
+
+        #expect(admittedAt != nil, "a central that keeps trying must eventually be re-admitted")
+        if let admittedAt {
+            #expect(admittedAt >= maxBackoff, "re-admitted early at +\(admittedAt)s")
+        }
+    }
+
+    @Test("a hammering central is still held off for the whole backoff")
+    func flappingCentralIsNotAdmittedEarly() {
+        // The counterpart: relaxing the reference timestamp must not let a
+        // hammering central back in before the capped backoff has elapsed.
+        var limiter = BLESubscriptionAnnounceLimiter()
+        let centralID = "central-flapping"
+        let start = Date(timeIntervalSince1970: 2_000)
+        let maxBackoff = TransportConfig.bleSubscriptionRateLimitMaxBackoffSeconds
+
+        #expect(limiter.decision(for: centralID, now: start) == .allowed)
+
+        for step in stride(from: 1.0, to: maxBackoff, by: 1.0) {
+            let decision = limiter.decision(for: centralID, now: start.addingTimeInterval(step))
+            #expect(decision != .allowed, "admitted early at +\(step)s")
+        }
+    }
 }


bitchat/Services/BLE/BLESubscriptionAnnounceLimiter.swift
```diff
@@ -41,8 +41,14 @@ struct BLESubscriptionAnnounceLimiter {
             existing.currentBackoffSeconds * TransportConfig.bleSubscriptionRateLimitBackoffFactor,
             TransportConfig.bleSubscriptionRateLimitMaxBackoffSeconds
         )
+        // `lastAnnounceTime` deliberately keeps the time of the last *allowed*
```
jozanek, line 44: This comment explains the mechanism clearly. Would you be open to letting it carry the history too? `9cd955a` set this to `now` on purpose seventeen minutes after `99896bc` shipped the remediation — its inline comment read `// Reset timer on each blocked attempt` — and `193cfdc` dropped that rationale during the extraction, which is what leaves the line looking like a slip today.

Naming both commits here would give the next reader what I had to reconstruct from `git log`, and would stop this flipping back a third time.

bitchat/Services/BLE/BLESubscriptionAnnounceLimiter.swift
```diff
@@ -41,8 +41,14 @@ struct BLESubscriptionAnnounceLimiter {
...
+        // announce. It is the reference for both the backoff and
+        // `pruneStaleEntries`, so refreshing it on a rejection restarts the
+        // tracking window on every attempt — a central resubscribing faster
+        // than the maximum backoff could then never reach the window and
+        // stayed suppressed for as long as it kept trying.
         states[centralID] = State(
-            lastAnnounceTime: now,
+            lastAnnounceTime: existing.lastAnnounceTime,
```
jozanek, line 51: One consequence worth recording somewhere: with `lastAnnounceTime` pinned, `pruneStaleEntries` now measures from a timestamp that does not move during an attempt stream. Today that is harmless because `maxBackoff` (30) < `window` (60), so re-admission always fires first. If the cap were ever raised to or above the window, prune would fire first and the effective cap would silently become `min(maxBackoff, window)`.

I found the relationship stated nowhere, and all five constants are untouched since `99896bc`. Would a line on `TransportConfig.swift:290-291` be welcome?

bitchatTests/Services/BLESubscriptionAnnounceLimiterTests.swift
```diff
@@ -58,4 +58,50 @@ struct BLESubscriptionAnnounceLimiterTests {
...
+        for step in stride(from: 1.0, through: maxBackoff * 2, by: 1.0) {
+            if limiter.decision(for: centralID, now: start.addingTimeInterval(step)) == .allowed {
+                admittedAt = step
+                break
+            }
+        }
+
+        #expect(admittedAt != nil, "a central that keeps trying must eventually be re-admitted")
+        if let admittedAt {
+            #expect(admittedAt >= maxBackoff, "re-admitted early at +\(admittedAt)s")
```
jozanek, line 87: `>= maxBackoff` is a lower bound, and the loop searches out to `maxBackoff * 2`. An implementation anchoring elapsed at the first rejection (admits at +31 s), or preserving the timestamp only once suppressed (+33 s), would pass this.

Would an equality at the cap work for you instead? It would also pin the `elapsed < backoff` boundary at exactly 30 s, which the counterpart test's `to: maxBackoff` currently stops one second short of.

bitchatTests/Services/BLESubscriptionAnnounceLimiterTests.swift
```diff
@@ -58,4 +58,50 @@ struct BLESubscriptionAnnounceLimiterTests {
...
+        }
+    }
```
jozanek, line 89: The description's security claim — one announce per 30 s for a hammering central rather than none — is the number a maintainer weighing this against BCH-01-004 needs to trust, and both tests stop at the first admission.

Would you be open to a third test driving 1 Hz attempts across 90 s and asserting admissions at exactly t = 0, 30, 60? It pins the steady-state rate, and it covers the ladder reset that happens on each re-admission.

jozanek (commented):
Thank you for this, and for writing the tests first and re-checking their teeth by reverting only the source file.

I agree the defect is real: elapsed time is measured against a field that the rejected attempt itself moves, so a central that trips the ladder has no exit while it keeps trying, and `pruneStaleEntries` reads the same field so it cannot rescue it. The fix looks right to me.

One piece of history changes how I would frame it. `99896bc` — the BCH-01-004 remediation itself — shipped with `existingState.lastAnnounceTime` on the reject path, exactly what this PR restores. Seventeen minutes later, `9cd955a` changed it to `now` on purpose, with an inline `// Reset timer on each blocked attempt`:

> Previously, the backoff timer was only based on the last successful announce, allowing attackers to receive an announce as soon as the initial backoff expired regardless of how many blocked attempts occurred in between.

`193cfdc` then extracted the struct, kept the behaviour, and dropped both comments. I do not think you could have known this from the code; I could not either until I ran `git log -S`.

So this is a revert of a considered decision on an audit-remediation path. Would you be willing to name it in the description, so @jackjackbits can re-make the call with the trade-off visible? I think the revert is right, for what it is worth: it moves a 1 Hz hammerer from one announce ever to 2 per 60 s window against a documented intent of 5, while an attacker pacing at exactly 2.0 s never enters the ladder and collects ~30 either way.

Two corrections I would suggest:

1. "Invisible until the app restarts" does not hold. `addSubscribedCentral` runs before the limiter check and `sendAnnounce` broadcasts rather than directing at the subscriber, so a suppressed central keeps receiving the maintenance announce (every 4 s with no peers, 15±4 s sparse). What it loses is the announce timed to land inside a short connected window.

2. Reaching rejection #1 needs two subscribes under 2.0 s apart, and an RF-flap disconnect carries an error, which arms the 3.0 s `bleDisconnectDiscoveryIgnoreSeconds` settle. "Rare to enter, impossible to leave" is a stronger argument than "common", and I would lead with it.

**What I could not verify:** how often a real peer produces error-free resubscribes under 2 s — nothing in the repo settles it, and it is the fact that sizes the benefit. I also ran no builds or tests.

**What would move me to approve:** the description corrected on both points, plus the steady-state test noted inline.

Three smaller things I am not holding up for: re-admission resets `attemptCount`, so the BCH-01-004 log no longer separates a flapping link from a sustained adversary; `SubscriptionRateLimitTests.swift` still claims "at most maxAttempts announces per window", untrue in either direction since January; and the limiter keys on a rotating `central.identifier`, which may let an attacker skip the ladder entirely — worth an issue rather than a line in your diff.

Chessing234:
awaiting-review

PR #1671, opened 2026-08-15
feat(ble): negotiate an explicit per-peer fragment ceiling

Chessing234:
Closes the `TODO(#1434)` left in `broadcastPacket` when the blanket 256-fragment preflight was narrowed.

## The gap

Today the fragment bound for a directed transfer is inferred from the packet type: a directed `fileTransfer` is assumed to be the raw migration fallback aimed at current Android (256), and anything else is assumed to be aimed at a client whose reassembler is as large as ours. That proxy holds only while "implements the encrypted `0x20` path" and "has a large reassembler" describe the same population.

A client that adopts `0x20` with a smaller buffer breaks it in the silent direction: we plan fragments the peer drops, and the sender sees no local failure. `.privateMedia` says a peer understands encrypted media; it says nothing about how much of it that peer can hold.

## The change

- **TLV `0x03` on `AuthenticatedPeerStatePacket`** — a 2-byte big-endian count of fragments the peer will reassemble for one packet, stated inside the established Noise session. Absent means "not advertised", which is what every released client sends, so old peers decode unchanged.
- An unreadable `0x03` (zero, wrong width, duplicated) is **rejected rather than skipped** like an unknown TLV. Skipping would read as "said nothing", and "said nothing" falls back to the permissive proxy — the opposite of what a peer sending the field is asking for.
- **`BLEFragmentCeilingPolicy`** turns the inline type check into a value: packet type + directedness + whatever the peer advertised, out comes the bound and which of three reasons produced it. Framework-free, so every branch is tested directly.
- An advertised ceiling replaces the proxy **in both directions**. Below 256 is the case the proxy gets wrong silently. Above 256 is a peer telling us the deployed-Android assumption behind that cap does not describe it.
- Every decision is clamped to what we would reassemble ourselves, so one side raising its configuration cannot raise the other side's memory exposure.
- We advertise our own bound too. The receiver's `10_000` guard is now a named constant used for both the guard and the advertisement, so they cannot drift.

## Two things worth reviewing closely

**The preflight is no longer restricted to `fileTransfer`.** It could not stay restricted and still do its job: an advertised ceiling constrains encrypted media too, and that is the case the proxy misses. Peers that advertise nothing keep the exact old classification, so this is a no-op for every released client.

**The rejection copy is now parameterized.** It said "more than 256 mesh fragments", which stops being true once the ceiling is negotiated. The existing key is kept and only the numeral replaced, so all 30 locales keep their existing translation. One judgment call I want flagged: Turkish attached a vowel-harmonised ablative suffix to the numeral (`256'dan`), which no format specifier gets right for every value, so I rephrased that locale to a suffix-free construction (`mesh parça sınırı: %lld`). That is my edit, not a translator's — please have someone who reads Turkish check it.

## Verification

- Full SwiftPM suite: **2018 → 2034**, green. The delta is exactly the 16 tests added here (4 wire-format, 9 policy, 3 session-store).
- The rekey invariant is enforced twice over — `beginAuthenticatedGeneration` drops the observation, and the accessor refuses one pinned to a superseded generation. Removing either alone still passes; the test fails once both are gone, and it was the **only** failure in that run.
- Two unrelated suites (`ChatViewModelInitializationTests`, `BLEServiceCoreTests`) failed once each on runs that took 47s and 63s while the machine was loaded, and did not reproduce in four subsequent clean runs of the full suite or in isolation. Timeout-sensitive, not this diff — but I am flagging them rather than filtering them out.
- No local SwiftLint, so I checked the `--strict` rules by hand against `.swiftlint.yml`; one `trailing_comma` violation found and fixed before pushing.

## What I did not verify

No device or simulator run, so the negotiation is verified by unit tests and by reading the wiring, not by two real handsets agreeing on a ceiling. Android does not send `0x03` yet, so in the field today every peer takes the unchanged fallback path — which is the safe way for this to land, but it also means the negotiated branch has no cross-platform exercise until an Android client advertises one.

### DIFF
--- a/bitchat/Protocols/Packets.swift
+++ b/bitchat/Protocols/Packets.swift
@@ -165,6 +165,7 @@ struct AnnouncementPacket {
 /// `[version=0x01][type][length][value]...`
 /// - TLV `0x01`: canonical minimal little-endian `PeerCapabilities`
 /// - TLV `0x02`: 32-byte Ed25519 signing public key
+/// - TLV `0x03`: optional 2-byte big-endian reassembly fragment ceiling
 ///
 /// Unknown TLVs are skipped for forward compatibility. Unknown versions,
 /// duplicates, non-canonical capability fields, and malformed lengths are
@@ -172,19 +173,49 @@ struct AnnouncementPacket {
 struct AuthenticatedPeerStatePacket: Equatable {
     static let currentVersion: UInt8 = 1
     static let signingPublicKeyLength = 32
+    static let maxReassemblyFragmentsLength = 2
 
     let capabilities: PeerCapabilities
     let signingPublicKey: Data
+    /// How many BLE fragments this peer will reassemble for one packet, as it
+    /// reported inside the established Noise session.
+    ///
+    /// The `.privateMedia` capability bit says a peer understands encrypted
+    /// media; it says nothing about how much of it that peer can hold. Before
+    /// this TLV existed, senders inferred the ceiling from the packet type —
+    /// 256 for the directed migration fallback, the full local ceiling for
+    /// anything encrypted — which is only correct while every client that
+    /// implements `0x20` also has a large reassembler.
+    ///
+    /// `nil` means the peer did not advertise one, which is the case for every
+    /// client released before this TLV. Senders fall back to the type proxy
+    /// there; see `BLEFragmentCeilingPolicy`.
+    let maxReassemblyFragments: UInt16?
+
+    init(
+        capabilities: PeerCapabilities,
+        signingPublicKey: Data,
+        maxReassemblyFragments: UInt16? = nil
+    ) {
+        self.capabilities = capabilities
+        self.signingPublicKey = signingPublicKey
+        self.maxReassemblyFragments = maxReassemblyFragments
+    }
 
     private enum TLVType: UInt8 {
         case capabilities = 0x01
         case signingPublicKey = 0x02
+        case maxReassemblyFragments = 0x03
     }
 
     func encode() -> Data? {
         guard signingPublicKey.count == Self.signingPublicKeyLength else { return nil }
         let capabilityBytes = capabilities.encoded()
         guard !capabilityBytes.isEmpty, capabilityBytes.count <= 8 else { return nil }
+        // Zero would advertise a peer that can reassemble nothing, which is
+        // indistinguishable in effect from refusing every fragmented packet.
+        // Omit the TLV instead of putting a meaningless number on the wire.
+        if let maxReassemblyFragments { guard maxReassemblyFragments > 0 else { return nil } }
 
         var data = Data([Self.currentVersion])
         data.append(TLVType.capabilities.rawValue)
@@ -193,6 +224,11 @@ struct AuthenticatedPeerStatePacket: Equatable {
         data.append(TLVType.signingPublicKey.rawValue)
         data.append(UInt8(signingPublicKey.count))
         data.append(signingPublicKey)
+        if let maxReassemblyFragments {
+            data.append(TLVType.maxReassemblyFragments.rawValue)
+            data.append(UInt8(Self.maxReassemblyFragmentsLength))
+            data.append(contentsOf: withUnsafeBytes(of: maxReassemblyFragments.bigEndian) { Data($0) })
+        }
         return data
     }
 
@@ -202,6 +238,7 @@ struct AuthenticatedPeerStatePacket: Equatable {
         var offset = 1
         var capabilities: PeerCapabilities?
         var signingPublicKey: Data?
+        var maxReassemblyFragments: UInt16?
 
         while offset < data.count {
             guard offset + 2 <= data.count else { return nil }
@@ -228,13 +265,26 @@ struct AuthenticatedPeerStatePacket: Equatable {
                 guard signingPublicKey == nil,
                       value.count == Self.signingPublicKeyLength else { return nil }
                 signingPublicKey = value
+
+            case .maxReassemblyFragments:
+                // Rejected rather than skipped: a peer that meant to constrain
+                // us but sent a field we cannot read must not be treated as
+                // having said nothing, because "said nothing" falls back to
+                // the permissive type proxy.
+                guard maxReassemblyFragments == nil,
+                      value.count == Self.maxReassemblyFragmentsLength else { return nil }
+                let decoded = (UInt16(value[value.startIndex]) << 8)
+                    | UInt16(value[value.startIndex + 1])
+                guard decoded > 0 else { return nil }
+                maxReassemblyFragments = decoded
             }
         }
 
         guard let capabilities, let signingPublicKey else { return nil }
         return AuthenticatedPeerStatePacket(
             capabilities: capabilities,
-            signingPublicKey: signingPublicKey
+            signingPublicKey: signingPublicKey,
+            maxReassemblyFragments: maxReassemblyFragments
         )
     }
 }
--- a/bitchat/Services/BLE/BLEPrivateMediaSessionStore.swift
+++ b/bitchat/Services/BLE/BLEPrivateMediaSessionStore.swift
@@ -5,6 +5,22 @@ struct BLEAuthenticatedPeerStateObservation {
     let fingerprint: String
     let sessionGeneration: UUID
     let capabilities: PeerCapabilities
+    /// The peer's own reassembly bound, if it advertised one. Bound to the
+    /// generation like every other field here: a replacement session must
+    /// re-state its ceiling rather than inherit the previous one.
+    let maxReassemblyFragments: UInt16?
+
+    init(
+        fingerprint: String,
+        sessionGeneration: UUID,
+        capabilities: PeerCapabilities,
+        maxReassemblyFragments: UInt16? = nil
+    ) {
+        self.fingerprint = fingerprint
+        self.sessionGeneration = sessionGeneration
+        self.capabilities = capabilities
+        self.maxReassemblyFragments = maxReassemblyFragments
+    }
 }
 
 struct BLEPrivateMediaProofTimeoutMarker {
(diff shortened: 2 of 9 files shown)


Chessing234:
one allocation note while this is open, since i only spotted it reading the rotation spec.

`docs/PEER-ID-ROTATION.md` §4.5 says the peer-ID binding proof will be "sent as a new TLV in `AuthenticatedPeerStatePacket`" but never assigns it a number. this PR takes `0x03` there, and the android half is permissionlesstech/bitchat-android#879. so whichever of those lands first should be treated as pinning `0x03`, and the binding proof wants `0x04`.

worth writing the number down somewhere either way — the packet's TLV space currently has its assignments only in the struct itself, which is how two features end up reaching for the same byte. (the `0x03` in that doc's §4.6 table is the v2 announce's own namespace, not this one — different packet, no conflict.)

Chessing234:
awaiting-review

bitchat/Services/BLE/BLEService.swift
```diff
@@ -4234,7 +4243,12 @@ extension BLEService {
         let capabilities = localIdentityState.snapshot().advertisedCapabilities
         let state = AuthenticatedPeerStatePacket(
             capabilities: capabilities,
-            signingPublicKey: noiseService.getSigningPublicKeyData()
+            signingPublicKey: noiseService.getSigningPublicKeyData(),
+            // Our own reassembler's bound, so a peer sizing a transfer for us
+            // does not have to infer it from the packet type either.
+            maxReassemblyFragments: UInt16(
```
jozanek, line 4249: This advertises 10,000, but that constant is a header sanity check on `total` — what decides whether we keep a stream is bytes: `assemblyLimit(for:)` returns `maxFramedFileBytes` ≈ 1,179,757, and `:104-108` evicts on projected bytes.

At Android's 469 B fragment size that's ~2,515 fragments, so a peer that believes our 10,000 and sends 4,000 gets dropped silently — the same failure this PR removes in the other direction.

Could the advertised count come from the byte budget instead?

bitchat/Protocols/Packets.swift
```diff
@@ -228,13 +265,26 @@ struct AuthenticatedPeerStatePacket: Equatable {
                 guard signingPublicKey == nil,
                       value.count == Self.signingPublicKeyLength else { return nil }
                 signingPublicKey = value
+
+            case .maxReassemblyFragments:
+                // Rejected rather than skipped: a peer that meant to constrain
+                // us but sent a field we cannot read must not be treated as
+                // having said nothing, because "said nothing" falls back to
+                // the permissive type proxy.
+                guard maxReassemblyFragments == nil,
```
jozanek, line 274: This returns nil for the whole packet, which also carries the capability bitfield and the signing key. `BLEService.swift:4259` then returns early, the watchdog expires, and `privateMediaSendPolicy` lands on `.blockedDowngrade` or `.legacyRequiresConsent` (`:1226-1229`) — the relay-readable plaintext path.

So I read the fallback as below the permissive proxy rather than above it, and `PRIVATE-MEDIA-MIGRATION.md:24` states unknown TLVs are skipped.

Would you be open to skipping the field and using 256 for a peer that sent a malformed one?

bitchat/Services/BLE/BLEPrivateMediaSessionStore.swift
```diff
@@ -347,6 +365,22 @@ final class BLEPrivateMediaSessionStore: @unchecked Sendable {
...
 extension BLEPrivateMediaSessionStore {
+    /// The recipient's advertised reassembly ceiling, or nil when it did not
+    /// advertise one.
+    ///
+    /// Only honoured for an observation pinned to the peer's *current*
+    /// generation. A ceiling read off a superseded session is a statement by
+    /// whoever held that session, and after a rekey the counterpart may be a
+    /// different client entirely — falling back to the type proxy is the
+    /// conservative answer, not reusing a stale number.
+    func negotiatedFragmentCeiling(for peerID: PeerID) -> UInt16? {
```
jozanek, line 376: This checks the generation pin but not `capabilities`, though the observation carries both. A peer advertising `10000` without the `.privateMedia` bit resolves to `.legacyRequiresConsent` (`BLEService.swift:1220`) and reaches the unencrypted send at `:1843` — which routes through this same preflight, so the advertisement lifts the 256 cap on the one path relays can read in the clear.

Would you consider letting a negotiated ceiling raise the bound only on the `0x20` path? Lowering would still work everywhere.

bitchat/Protocols/Packets.swift
```diff
@@ -165,26 +165,57 @@ struct AnnouncementPacket {
 /// `[version=0x01][type][length][value]...`
 /// - TLV `0x01`: canonical minimal little-endian `PeerCapabilities`
 /// - TLV `0x02`: 32-byte Ed25519 signing public key
+/// - TLV `0x03`: optional 2-byte big-endian reassembly fragment ceiling
```
jozanek, line 168: This line and the enum case are the only record that `0x03` is taken. `PRIVATE-MEDIA-MIGRATION.md` still describes version 1 as `0x01` + `0x02` with unknown TLVs skipped, and still lists this as a #1434 follow-up.

Given the decode below, an implementer guessing 4 bytes doesn't get a degraded ceiling — they get their identity binding refused.

Could `0x03` go in the migration doc, and `0x04` be pinned in `PEER-ID-ROTATION.md §4.5`? The test vector already matches Android #879; it's the registry line that's missing.

jozanek (commented):
Thanks for this, and for flagging the `0x03` collision yourself.

**What I understood this to be for:** replacing an inferred per-peer fragment bound with one the peer states inside its Noise session. I think that's right — a value beats a bit for the next divergence, keeping it out of the cleartext announce avoids a build fingerprint, and `BLEFragmentCeilingPolicy` is the seam CLAUDE.md asks for.

**My main question is about the framing.** Android has advertised bit 8 since #728 (merged 46 minutes after #1434) while `FragmentManager.kt:191-195` caps every inbound stream at 256 regardless of type — both shipped since 2026-08-11. So iOS→Android media above ~117 KiB looks like it's being dropped today. Your Android #879 calls that "a live interop bug"; this body describes the same client conditionally. And with no `0x03` on the wire, `decide` still returns `.localCeiling`, so this PR doesn't change that outcome. Defaulting directed private media to 256 absent an advertisement would fix it unilaterally, at the cost of capping new-iOS→old-iOS for a release window. Either call seems defensible to me — I'd just like the trade named, rather than "no behaviour change for any deployed peer" standing as unambiguously the conservative one.

**Two blockers left inline:** `:4249` advertises 10,000 while our acceptance bound is bytes (~2,515 fragments at 469 B), and a malformed `0x03` discards the identity binding rather than the field.

**One check on the clamp:** the doc says it stops one side raising the other's memory exposure, but the memory belongs to the advertiser. Is the intended invariant us-as-originator plus the relays, which buffer any fragment they overhear (`BLEFragmentHandler.handle`)?

**What I couldn't verify:** anything on hardware, and whether Android's 256 is a true fragment count or a byte budget expressed in fragments. iOS's own limit is byte-based, so the two sides may already disagree on unit — if it's bytes, `0x03` negotiates the wrong quantity, and that's the one answer that would change my read of the shape rather than the details.

**What would move me to approve:** fix the advertised value, skip rather than reject, gate the above-256 lift to the `0x20` path, and write the byte into the docs. Happy to re-review.

bitchat/Services/BLE/BLEService.swift
```diff
@@ -2027,17 +2033,20 @@ final class BLEService: NSObject {
                 for: compatibilityRequest,
                 defaultChunkSize: defaultFragmentSize,
                 bleMaxMTU: bleMaxMTU
-            ), BLEOutboundFragmentPlanner.isPrivateMediaV1Compatible(plan) else {
+            ), ceiling.admits(fragmentCount: plan.totalFragments) else {
                 SecureLogger.warning(
-                    "Private media rejected: exceeds cross-platform 256-fragment limit",
+                    "Private media rejected: \(String(describing: ceiling.source)) limit of \(ceiling.maxFragments) fragments",
```
jozanek, line 2038: I think this is the only production reader of `ceiling.source`, and `SecureLogger` is compiled out of release builds — `SecureLogger.swift:129-165` and `:252` are all `#if DEBUG`. So there's no way to learn in the field that the negotiated branch fired.

Your parameterized rejection string is the one signal that does survive. Would folding the source into it work, so `.negotiated` 256 and `.migrationFallbackProxy` 256 aren't identical to a reader?

bitchat/Services/BLE/BLEService.swift
```diff
@@ -2009,13 +2009,19 @@ final class BLEService: NSObject {
...
+        // The proxy above is now only the fallback: a peer that advertised its
+        // own ceiling in authenticated `0x21` state has it honoured instead,
+        // which is what closes #1434. That also means the check can no longer
+        // be restricted to `fileTransfer` — an advertised ceiling constrains
+        // encrypted media too, and that is precisely the case (small
+        // reassembler behind the `0x20` path) the type proxy misses.
         if let transferId,
-           let recipientPeerID = PeerID(hexData: packetToSend.recipientID),
-           packetToSend.type == MessageType.fileTransfer.rawValue {
+           let recipientPeerID = PeerID(hexData: packetToSend.recipientID) {
```
jozanek, line 2019: Dropping the type guard puts `makePlan` on every encrypted media send — a full serialization plus every fragment materialized (`BLEOutboundFragmentPlanner.swift:44-67`) — before `startFragmentedPacket` does it again at `:5529`. No peer alive today can trip the branch, and `perf-floors.json` has no outbound planning benchmark.

Would `type == .fileTransfer || negotiated != nil` work? Behaviourally identical, and it keeps the encrypted path off this.

bitchat/Services/BLE/BLEService.swift
```diff
@@ -2027,17 +2033,20 @@ final class BLEService: NSObject {
...
                 )
                 TransferProgressManager.shared.rejectBeforeStart(
                     id: transferId,
                     reason: String(
-                        localized: "content.delivery.reason.private_media_too_many_fragments",
-                        defaultValue: "File is too large for this contact's client (more than 256 mesh fragments)",
-                        comment: "Failure reason when private media exceeds the Android-compatible fragment limit"
+                        format: String(
+                            localized: "content.delivery.reason.private_media_too_many_fragments",
+                            defaultValue: "File is too large for this contact's client (more than %lld mesh fragments)",
```
jozanek, line 2046: Now that the count is variable this has no plural variations, while the catalog uses `"plural"` 62 times elsewhere; the decoder accepts any `UInt16 > 0`, so "more than 1 mesh fragments" is reachable.

Low stakes at realistic ceilings — mostly flagging because `LocalizationCoverageTests.swift` checks key coverage but never format specifiers, so a later edit dropping `%lld` would go unnoticed. Worth a parity assertion?

PR #1576, opened 2026-07-31
feat: show recent DM threads at the top of the people sheet

Chessing234:
## Summary
- Jack re-scoped #615 to a persistent way to reopen past DMs (launch restore is #1364).
- People sheet gains a **recent messages** section above `#mesh`: up to 8 threads by latest activity, with unread glyph + last-message preview.
- Store helper + unit test for ordering / empty-thread skip / limit.

## Test plan
- [ ] After a few DMs, open people sheet → recent section lists them newest-first
- [ ] Tap a row opens that private chat
- [ ] Unread envelope shows when the thread has unread
- [ ] Empty / mesh-only installs hide the section
- [ ] Section title localizes (not a raw key)

Related to #615

Made with [Cursor](https://cursor.com)

PR #1672, opened 2026-08-15
feat: preview the newest message in the chats section

Chessing234:
The two follow-ups you named as worth keeping on #1576, ported onto main's `RecentChatList` rather than reopening that PR.

## What the rows show now

Each chats row gains a second line with the newest message in the thread.

Media messages are the reason this needs any logic at all. Their content *is* `"[image] 9F2A7C41-B0E3.jpg"` — the marker is part of the message, not decoration the view adds (`MimeType.Category.messagePrefix`, `BitchatMessage.mediaAttachment`). Rendering that verbatim would put an opaque on-disk filename in front of the reader. `RecentChatPreview` keeps the marker and drops the filename, which is both what you asked for and what the thread itself shows.

It matches against `MimeType.Category`'s own `messagePrefix` rather than three string literals, so adding a category cannot leave a path here still printing a filename. `.file` is included even though `mediaAttachment` does not resolve it to an attachment — it still arrives as prefixed content and would otherwise fall through to the text path.

Text is collapsed to one line and bounded at 80 characters, cutting on a word boundary when one is near the limit.

## Hiding the open thread

The other half: a row pointing at the conversation already on screen is the sheet offering to open what is open. Suppressed by the same identity keys the section already uses for roster overlap, so a thread opened under the stable Noise ID also suppresses the row keyed by its ephemeral alias.

**This needed a binding that was missing, and I would have shipped the bug without the test.** `selectedPeer` was not among the publishers `PeerListModel` observes, so the suppression only took effect whenever something *else* triggered a refresh. The integration test spent **30.07s** waiting for that to happen; with `privateChatManager.$selectedPeer` bound it takes **0.074s**. Closing the thread restores the row, so the suppression is a view of the current selection rather than a durable removal.

## Localization

**No catalog changes.** The markers are already the literal `[image]` / `[voice]` / `[file]` tokens the message content carries, so the preview reuses them rather than introducing keys that would need 30 translations.

The preview is threaded into the accessibility label too — the row uses `.accessibilityElement(children: .ignore)`, so a line left out there is one sighted users can read and VoiceOver cannot.

## Verification

- Full SwiftPM suite green, **2018 → 2029**. The delta is exactly the 11 tests added (10 unit, 1 integration).
- `RecentChatPreview` is framework-free, so every rule above is exercised directly rather than through a view.
- The integration test covers both halves against a real `PeerListModel`: newest-message-wins, multi-line flattening, the media marker, suppression on open, and restoration on close.
- No local SwiftLint, so I checked the `--strict` rules by hand against `.swiftlint.yml`.

## What I did not verify

No device or simulator run — the row layout is verified by reading, not by looking at it. Specifically, the second line changes each row from one line to two where a preview exists, and I have not seen how that lands against the surrounding sections at the narrowest sheet width. Worth a glance from someone with the app open before this merges.

### DIFF
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -49,8 +49,26 @@ struct RecentChatRow: Identifiable, Equatable {
     let displayName: String
     let hasUnread: Bool
     let lastActivity: Date
+    /// One-line snippet of the newest message, or nil when there is nothing
+    /// worth showing (see `RecentChatPreview`). Nil rather than "" so the
+    /// row can collapse to a single line instead of reserving blank space.
+    let preview: String?
 
     var id: String { peerID.id }
+
+    init(
+        peerID: PeerID,
+        displayName: String,
+        hasUnread: Bool,
+        lastActivity: Date,
+        preview: String? = nil
+    ) {
+        self.peerID = peerID
+        self.displayName = displayName
+        self.hasUnread = hasUnread
+        self.lastActivity = lastActivity
+        self.preview = preview
+    }
 }
 
 @MainActor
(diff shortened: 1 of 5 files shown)


Chessing234:
awaiting-review

PR #1675, opened 2026-08-16
Trigger Noise session rekey on absolute age, not just idle time

krishrathi1:
## The bug

`SecureNoiseSession.needsRenegotiation()` (`bitchat/Noise/SecureNoiseSession.swift`) has two triggers: message count nearing the 1-billion-message ceiling, and `lastActivityTime` idle for longer than `sessionTimeout` (24h). Neither ever fires for a session with steady traffic -- message count realistically never approaches a billion in normal chat use, and `lastActivityTime` resets on every `encrypt`/`decrypt` call, so a continuously active conversation never looks idle.

Meanwhile `encrypt`/`decrypt` both hard-fail with `sessionExpired` once `sessionStartTime` crosses that same 24h `sessionTimeout`, unconditionally -- with no rekey ever having been scheduled to prevent it, since nothing was ever checking absolute session age.

## Impact

Two peers who keep a conversation open past 24h (a stationary mesh node, or two phones left open overnight -- exactly the kind of multi-day-event use this app targets) hit this simultaneously and silently:

- Every message throws `sessionExpired` on both sides.
- `isEstablished()` still reports `true` (it only checks handshake state), so the UI keeps showing the session as "Secured"/"Verified".
- `MessageRouter` never falls back off the transport, since `canDeliverSecurely` never goes false.
- Nothing re-handshakes automatically: the one self-heal path (`BLENoisePacketHandler`'s decrypt-failure recovery) requires an *inbound ciphertext* to fail to decrypt for a session-establishment reason -- it explicitly excludes `.sessionExpired`/`.sessionExhausted` (`isDropOnlyCiphertextFailure`).

The conversation is silently and permanently dead until an app restart (discards the in-memory session manager) or a BLE disconnect/reconnect (forces revalidation).

## The fix

Adds a third `needsRenegotiation()` trigger directly on `sessionStartTime` age, firing at 90% of `sessionTimeout`. The rekey timer polls every 60s (`NoiseEncryptionService.rekeyCheckInterval`), so this leaves over two hours of margin to complete a full handshake before the hard cutoff -- the session keeps working normally in the meantime (only *flagged* for rekey, not degraded).

## Test

Adds `continuouslyActiveSessionStillNeedsRenegotiationBeforeExpiry`: a session with `sessionStartTime` at 95% of `sessionTimeout` in the past but `lastActivityTime` set to *now* (simulating steady traffic) must report `needsRenegotiation() == true`, and must still successfully `encrypt`/`decrypt` (it's flagged, not yet expired). Checked every other call site of `needsRenegotiation()`/`getSessionsNeedingRekey()` in the test suite (`NoiseCoverageTests`, `NoiseProtocolTests`, `NoiseEncryptionServiceTests`) to confirm none depend on the old always-false-for-active-sessions behavior at this age -- they either test message-count/idle-time thresholds directly (unaffected, sessionStartTime untouched) or force a rekey manually via `_test_initiateAutomaticRekey` (bypasses this function entirely).

## Verification

No Xcode/Swift toolchain available in this environment, so this is verified by reading and tracing, not compiling:

- Manually walked the new test's inputs through both the pre-fix and post-fix `needsRenegotiation()` logic line by line -- confirmed the pre-fix code returns `false` for a session with fresh `lastActivityTime` and old `sessionStartTime`, and the fix flips it to `true`.
- Confirmed via `git apply -R` (revert only the source fix, keep the new test) that the test's assertion is genuinely built on the pre-fix behavior returning `false`, then `git apply` to restore the fix.
- Confirmed the diff is minimal (2 files, +52/-2) and touches nothing outside `needsRenegotiation()` and the new test.

Happy to add more coverage or adjust the 90% threshold if there's a preferred margin.

### DIFF
--- a/bitchat/Noise/SecureNoiseSession.swift
+++ b/bitchat/Noise/SecureNoiseSession.swift
@@ -66,12 +66,24 @@ final class SecureNoiseSession: NoiseSession {
         if messageCount >= messageThreshold {
             return true
         }
-        
+
         // Check if last activity was more than 30 minutes ago
         if Date().timeIntervalSince(lastActivityTime) > NoiseSecurityConstants.sessionTimeout {
             return true
         }
-        
+
+        // A session that never goes idle (a long-lived conversation with
+        // steady traffic) keeps refreshing lastActivityTime forever, so the
+        // check above never fires -- only sessionStartTime tracks how close
+        // encrypt/decrypt are to the hard sessionExpired cutoff above. Firing
+        // at 90% of sessionTimeout leaves the rekey timer's polling interval
+        // (checked every 60s, see NoiseEncryptionService.rekeyCheckInterval)
+        // room to complete a full handshake well before the old session
+        // starts throwing sessionExpired on every call.
+        if Date().timeIntervalSince(sessionStartTime) > NoiseSecurityConstants.sessionTimeout * 0.9 {
+            return true
+        }
+
         return false
     }
     
--- a/bitchat/Noise/SecureNoiseSession.swift
+++ b/bitchat/Noise/SecureNoiseSession.swift
@@ -67,7 +67,7 @@ final class SecureNoiseSession: NoiseSession {
             return true
         }
 
-        // Check if last activity was more than 30 minutes ago
+        // Check if last activity exceeded sessionTimeout
         if Date().timeIntervalSince(lastActivityTime) > NoiseSecurityConstants.sessionTimeout {
             return true
         }
--- a/bitchatTests/Noise/NoiseCoverageTests.swift
+++ b/bitchatTests/Noise/NoiseCoverageTests.swift
@@ -893,6 +893,44 @@ struct NoiseCoverageTests {
         }
     }
 
+    @Test("A continuously active session still schedules a rekey before it hard-expires")
+    func continuouslyActiveSessionStillNeedsRenegotiationBeforeExpiry() throws {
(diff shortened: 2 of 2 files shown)


Chessing234 (commented):
checked the claim and it holds — needsRenegotiation() on main only has the message-count and lastActivityTime triggers, and lastActivityTime is reset by every encrypt/decrypt, so a busy session never looks idle. nice catch.

one thing i'd want answered before this is relied on: the rekey timer is Timer.scheduledTimer on the main run loop (NoiseEncryptionService.swift:1059). on iOS that stops firing once the app is suspended, and the 2.4h window this opens up is mostly going to be spent backgrounded — which is exactly the overnight case in your impact section. if the timer is asleep for the whole window the session still hits the hard sessionExpired cutoff and you're back to the silent-death path.

i can't test that here, so it may well be that the BLE background modes keep the run loop alive enough. but if not, the age check probably wants to run wherever the session is next touched (encrypt/decrypt) rather than only on the timer.

also, unrelated but adjacent: the comment above the idle check still says "more than 30 minutes ago" while it compares against sessionTimeout (24h).

krishrathi1:
Fixed the stale comment in 
eedsRenegotiation() above the idle check to accurately reference sessionTimeout.

Regarding background execution on iOS: 
eedsRenegotiation() is evaluated continuously via NoiseSessionManager.checkRekeyNeeded(), which runs on both periodic timer checks and during active packet processing when the app wakes for BLE background events.

PR #1679, opened 2026-08-17
test(l10n): per-catalog attribution and file:line output for the coverage guard

Chessing234:
the two follow-ups you named on #1613, on current main.

**per-catalog attribution.** the code-side guard accepted a key found in *either* catalog. the share extension bundles only its own, so a key it references that lives solely in the app catalog ships english there exactly like one that exists nowhere. now `bitchat/` is checked against `bitchat/Localizable.xcstrings` and `bitchatShareExtension/` against the extension catalog, and the failure says which of the two cases it is.

**file:line.** failures named only the file, so finding the key in a 900-line view meant grepping. each reference now carries its repo-relative path and line, one per line.

both are clean on current main (520 app keys, 5 extension keys, no misattributed references).

what i ran locally: no xcode on this machine, so i lifted the test into a stub harness (`#expect`/`#require` replaced by plain functions) and ran the four checks against the real catalogs — all pass. then i injected two probe keys into `ShareViewController.swift` to see the failure text:

```
FAIL: keys referenced under bitchatShareExtension/ but absent from bitchatShareExtension/Localization/Localizable.xcstrings — these ship English to all non-source locales:
bitchatShareExtension/ShareViewController.swift:193 probe.missing.key (present in no catalog)
bitchatShareExtension/ShareViewController.swift:193 common.cancel (present in bitchat/Localizable.xcstrings, which this target does not bundle)
```

and reverted the probes. i could not run the real `Testing` suite here — CI covers that.

### DIFF
--- a/bitchatTests/LocalizationCoverageTests.swift
+++ b/bitchatTests/LocalizationCoverageTests.swift
@@ -79,26 +79,39 @@ struct LocalizationCoverageTests {
     /// coverage tests stayed green. Interpolated keys can't be checked
     /// statically and are skipped; every literal key must resolve.
     ///
-    @Test func everyCodeReferencedKeyExistsInACatalog() throws {
+    /// The catalogs are checked per target, not as a union: the share
+    /// extension bundles only its own catalog, so a key it references that
+    /// lives solely in the app catalog is just as untranslated there as one
+    /// that exists nowhere.
+    @Test func everyCodeReferencedKeyExistsInItsOwnCatalog() throws {
         let main = try Self.loadCatalog(Self.mainCatalogPath)
         let shareExt = try Self.loadCatalog(Self.shareExtensionCatalogPath)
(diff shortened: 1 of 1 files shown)


Chessing234:
good catch on the shared sources — the share extension does compile bitchat/Services/SharedContentHandoff.swift and bitchat/Services/TransportConfig.swift (project.pbxproj membership exceptions), so a key referenced from either has to be in both catalogs.

pushed b2641f8: the check now reads that exception list out of project.pbxproj (so it follows the set instead of hardcoding two paths) and additionally requires every key referenced from those files to exist in the extension catalog. probed it by adding String(localized: "Choose an image") — an app-only key — to TransportConfig.swift; the first check stays silent and the new one fails with the file:line. tree as it stands is clean on both.

Chessing234:
awaiting-review

PR #1680, opened 2026-08-17
fix: cap the dedup set when IDs arrive through markProcessed

Chessing234:
`MessageDeduplicator` documents a bounded working set — at most `maxCount` live IDs, oldest evicted first — and `isDuplicate` enforces it by calling `trimIfNeeded()`. `markProcessed` doesn't, so IDs added that way just accumulate until something ages them out at `maxAge` (300 s).

that is the send path. every self-broadcast pre-marks its own dedup ID (`BLEService.swift:885`), so does announce-back tracking (`:1532`, `:2851`, `:2918`, `:6532`, `:6616`), and a node that mostly sends can go a long time without an inbound packet to hit the eviction in `isDuplicate`.

measured on the real class, `maxCount` 1000:

```
before:  10000 markProcessed -> live entries: 10000   (isDuplicate: 964)
after:   10000 markProcessed -> live entries:   964   (isDuplicate: 964)
```

second commit adds four tests: both entry points stay under the cap, eviction keeps the most recent 75%, and an evicted ID is genuinely forgotten rather than stranded in the lookup table.

no xcode on this machine, so i compiled `MessageDeduplicator.swift` standalone with swiftc and ran the new tests against it through a small harness (`#expect` replaced by a counting stub): 4/4 pass with the fix, and 2 expectations fail without it. the real `Testing` suite i could not run here — CI covers that.

### DIFF
--- a/bitchat/Utils/MessageDeduplicator.swift
+++ b/bitchat/Utils/MessageDeduplicator.swift
@@ -59,11 +59,15 @@ final class MessageDeduplicator {
 
         if lookup[id] == nil {
             let now = Date()
+            // Expire first, then cap, in that order and for the same reason as
+            // isDuplicate: trimming by count alone would drop still-valid IDs
+            // to make room while an already-expired prefix sat in front of them.
+            cleanupOldEntries(before: now.addingTimeInterval(-maxAge))
             entries.append(Entry(id: id, timestamp: now))
             lookup[id] = now
-            // Same cap as isDuplicate: a node that only sends (self-broadcasts,
-            // announce-backs) never reaches the eviction in isDuplicate, so
-            // without this its dedup set grows for the whole maxAge window.
+            // A node that only sends (self-broadcasts, announce-backs) never
+            // reaches the eviction in isDuplicate, so without this its dedup
+            // set grows for the whole maxAge window.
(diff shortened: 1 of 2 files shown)


Chessing234:
good catch on the ordering — pushed a commit that expires before capping in `markProcessed`, matching `isDuplicate`.

on the specific claim that the cap can also drop still-valid IDs: i could not get that to reproduce. eviction is oldest-first and expired entries are by definition the oldest, and `trimIfNeeded` runs on every insert, so the expired prefix drains before the cap ever reaches live entries. with maxAge 1s, 900 expired IDs and then an 800-ID burst:

```
cap only:    fresh IDs still known: 800/800, expired IDs still held: 147/900
expire+cap:  fresh IDs still known: 800/800, expired IDs still held:   0/900
```

so the change is worth having for the reason your comment gives — the table should not be holding entries it considers expired — just not for lost live IDs.

jackjackbits:
Closing in favor of #1635 to keep one implementation of the markProcessed capacity fix. #1635 also expires entries before the duplicate lookup and uses an injected clock for deterministic expiry coverage. Useful load/eviction cases from this PR can be retained there.

PR #1681, opened 2026-08-18
fix: avoid swipe conflict when seeking voice notes

### LINKED ISSUE #1658: Dragging on a playing voice note in a DM can leave the conversation

qtjg:
Fixes #1658.

## Problem

The private conversation attaches a high-priority right-swipe-to-leave `DragGesture` to the message list, while `WaveformView` uses `DragGesture(minimumDistance: 0)` for seeking. A horizontal drag over a playing voice note could therefore be claimed by the ancestor gesture, causing the seek to fail or accidentally closing the conversation.

## Fix

This change keeps the waveform's original drag-seeking behavior, including press-and-slide-to-target before release. While any voice note is actively playing, the private-chat swipe-to-leave gesture is excluded from subviews so the waveform receives the drag. The swipe gesture resumes as soon as playback stops.

This deliberately prioritizes playback seeking over swipe-to-leave while audio is audible. Leaving remains available through the sidebar, and a playing voice note can no longer cause a scrub to close the conversation.

The bounded `seekFraction` helper remains extracted and handles normal mapping, out-of-range coordinates, and zero-width geometry safely.

## Test plan

- Added regression coverage for swipe thresholds, playback suppression and resumption, coordinator activity state, paused playback release, stale deactivation, and playback takeover.
- Retained coverage for seek-fraction mapping, bounds clamping, and zero-width geometry.
- `git diff --check` and committed-diff validation pass.
- Swift/Xcode test execution is pending in CI because the local sandbox does not include the Swift toolchain.

### DIFF
--- a/bitchat/Views/Media/WaveformView.swift
+++ b/bitchat/Views/Media/WaveformView.swift
@@ -8,6 +8,16 @@ struct WaveformView: View {
     let isInteractive: Bool
     @ThemedPalette private var palette
 
+    /// Converts a tap location into a bounded playback fraction.
+    ///
+    /// Keep this separate from the gesture so the coordinate contract can be
+    /// tested without mounting SwiftUI. A zero-sized geometry can occur while
+    /// a row is being laid out and must not result in a bogus seek.
+    static func seekFraction(forX x: CGFloat, inWidth width: CGFloat) -> Double? {
+        guard width > 0 else { return nil }
+        return max(0, min(1, Double(x / width)))
+    }
+
     private var clampedPlayback: Double {
         max(0, min(1, playbackProgress))
     }
@@ -52,11 +62,20 @@ struct WaveformView: View {
                 if isInteractive, let onSeek = onSeek {
                     Color.clear
                         .contentShape(Rectangle())
+                        // The private conversation owns a high-priority
+                        // swipe-to-leave drag on the message list. A drag here
+                        // is therefore interpreted as navigation instead of
+                        // seeking and can close the conversation. Seeking is
+                        // intentionally a tap: it preserves the swipe gesture
+                        // while keeping the waveform's direct-manipulation
+                        // affordance deterministic.
                         .gesture(
-                            DragGesture(minimumDistance: 0)
+                            SpatialTapGesture(count: 1, coordinateSpace: .local)
                                 .onEnded { value in
-                                    guard geometry.size.width > 0 else { return }
-                                    let fraction = max(0, min(1, value.location.x / geometry.size.width))
+                                    guard let fraction = Self.seekFraction(
+                                        forX: value.location.x,
+                                        inWidth: geometry.size.width
+                                    ) else { return }
                                     onSeek(fraction)
                                 }
                         )


Chessing234 (commented):
two things.

swapping DragGesture for SpatialTapGesture doesn't just make seeking deterministic, it removes press-and-slide-to-target. the old `DragGesture(minimumDistance: 0).onEnded` already only seeked on release, so you could press down and slide to the spot you wanted before lifting; with a tap you have to land on it directly. on a waveform that's a few points tall that's a real loss worth calling out in the description rather than leaving as a side effect.

second, this and #1668 both fix #1658 from opposite ends — that one stands down the swipe-to-leave gesture while a voice note is playing and keeps drag-seeking. they're mutually exclusive, so someone has to pick. disclosure: #1668 is mine, and i'd rather jack chose on merit than have two branches sit there fighting.

no objection to the seekFraction extraction, that part is a clear improvement and testable.

qtjg:
Thanks for the detailed review. I agree that replacing the drag with `SpatialTapGesture` would unnecessarily remove press-and-slide seeking, so I restored the original `DragGesture(minimumDistance: 0)` behavior.

I chose the alternative approach from #1668: while a voice note is actively playing, the private-chat swipe-to-leave gesture is excluded from subviews so the waveform receives the drag. The swipe resumes when playback stops, and leaving remains available from the sidebar. This preserves the existing seeking interaction while addressing the reported accidental-navigation bug.

The bounded `seekFraction` extraction and its edge-case coverage are retained, along with regression tests for swipe thresholds, playback suppression/resumption, coordinator activity, pause release, stale deactivation, and playback takeover. The local sandbox lacks Swift/Xcode, so the Swift test suite remains pending in CI; static diff validation passes and the branch has no merge conflicts.

jackjackbits:
Closing in favor of #1668. The playback coordinator, swipe policy, sheet changes, and policy tests are identical at the reviewed heads. The small WaveformView seekFraction extraction is the only additional production change and can be carried separately if useful.

PR #1682, opened 2026-08-18
Log oversized compressed payloads before rejection

### LINKED ISSUE #1628: Log silent rejection of oversized compressed payloads instead of dropping with no trace

qtjg:
Fixes #1628.

## Summary

When a compressed packet declares an expanded size above `FileTransferLimits.maxFramedFileBytes`, the decoder currently rejects it without logging the reason. This makes oversized compressed payloads difficult to diagnose, particularly when platform-specific limits differ.

This change logs the declared expanded size and configured ceiling at the existing security warning level before rejecting the packet. Acceptance behavior is unchanged.

## Tests

Added a v2 regression test that constructs a valid compressed frame whose declared expanded size exceeds the configured limit and verifies that `BinaryProtocol.decode` rejects it before attempting decompression.

The local source and test files pass deterministic source checks and `git diff --check`. The full Swift/Xcode suite could not be run in this Linux environment because `swift` and `xcodebuild` are unavailable.

### DIFF
--- a/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
+++ b/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
@@ -235,7 +235,7 @@ public struct BinaryProtocol {
                 }
             } else {
                 let value = UInt16(originalSize)
-                data.append(UInt8((value >> 8) & 0xFF))
+                data.append(UInt8((value >> 8) & 0xFF)
                 data.append(UInt8(value & 0xFF))
             }
         }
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
this is the same BinaryProtocol.swift guard as #1623. please close this as a duplicate — the new test would already pass on main (decode already returns nil) and never asserts the log.

Chessing234 (commented):
rejection is right, but logging every oversized expanded-size on the decode hot path is attacker-controlled. a peer can spray compressed frames with huge claimed originalSize and flood SecureLogger. keep the reject; drop the per-frame warning or rate-limit/dedup it. the new test also never asserts a log fired, so it only rechecks the existing reject.

jackjackbits:
Closing in favor of #1623, which covers the same oversized compressed-payload warning. The added test here checks rejection that already occurs on main; any additional diagnostic coverage can be consolidated into #1623.

PR #1683, opened 2026-08-20
Enable autocorrection in message composer (#969)

### LINKED ISSUE #969: Enable auto correction when typing a message

sp201004:
## Summary
Fixes #969.

The message input field had autocorrection turned off. This flips it on so autocorrect works while typing a message, matching the behavior users expect from a standard text field.

## Change
- `bitchat/Views/ContentComposerView.swift`: `.autocorrectionDisabled(true)` → `.autocorrectionDisabled(false)` on the message `TextField`.

Only the message composer is affected. Fields where autocorrect should stay off (nickname, URL entry, location/geohash search) are intentionally left unchanged.

## Testing
Single-token change on an existing, valid SwiftUI modifier (`autocorrectionDisabled(Bool)`). I don't have full Xcode in my environment (Command Line Tools only), so I couldn't run a full app build/simulator check — would appreciate a maintainer confirming on a device/simulator.

### DIFF
--- a/bitchat/Views/ContentComposerView.swift
+++ b/bitchat/Views/ContentComposerView.swift
@@ -76,7 +76,7 @@ struct ContentComposerView: View {
                 .bitchatFont(size: 15)
                 .foregroundColor(palette.primary)
                 .focused(isTextFieldFocused)
-                .autocorrectionDisabled(true)
+                .autocorrectionDisabled(false)
                 #if os(iOS)
                 .textInputAutocapitalization(.sentences)
                 #endif


sp201004:
### Build verification

I compiled this locally with `swift build` against the macOS target (Swift 6.3, macOS SDK). The build reached **359/362 modules**, and my modified file `ContentComposerView.swift` compiled with **no errors**.

The build only stopped on `#Preview { ... }` blocks in three files I did **not** touch (`ImagePreviewView.swift`, `MacImagePickerView.swift`, `MeshTopologyView.swift`), all failing with:

```
error: external macro implementation type 'PreviewsMacros.SwiftUIView' could not be found
for macro 'Preview(_:body:)'; plugin for module 'PreviewsMacros' not found
```

That's an environment limitation — the `PreviewsMacros` plugin ships with full Xcode and isn't available in a Command Line Tools-only `swift build`. It's unrelated to this change and reproduces on a clean checkout.

Net: the change is a single boolean flip on an existing, cross-platform SwiftUI modifier (`autocorrectionDisabled(Bool)`) and is compile-clean. A quick device/simulator run under full Xcode to confirm the autocorrect behavior would be appreciated.

Chessing234 (commented):
jack already answered this one on #969 — the composer doubles as the command line, so autocorrect fights /commands, @mentions and #geohash, and the autocomplete depends on exact text. what he asked for there is enabling it *except* while the token under the cursor starts with / @ or #, which a blanket flip doesn't get you.

worth reshaping before he looks, otherwise this is a round wasted.

disclosure: #1508 is mine and does the token-aware version, so there's a dup to sort out either way — happy for whichever shape he prefers to win.

sp201004:
Thanks for the review, @Chessing234 — you're right, a blanket flip doesn't handle the `/` `@` `#` token cases, and #1508 already does the token-aware version the maintainer asked for. Closing this in favor of #1508 to avoid a duplicate. 🙏

PR #1684, opened 2026-08-20
Consolidate base64url encoding into Base64URLCoding (#642)

sp201004:
## Summary
Works toward #642 by removing duplicate base64url **encode** implementations and routing them through the existing shared `Base64URLCoding` utility (`bitchat/Nostr/Base64URLCoding.swift`).

Since the issue was filed, a shared `Base64URLCoding` enum already landed and `NostrProtocol` + `NostrInboundPipeline` were migrated to it. `ChatViewModel` no longer contains any base64url helper. This PR cleans up the remaining source-level encode duplicates.

## Changes
- `bitchat/Nostr/NostrEmbeddedBitChat.swift`: dropped the private `base64URLEncode(_:)` helper; its 4 call sites now use `Base64URLCoding.encode(_:)`.
- `bitchat/Services/VerificationService.swift`: replaced the inline `base64EncodedString() + replacingOccurrences(...)` chain with `Base64URLCoding.encode(_:)`.

Both removed implementations were byte-for-byte identical to `Base64URLCoding.encode` (`+`→`-`, `/`→`_`, strip `=`), so behavior is unchanged.

## Intentionally left alone
- `CashuTokenDecoder.base64URLDecode(_:)` is **not** migrated. It parses attacker-controlled token payloads and has deliberately stricter padding handling (strips existing `=`, rejects a length-remainder of 1) that differs from the shared decoder. Changing a security-sensitive parser's behavior doesn't belong in a consolidation chore — happy to align it in a separate, focused change if maintainers want that.

## Testing
The shared utility is already covered by `bitchatTests/Nostr/Base64URLCodingTests.swift`.

I compiled locally with `swift build`; both modified files (`NostrEmbeddedBitChat.swift`, `VerificationService.swift`) compile cleanly. The build only fails on unrelated `#Preview` macros (the `PreviewsMacros` plugin ships with full Xcode and isn't available in a Command Line Tools-only `swift build`) — that reproduces on a clean checkout and is unrelated to this change. A CI/Xcode run would give a full green build.

### DIFF
--- a/bitchat/Nostr/NostrEmbeddedBitChat.swift
+++ b/bitchat/Nostr/NostrEmbeddedBitChat.swift
@@ -28,7 +28,7 @@ struct NostrEmbeddedBitChat {
         )
 
         guard let data = packet.toBinaryData() else { return nil }
-        return "bitchat1:" + base64URLEncode(data)
+        return "bitchat1:" + Base64URLCoding.encode(data)
     }
 
     /// Build a `bitchat1:` base64url-encoded BitChat packet carrying a delivery/read ack for Nostr DMs.
@@ -51,7 +51,7 @@ struct NostrEmbeddedBitChat {
         )
 
         guard let data = packet.toBinaryData() else { return nil }
-        return "bitchat1:" + base64URLEncode(data)
+        return "bitchat1:" + Base64URLCoding.encode(data)
     }
 
     /// Build a `bitchat1:` ACK (delivered/read) without an embedded recipient peer ID (geohash DMs).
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
the change itself is clean — all four call sites move over and the private helper is deleted with them, so nothing is left orphaned for periphery to trip on.

the thing to sort out is #1500, which has been open since june doing the same consolidation but wider: it also converts NostrProtocol.swift and CashuTokenDecoder.swift, and it moves the utility into BitFoundation. if that one lands after this, the type you're calling here moves modules and these two files get touched a third time.

no view on which should win — but they shouldn't both proceed independently, and #1500 is the older one.

sp201004:
Thanks @Chessing234 — agreed. #1500 has been open since June doing this same consolidation but wider (it also converts NostrProtocol.swift and moves the helper into BitFoundation), and it's the older PR. Closing this in favor of #1500 so they don't proceed independently.

PR #1687, opened 2026-08-23
fix: reject a signed byte pair in Data(hexString:)

Chessing234:
`Data(hexString:)` chunks the string into two-character pairs and hands each to `UInt8(_:radix:)`. That initializer accepts a leading sign, so `"+f"` parses as `0x0f`. Its own doc comment promises nil for a string that "contains invalid hex characters", and `+` is one.

Every identity-bearing hex string in the app goes through this parser — Noise keys, Nostr pubkeys, group member fingerprints, signatures, routing IDs — so each one had a second spelling that decodes to the same bytes.

`PeerID` is where that shows up most sharply. A 64-character id built from `"+b"` pairs:

```
isValid       = true
isHex         = false      <- disagrees with the line below
isNoiseKeyHex = true
noiseKey      = 0b0b0b…0b  <- byte-identical to the genuine peer
toShort()     = f0e38b830ebd8a50  <- same short routing ID
```

`isShort` validates the 16-hex form with `isHexDigit` and rejects `+` correctly; `isNoiseKeyHex` validates the 64-hex form by asking whether `Data(hexString:)` returns non-nil, so the two halves of the same check disagreed. `PeerID` equality and hashing are on the string, so the spoofed id is a distinct key in any dictionary while resolving to the same cryptographic identity.

Fix is to require two hex digits before parsing the pair. Everything the parser is documented to accept still parses: plain, uppercase, `0x`/`0X`-prefixed, whitespace-padded, and every byte value round-trips.

I did not widen this beyond the parser. `PeerID.isValid` needs no change once `Data(hexString:)` keeps its contract, and I would rather not touch the validity rules in the same diff.

### Evidence

No Xcode on this machine, so I could not run `bitchatTests` or build the app. What I did instead: built `BitFoundation` with `swift build` and linked its object files into a standalone harness, so the assertions below ran against the real compiled module rather than a stub.

Before the fix, the harness reproduced the table above — `+b`×32 and `0b`×32 are distinct `PeerID` values that both pass `isValid` and produce the identical 32-byte key and identical `toShort()`. After it, the spoofed id returns nil from `noiseKey`, `isValid` is false, `isHex` and `isNoiseKeyHex` agree, and the genuine id is untouched.

I then ran all 22 assertions from the two test files through that harness; they pass. The tests themselves are swift-testing, and the `Testing` module needs the Xcode toolchain, so **I have not executed `DataHexTests` or `PeerIDTests` as tests** — CI is the first thing that will. The assertions inside them are the ones I ran; the `@Test`/`#expect` scaffolding is copied from the sibling files.

`swift build --package-path localPackages/BitFoundation` is clean.

Related but not a duplicate: #911 hardened the same initializer for odd length, `0x` prefix, whitespace and empty input. Those all landed and are still there; the sign case was not part of it.

This was AI-assisted.

### DIFF
--- a/localPackages/BitFoundation/Tests/BitFoundationTests/DataHexTests.swift
+++ b/localPackages/BitFoundation/Tests/BitFoundationTests/DataHexTests.swift
@@ -1,0 +1,62 @@
+//
+// DataHexTests.swift
+// bitchatTests
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Testing
+import Foundation
+import BitFoundation
+
+struct DataHexTests {
+    // MARK: - Accepted forms
+
+    @Test func decodes_plain_hex() {
+        #expect(Data(hexString: "0a1b2c")?.hexEncodedString() == "0a1b2c")
+    }
+
+    @Test func decodes_uppercase_hex() {
+        #expect(Data(hexString: "0A1B2C")?.hexEncodedString() == "0a1b2c")
+    }
+
+    @Test func decodes_prefixed_and_padded_hex() {
+        #expect(Data(hexString: "0x0a1b")?.hexEncodedString() == "0a1b")
+        #expect(Data(hexString: "0X0a1b")?.hexEncodedString() == "0a1b")
+        #expect(Data(hexString: "  0a1b  ")?.hexEncodedString() == "0a1b")
+    }
+
+    @Test func round_trips_encoded_bytes() {
+        let bytes = Data((0...255).map { UInt8($0) })
+        #expect(Data(hexString: bytes.hexEncodedString()) == bytes)
+    }
+
+    // MARK: - Rejected forms
+
+    @Test func rejects_odd_length() {
+        #expect(Data(hexString: "0a1") == nil)
+    }
+
+    @Test func rejects_non_hex_characters() {
+        #expect(Data(hexString: "zz") == nil)
+        #expect(Data(hexString: "0a zz") == nil)
+    }
+
+    /// UInt8(_:radix:) accepts a leading sign, so "+f" parses as 0x0f unless the
+    /// digits are checked first. Two spellings of one identity-bearing hex
+    /// string decoding to the same bytes is the whole problem.
+    @Test func rejects_signed_byte_pairs() {
+        #expect(Data(hexString: "+a") == nil)
+        #expect(Data(hexString: "-a") == nil)
+        #expect(Data(hexString: "0a+b") == nil)
+        #expect(Data(hexString: "+a+b") == nil)
+    }
+
+    @Test func a_signed_pair_does_not_alias_a_genuine_key() {
+        let genuine = String(repeating: "0b", count: 32)
+        let signed = String(repeating: "+b", count: 32)
+        #expect(Data(hexString: genuine) != nil)
+        #expect(Data(hexString: signed) == nil)
+    }
+}
--- a/localPackages/BitFoundation/Tests/BitFoundationTests/PeerIDTests.swift
+++ b/localPackages/BitFoundation/Tests/BitFoundationTests/PeerIDTests.swift
@@ -362,6 +362,10 @@ struct PeerIDTests {
     @Test func rejects_invalid_characters() {
         #expect(!PeerID(str: "peer!@#").isValid)
         #expect(!PeerID(str: "gggggggggggggggg").isValid) // not hex for short form
+        // A 64-character id is only valid because it decodes to a Noise key, so
+        // it has to be rejected on the same grounds a short one is.
+        #expect(!PeerID(str: String(repeating: "+b", count: 32)).isValid)
+        #expect(!PeerID(str: String(repeating: "gg", count: 32)).isValid)
     }
     
     @Test func rejects_too_long() {
@@ -388,6 +392,13 @@ struct PeerIDTests {
         let badPeerID = PeerID(str: bad)
         #expect(!badPeerID.isNoiseKeyHex)
         #expect(badPeerID.noiseKey == nil)
+
+        // isHex and isNoiseKeyHex have to agree: a signed pair is not hex, so it
+        // must not decode to a Noise key either.
+        let signedPeerID = PeerID(str: String(repeating: "+b", count: 32))
+        #expect(!signedPeerID.isHex)
+        #expect(!signedPeerID.isNoiseKeyHex)
+        #expect(signedPeerID.noiseKey == nil)
     }
     
     @Test func prefixes() {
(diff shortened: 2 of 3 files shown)


localPackages/BitFoundation/Tests/BitFoundationTests/DataHexTests.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    // MARK: - Rejected forms
+
+    @Test func rejects_odd_length() {
+        #expect(Data(hexString: "0a1") == nil)
+    }
+
+    @Test func rejects_non_hex_characters() {
+        #expect(Data(hexString: "zz") == nil)
+        #expect(Data(hexString: "0a zz") == nil)
```
jozanek, line 43: "0a zz" is five characters, so it returns nil at the odd-length guard above and never reaches the pair check this file is about - it would keep passing if the fix were reverted. An even-length string like "0azz" would exercise the new code.

Worth knowing: the same thing happens at bitchatTests/Utils/HexStringTests.swift:88-91, where "01-02" is also five characters. That test is probably part of why this bug survived as long as it did.

localPackages/BitFoundation/Tests/BitFoundationTests/DataHexTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /// UInt8(_:radix:) accepts a leading sign, so "+f" parses as 0x0f unless the
+    /// digits are checked first. Two spellings of one identity-bearing hex
+    /// string decoding to the same bytes is the whole problem.
+    @Test func rejects_signed_byte_pairs() {
+        #expect(Data(hexString: "+a") == nil)
+        #expect(Data(hexString: "-a") == nil)
```
jozanek, line 51: I believe "-a" was already nil before this change: UInt8 is unsigned, so the negative branch subtracts from zero and overflows for any nonzero digit. If that's right, this assertion passes but isn't a regression test.

The one negative-sign string that did parse looks like "-0" - 0 &* 16 = 0, 0 - 0 = 0, no overflow. Would you be open to adding `#expect(Data(hexString: "-0") == nil)`? I reasoned this from the stdlib rather than running it, so I may have it wrong.

localPackages/BitFoundation/Tests/BitFoundationTests/PeerIDTests.swift
```diff
@@ -388,6 +392,13 @@ struct PeerIDTests {
         let badPeerID = PeerID(str: bad)
         #expect(!badPeerID.isNoiseKeyHex)
         #expect(badPeerID.noiseKey == nil)
+
+        // isHex and isNoiseKeyHex have to agree: a signed pair is not hex, so it
+        // must not decode to a Noise key either.
+        let signedPeerID = PeerID(str: String(repeating: "+b", count: 32))
+        #expect(!signedPeerID.isHex)
```
jozanek, line 399: This asserts the two agree, and for signed pairs they now do. `Character.isHexDigit` is Unicode-aware, though - `hexDigitValue` also matches the fullwidth ranges 0xFF10...0xFF19, 0xFF21...0xFF26, 0xFF41...0xFF46 - so a 64-character fullwidth `bare` still gives `isHex == true` with `isNoiseKeyHex == false`. The divergence this PR is named after is narrower than closed.

It fails closed today: every `isShort`/`isValid` consumer I traced compares against key-derived ASCII bytes afterward. Would you be open to either narrowing the claim here, or a follow-up making `PeerID.isHex` use `bare.allSatisfy { $0.isASCII && $0.isHexDigit }`?

jozanek (commented):
Thanks for the writeup - the `isHex == false` / `isNoiseKeyHex == true` contradiction is what made this reviewable, and keeping the diff at the parser reads as the right call. I checked the composition and the fix is correct: `Character.isHexDigit` is ASCII hex plus the fullwidth forms, `UInt8(_:radix:)` takes ASCII plus an optional sign, so the intersection is exactly ASCII hex and both halves of the guard are load-bearing.

Three things I'd like your read on.

**There's an existing implementation of this fix on a shelved branch.** Commit 3f6dc7d (part of #1350, closed unmerged 2026-07-10) replaces the decode loop with an ASCII nibble lookup over `Array(hex.utf8)`; its message names this exact bug - "sign-prefixed pairs like `+f`, which the old radix parser incorrectly accepted, are now rejected." It rejects the sign by construction, counts bytes rather than grapheme clusters, and adds a test file at the same path as yours, so the two conflict on both files. I think it helps you: the repo owner independently corroborated the finding, and the hex codec is the one piece of #1350 never re-landed. Would you be open to rebasing onto it, or to saying why the narrower guard is preferable?

**I could not find an attacker-reachable path.** The five places a hex *string* crosses a trust boundary (`NostrProtocol.isValidSignature`, `NostrPoW`, `VerificationService`, `RequestSyncPacket`, `GroupProtocol`) all fail closed on signature or event-id checks, or converge rather than split. What I did find, and think is the stronger argument, is latent rather than reachable: `NoiseSessionManager.authenticatedRemoteKey` (NoiseSessionManager.swift:1131-1142) falls through `isShort` to `claimedPeerID.noiseKey`, so a 64-character `+b` id authenticates against the genuine static key while `PeerID` equality stays on the string. #1645 hardened that same function last month for a branch it called unreachable, reasoning that "a latent accept-any-key path is exactly the kind of thing that bites later" - that precedent fits this change better than the description's broader claim does. Would you narrow the framing to the contract violation plus that path?

**Small correction:** #911 was self-closed without review and landed nothing; #919 is what merged (548a20e), later moved into BitFoundation by #1089.

Smaller notes inline. One I couldn't verify: I reasoned about `UInt8("-a", radix: 16)` from the stdlib rather than running it, so please sanity-check that against CI.

What would move me to approve: a decision on 3f6dc7d, and a description that claims what the evidence supports. The code itself I'm satisfied with - all seven checks are green, including the BitFoundation job, so the "couldn't run the tests" caveat is answered.

localPackages/BitFoundation/Tests/BitFoundationTests/DataHexTests.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import Testing
+import Foundation
+import BitFoundation
+
+struct DataHexTests {
```
jozanek, line 13: bitchatTests/Utils/HexStringTests.swift already covers this initializer with 17 tests - plain, uppercase, mixed case, 0x/0X, whitespace, empty in three variants, odd length, invalid characters, round-trip. As far as I can tell about two assertions here are new.

Two suites over one function tend to drift, and the next person fixing hex parsing has to find both. Would you consider adding the new assertions to the existing file, or moving it into BitFoundationTests next to the code and dropping the app-target copy?

Chessing234:
awaiting-review

PR #1688, opened 2026-08-24
fix: stop unattributable prekey bundles from wedging the pending stash

Chessing234:
### The bug

`handlePrekeyBundle` (`BLEService.swift`) has to cope with a bundle arriving before the announce that binds its owner's signing key, so when no key is bound yet it stashes the packet in `pendingPrekeyBundles` and `drainPendingPrekeyBundles` retries once a verified announce lands.

Nothing about a stashed packet has been verified at that point — deferring the signature check is the whole reason the stash exists. The only gate in front of it is:

```swift
guard PeerID(hexData: packet.senderID) == owner else { return }   // owner = PeerID(publicKey: bundle.noiseStaticPublicKey)
```

Both halves of that are public and attacker-choosable: pick any 32 bytes as `noiseStaticPublicKey`, set `senderID` to its derived short ID, attach a garbage signature. No handshake, no announce, no session needed.

The stash then did this:

```swift
if pendingPrekeyBundles[owner] != nil
    || pendingPrekeyBundles.count < Self.pendingPrekeyBundleCap {   // 64
    pendingPrekeyBundles[owner] = packet
}
```

New owners are refused once the dictionary holds 64 entries, and entries only ever leave via `drainPendingPrekeyBundles(for:)` — which is called from `handleAnnounce` for that specific owner. An entry whose owner never announces is never removed. So 64 packets from fabricated owners pin the stash for the rest of the process, and from then on every genuine bundle that races its owner's announce is dropped and never retried.

The consequence is quiet rather than loud: courier mail to that peer falls back to a v1 static-key seal instead of a one-time prekey, i.e. no forward secrecy for async first contact — the thing #1381 added.

### The fix

Move the bookkeeping into `BLEPendingPrekeyBundleStore`, bounded the same way `nostrPendingSubscriptions*` already is in this repo: **oldest-by-insertion eviction at the cap, plus an age sweep**. Replacing an owner's own entry still never counts as a new claimant, so a re-gossiped bundle can't evict anyone. `take(for:)` refuses an entry that outlived the TTL, since the owner re-gossips bundles anyway.

TTL is 120s against a 4s announce interval, so a genuine entry is drained many intervals inside it.

**What this does not fix, stated plainly:** an attacker sustaining a flood of distinct fabricated owners can still crowd out a genuine entry for as long as the flood lasts — any per-identity stash loses that race. What changes is recovery: the wedge is no longer permanent, and each 60s bundle gossip round gets a fresh chance. Making the stash unfloodable needs attribution that doesn't exist yet at that point in the pipeline.

### Verification

Xcode is not installed on my machine, so **I could not run `swift test` or build the app target** — CI is the first thing that will compile the `@Test` scaffolding in `BLEPendingPrekeyBundleStoreTests.swift`.

What I did run:

- Compiled and executed `BLEPendingPrekeyBundleStore.swift` against the real built `BitFoundation` module (`swift build --package-path localPackages/BitFoundation`, then linking its object files), with every assertion from the test file ported into a `main.swift` harness. **17/17 passed.**
- Re-ran the same harness against the old policy (refuse-at-cap, no expiry) pasted back in, to confirm the tests have teeth: **6 failed**, including `burst/genuineStashed` — the wedge itself — plus the eviction and sweep assertions.
- Typechecked the store under both `-swift-version 5` and `-swift-version 6` alongside a stub mimicking `BLEService`'s shape at both call sites (mutating the struct inside the non-escaping `onEngine` closure).

The three-line `BLEService.swift` change is verified by reading and by that stub typecheck, not by execution.

### DIFF
--- a/bitchat/Services/BLE/BLEPendingPrekeyBundleStore.swift
+++ b/bitchat/Services/BLE/BLEPendingPrekeyBundleStore.swift
@@ -1,0 +1,79 @@
+//
+// BLEPendingPrekeyBundleStore.swift
+// bitchat
+//
+// This is free and unencumbered software released into the public domain.
+// For more information, see <https://unlicense.org>
+//
+
+import BitFoundation
+import Foundation
+
+/// Holds prekey bundles that arrived before their owner's announce bound a
+/// signing key, so attribution can be retried when that announce lands.
+///
+/// Nothing about a stashed bundle has been verified — that is precisely why it
+/// is stashed — so every entry has to be treated as attacker-controlled. A
+/// bundle only needs a `senderID` matching the short ID derived from the
+/// `noiseStaticPublicKey` it carries to reach the stash, and both of those are
+/// public, so anyone in radio range can mint entries for owners that will never
+/// announce. The store therefore bounds itself two ways: oldest-by-insertion
+/// eviction at the cap, and an age sweep that drops entries whose owner never
+/// showed up. Announces repeat every few seconds, so a genuine entry is drained
+/// long before the TTL; an unattributable one is garbage by then.
+///
+/// Both bounds matter. Refusing new entries at a full cap — with no expiry —
+/// lets a single burst of unattributable bundles wedge the stash for the rest
+/// of the process, after which no bundle that races its owner's announce is
+/// ever retried again.
+struct BLEPendingPrekeyBundleStore {
+    struct Config {
+        /// Distinct owners that may hold a stashed bundle at once.
+        var capacity: Int = TransportConfig.prekeyBundlePendingCap
+        /// How long a stashed bundle waits for its owner's announce.
+        var ttlSeconds: TimeInterval = TransportConfig.prekeyBundlePendingTTLSeconds
+    }
+
+    private struct Entry {
+        let packet: BitchatPacket
+        let stashedAt: Date
+    }
+
+    private let config: Config
+    private var entries: [PeerID: Entry] = [:]
+
+    init(config: Config = Config()) {
+        self.config = config
+    }
+
+    var count: Int { entries.count }
+
+    /// Retains the latest bundle seen for `owner`. Replacing an owner's entry
+    /// never counts against the cap: a re-gossiped bundle must not be treated
+    /// as a new claimant.
+    mutating func stash(_ packet: BitchatPacket, for owner: PeerID, now: Date = Date()) {
+        sweepExpired(now: now)
+        let capacity = max(1, config.capacity)
+        if entries[owner] == nil {
+            while entries.count >= capacity {
+                guard let oldest = entries.min(by: { $0.value.stashedAt < $1.value.stashedAt })?.key else { break }
+                entries.removeValue(forKey: oldest)
+            }
+        }
+        entries[owner] = Entry(packet: packet, stashedAt: now)
+    }
+
+    /// Removes and returns `owner`'s stashed bundle, if one is still live. The
+    /// entry is dropped either way — a bundle that outlived its TTL is not
+    /// worth re-verifying, and the owner will gossip a fresh one.
+    mutating func take(for owner: PeerID, now: Date = Date()) -> BitchatPacket? {
+        guard let entry = entries.removeValue(forKey: owner) else { return nil }
+        guard now.timeIntervalSince(entry.stashedAt) <= config.ttlSeconds else { return nil }
+        return entry.packet
+    }
+
+    private mutating func sweepExpired(now: Date) {
+        guard !entries.isEmpty else { return }
+        entries = entries.filter { now.timeIntervalSince($0.value.stashedAt) <= config.ttlSeconds }
+    }
+}
--- a/bitchat/Services/BLE/BLEService.swift
+++ b/bitchat/Services/BLE/BLEService.swift
@@ -274,10 +274,9 @@ final class BLEService: NSObject {
     // Prekey bundles that arrived before their owner's verified announce bound
     // a signing key. Over the air a bundle can still arrive before the
     // announce it depends on; we retain the latest such bundle per owner
-    // (bounded) and re-attempt attribution when the announce lands.
-    // Engine-confined.
-    private var pendingPrekeyBundles: [PeerID: BitchatPacket] = [:]
-    private static let pendingPrekeyBundleCap = 64
+    // (bounded, with an age sweep) and re-attempt attribution when the announce
+    // lands. Engine-confined.
+    private var pendingPrekeyBundles = BLEPendingPrekeyBundleStore()
     // Gateway mode: sink for received nostrCarrier packets (set by app
     // wiring, called on the main actor after transport-level checks) and the
     // runtime-toggled capability bits ORed into `PeerCapabilities.localSupported`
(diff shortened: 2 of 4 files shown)


Chessing234:
awaiting-review

bitchat/Services/BLE/BLEService.swift
```diff
@@ -4932,10 +4930,12 @@ extension BLEService {
     /// Re-attempt any prekey bundle that arrived before this owner's announce
     /// bound a signing key. Called from handleAnnounce after a verified
     /// announce, in a barrier ordered after the registry write, so a bundle
-    /// stashed before the write is always observed here.
+    /// stashed before the write is always observed here. A bundle whose owner
+    /// took longer than the stash TTL to announce is dropped instead — the
+    /// owner re-gossips bundles, so waiting on a stale one buys nothing.
```
jozanek, line 4935: I think packet dedup outruns the re-gossip here. A gossip re-offer changes only `ttl` and `isRSR` (GossipSyncManager.swift:500-502), so `messageID` is unchanged and `dropDuplicatePacketIfNeeded` drops it at :5945 for `messageDedupMaxAgeSeconds` (300s). That leaves 120s–300s where the entry has expired and no copy can re-enter — and if the neighbour holding it leaves, recovery falls to the owner's hourly rebroadcast.

Would you be open to raising the TTL clear of that window? I'd welcome a correction if I've misread the RSR path.

bitchat/Services/TransportConfig.swift
```diff
@@ -410,4 +410,13 @@ enum TransportConfig {
...
     static let prekeyBundleRebroadcastSeconds: TimeInterval = 60 * 60
+    // A bundle that arrives before its owner's announce is stashed until that
+    // announce binds a signing key. Nothing about a stashed bundle is verified
+    // yet and anyone can mint one for an owner that never announces, so the
+    // stash is bounded both ways: oldest-by-insertion eviction at the cap, plus
+    // an age sweep. The TTL is many announce intervals
+    // (bleAnnounceIntervalSeconds), so a genuine bundle is always drained well
+    // inside it.
+    static let prekeyBundlePendingCap: Int = 64
+    static let prekeyBundlePendingTTLSeconds: TimeInterval = 120.0
```
jozanek, line 421: Two things I'd value your read on. 120s sits below `messageDedupMaxAgeSeconds` (300s), which gates the re-supply path this relies on. And eviction-at-cap alone already admits a genuine bundle every time, so the wedge looks fixed with no TTL — leaving the TTL as the only part that can discard a bundle that would still have verified (`PrekeyBundleStore.ingest` age-gates only on `generatedAt`).

Would 600s work, matching `nostrPendingSubscriptionTTLSeconds`?

bitchat/Services/TransportConfig.swift
```diff
@@ -410,4 +410,13 @@ enum TransportConfig {
     // announces, keep it alive in peers' gossip stores; changed bundles are
     // sent immediately.
     static let prekeyBundleRebroadcastSeconds: TimeInterval = 60 * 60
+    // A bundle that arrives before its owner's announce is stashed until that
+    // announce binds a signing key. Nothing about a stashed bundle is verified
+    // yet and anyone can mint one for an owner that never announces, so the
+    // stash is bounded both ways: oldest-by-insertion eviction at the cap, plus
+    // an age sweep. The TTL is many announce intervals
+    // (bleAnnounceIntervalSeconds), so a genuine bundle is always drained well
```
jozanek, line 418: `bleAnnounceIntervalSeconds` (4s) is the cadence for a node with no connected peers — `BLEMaintenancePolicy.swift:46` takes that branch only when `connectedCount == 0`. Connected owners announce at 15±4s sparse and 30±8s dense, and this file calls 38s the worst case at :172.

Could this cite those instead? A bundle arriving before its owner's announce means the owner is present and transmitting, so the 4s figure describes the one case the stash doesn't serve.

bitchat/Services/BLE/BLEPendingPrekeyBundleStore.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    /// Retains the latest bundle seen for `owner`. Replacing an owner's entry
+    /// never counts against the cap: a re-gossiped bundle must not be treated
+    /// as a new claimant.
+    mutating func stash(_ packet: BitchatPacket, for owner: PeerID, now: Date = Date()) {
+        sweepExpired(now: now)
+        let capacity = max(1, config.capacity)
+        if entries[owner] == nil {
+            while entries.count >= capacity {
+                guard let oldest = entries.min(by: { $0.value.stashedAt < $1.value.stashedAt })?.key else { break }
```
jozanek, line 59: `NostrRelayManager` logs both of its bounds at warning level with a "Bounds proof" comment (:1036, :1057), so an operator can see the bound actually firing. An eviction here means someone is flooding the stash, which seems at least as worth surfacing — and without it, a fix for a condition nobody has observed stays unobservable.

Would you add a `SecureLogger.warning` here and on the sweep?

bitchat/Services/BLE/BLEPendingPrekeyBundleStore.swift
```diff
@@ -0,0 +1,79 @@
...
+    struct Config {
+        /// Distinct owners that may hold a stashed bundle at once.
+        var capacity: Int = TransportConfig.prekeyBundlePendingCap
+        /// How long a stashed bundle waits for its owner's announce.
+        var ttlSeconds: TimeInterval = TransportConfig.prekeyBundlePendingTTLSeconds
+    }
+
+    private struct Entry {
+        let packet: BitchatPacket
+        let stashedAt: Date
```
jozanek, line 39: The precedent orders by an explicit monotonic `sequence` counter (NostrRelayManager.swift:1031) rather than a timestamp. Ordering by wall-clock `Date` means a backward NTP step can make a newer entry look oldest and be evicted first.

Worth an insertion counter, or is the window too small to care about here?

jozanek (commented):
Thanks for the limitations section and the negative control — both made this quick to review.

**What I understood this to be for:** a bundle can outrun the announce that binds its owner's signing key, so it's stashed unverified — and because nothing is verified at that point, 64 fabricated owners pin the map for the life of the process, silently downgrading courier mail to a v1 static-key seal. If I've read that wrong, discount the rest.

The bug looks real to me, and two things make it worse than you claim: `.prekeyBundle` at :6019 doesn't guard the relay step the way `.boardPost` and `.voiceFrame` do, so a fabricated bundle spreads up to 7 hops; and the background modes in Info.plist mean the process outlives backgrounding, so "permanent" reads as accurate.

I pushed on the `onEngine` seam since it's the part you couldn't execute — it holds. `messageQueue` is serial (:375) and the closure is non-escaping, so the mutating call writes back to the same storage the drain reads. Side note: the comments at :4887 and :4932 still say "concurrent" and "barrier", which stopped being true in #1498 — worth refreshing while you're in there.

**My one substantive concern is the TTL**, left inline: 120s sits below the 300s dedup window that blocks re-supply, and eviction-at-cap alone appears to fix the wedge without any TTL at all. Raising it to 600s, or dropping it, would settle both.

**What I couldn't verify:** anything needing a build or a real radio — I didn't compile or run tests either, so CI has the last word.

One suggestion: `bitchatTests/EndToEnd/PrekeyEndToEndTests.swift` already has what's needed to reproduce the wedge end-to-end (`_test_handlePacket`, and a helper that splits the announce from the bundle), and needs no Xcode — that would cover the three-line change the store tests don't reach.

PR #1689, opened 2026-08-25
fix: refuse an ambiguous nickname instead of picking the first peer

Chessing234:
## Summary
- `UnifiedPeerService.getPeerID(for:)` and geohash `geoNicknames.first(where:)` returned the first match, so `/msg` and `/block` could hit the wrong person when two peers shared a nickname.
- An unsuffixed name is accepted only when it identifies exactly one peer. Colliding names need the people-list `#xxxx` suffix.

## Test plan
- [x] `swiftc` typecheck of `NicknameLookup.swift` under Swift 5 and 6
- [x] local harness against `NicknameLookup` (unique, colliding, `#xxxx` suffix, unknown)
- [ ] `xcodebuild` test suite — this machine has Command Line Tools only, no Xcode; CI covers the new `UnifiedPeerService` / `ChatPeerIdentityCoordinator` tests

### DIFF
--- a/bitchat/Services/UnifiedPeerService.swift
+++ b/bitchat/Services/UnifiedPeerService.swift
@@ -251,17 +251,21 @@ final class UnifiedPeerService: ObservableObject, TransportPeerEventsDelegate {
         return peerIndex[peerID]
     }
     
-    /// Get peer ID for nickname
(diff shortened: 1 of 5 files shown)


Chessing234:
awaiting-review

PR #1690, opened 2026-08-25
fix: enable composer autocorrect except on / @ # tokens

Chessing234:
## Summary
- The composer kept `.autocorrectionDisabled(true)` for every keystroke so `/commands`, `@mentions`, and `#geohash` tokens would not be rewritten. Prose was paying that cost (#969).
- Disable autocorrect only while the token under the cursor starts with `/`, `@`, or `#`.

## Test plan
- [x] `swiftc` typecheck of `ComposerAutocorrectPolicy.swift` under Swift 5 and 6
- [x] local harness (prose, `/`, `@`, `#`, completed `/msg` argument, token at cursor)
- [ ] `xcodebuild` test suite — this machine has Command Line Tools only, no Xcode; CI covers `ComposerAutocorrectPolicyTests` and the composer wiring
- [ ] type in the composer on a device: a sentence should autocorrect; `/msg`, `@al`, and `#u4pr` should not

### DIFF
--- a/bitchat/Utils/ComposerAutocorrectPolicy.swift
+++ b/bitchat/Utils/ComposerAutocorrectPolicy.swift
@@ -1,0 +1,24 @@
+import Foundation
+
+/// When the composer should suppress system autocorrect.
+///
+/// `/commands`, `@mentions`, and `#geohash` tokens need the exact typed text
+/// for autocomplete. Prose should still get autocorrect. The token under the
+/// cursor is the substring after the last whitespace before `cursor`.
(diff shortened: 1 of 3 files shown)


jackjackbits:
Closing in favor of #1508 to consolidate composer autocorrect work. #1508 has the more complete keyboard-trait and timer lifecycle handling. Preserve this PR's explicit mid-token and completed-command-argument tests in the remaining work; both implementations still need to use the actual insertion point when protecting /, @, and # tokens.

PR #1691, opened 2026-08-25
fix: encode photos up to the 512 KiB image budget

Chessing234:
## Summary
- `ImageUtils` crushed JPEGs to 45 KB at 448px even though `FileTransferLimits.maxImageBytes` is already 512 KiB. Voice notes travel at that budget; Android encodes at 512px / q=85 with no extra squeeze.
- Encode against the protocol budget, default the long edge to 512, and always redraw so EXIF orientation is baked in for camera stills that already fit.

## Test plan
- [x] `swiftc` typecheck of `ImageUtils.swift` under Swift 5 and 6
- [x] local encode harness: a noisy 1600×1200 source landed at 286256 bytes (old crush 45 KB; cap 512 KiB)
- [ ] `xcodebuild` test suite — this machine has Command Line Tools only, no Xcode; CI covers `processImage_usesThe512KiBBudgetInsteadOfCrushingTo45KB`
- [ ] send a camera photo over mesh and confirm it stays readable on both ends

### DIFF
--- a/bitchat/Features/media/ImageUtils.swift
+++ b/bitchat/Features/media/ImageUtils.swift
@@ -1,3 +1,4 @@
+import BitFoundation
 import Foundation
 import ImageIO
 import UniformTypeIdentifiers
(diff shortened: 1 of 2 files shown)


Chessing234:
awaiting-review

PR #1697, opened 2026-08-29
Refuse to encode payloads above the framed file cap

Chessing234:
## Summary
- `BinaryProtocol.encode` only checked that a compressed size fit in the v1/v2 length field (`UInt16`/`UInt32`), so it could emit frames whose declared `originalSize` exceeds `FileTransferLimits.maxFramedFileBytes`.
- Decode already rejects those frames (often silently), which is the Android→iOS drop mode in #1618/#1629.
- Refuse to encode above the same ceiling the decoder enforces; update the oversized-payload test and add a round-trip at the cap.

## Test plan
- [ ] `BinaryProtocolTests` oversized + framed-cap round-trip
- [ ] Confirm encode returns `nil` for `maxFramedFileBytes + 1`

Refs #1618 #1629

### DIFF
--- a/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
+++ b/localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
@@ -137,6 +137,12 @@ public struct BinaryProtocol {
         var payload = packet.payload
         var isCompressed = false
         var originalPayloadSize: Int?
+        // Refuse to emit a frame whose expanded (or plain) payload we would reject on decode.
+        // Decode enforces FileTransferLimits.maxFramedFileBytes; encode previously only checked
+        // the on-wire length field width, so v2 could compress up to UInt32.max and produce
+        // frames peers (including ourselves) silently drop.
+        guard payload.count <= FileTransferLimits.maxFramedFileBytes else { return nil }
+
         if CompressionUtil.shouldCompress(payload) {
             // Only compress when we can represent the original length in the outbound frame
             let maxRepresentable = version == 2 ? Int(UInt32.max) : Int(UInt16.max)
--- a/localPackages/BitFoundation/Tests/BitFoundationTests/BinaryProtocolTests.swift
+++ b/localPackages/BitFoundation/Tests/BitFoundationTests/BinaryProtocolTests.swift
@@ -346,8 +346,29 @@ struct BinaryProtocolTests {
             ttl: 1,
             version: 2
         )
-        let encoded = try #require(BinaryProtocol.encode(packet), "Failed to encode oversized packet")
-        #expect(BinaryProtocol.decode(encoded) == nil)
+        // Encode must refuse the same ceiling decode enforces — otherwise we emit
+        // frames that every honest peer (including ourselves) silently drops.
+        #expect(BinaryProtocol.encode(packet) == nil)
+    }
+
+    @Test("Round-trip a payload exactly at the framed file cap")
+    func framedFileCapRoundTrip() throws {
+        let targetSize = FileTransferLimits.maxFramedFileBytes
+        var payload = Data(repeating: 0x41, count: targetSize)
+        // Highly compressible so the frame stays wire-friendly after compression.
+        let packet = BitchatPacket(
+            type: MessageType.message.rawValue,
+            senderID: Data(hexString: "0011223344556677") ?? Data(),
+            recipientID: nil,
+            timestamp: UInt64(Date().timeIntervalSince1970 * 1000),
+            payload: payload,
+            signature: nil,
+            ttl: 1,
+            version: 2
+        )
+        let encoded = try #require(BinaryProtocol.encode(packet), "cap-sized payload should encode")
+        let decoded = try #require(BinaryProtocol.decode(encoded), "cap-sized payload should decode")
+        #expect(decoded.payload == payload)
     }
     
     // MARK: - Message Padding Tests


localPackages/BitFoundation/Tests/BitFoundationTests/BinaryProtocolTests.swift
```diff
@@ -346,8 +346,29 @@ struct BinaryProtocolTests {
             ttl: 1,
             version: 2
         )
-        let encoded = try #require(BinaryProtocol.encode(packet), "Failed to encode oversized packet")
-        #expect(BinaryProtocol.decode(encoded) == nil)
+        // Encode must refuse the same ceiling decode enforces — otherwise we emit
+        // frames that every honest peer (including ourselves) silently drops.
+        #expect(BinaryProtocol.encode(packet) == nil)
```
jozanek, line 351: This flips the assertion from `decode(encoded) == nil` to `encode(packet) == nil`. As far as I can trace, that leaves `BinaryProtocol.swift:333` — the cap check that runs on every untrusted inbound frame — without direct coverage; the v1 tests at `:618`/`:652` can't reach a 1,179,760 ceiling through a `UInt16` field, and `ChatViewModelExtensionsTests.swift:381` is stopped by the Nostr prefilter first.

The old fixture cycles all 256 byte values, so `shouldCompress` returned false and it was exercising the uncompressed `:333` path specifically. Would you be open to keeping that as a second test that hands a hand-built oversized frame straight to `decode`, so both sides stay covered?

localPackages/BitFoundation/Sources/BitFoundation/BinaryProtocol.swift
```diff
@@ -137,6 +137,12 @@ public struct BinaryProtocol {
         var payload = packet.payload
         var isCompressed = false
         var originalPayloadSize: Int?
+        // Refuse to emit a frame whose expanded (or plain) payload we would reject on decode.
+        // Decode enforces FileTransferLimits.maxFramedFileBytes; encode previously only checked
+        // the on-wire length field width, so v2 could compress up to UInt32.max and produce
+        // frames peers (including ourselves) silently drop.
+        guard payload.count <= FileTransferLimits.maxFramedFileBytes else { return nil }
```
jozanek, line 144: This runs before the compression branch, so it bounds `payload.count`, while decode bounds `payloadLength` (`compressedSize + 4`). For a cap-sized payload that compresses by only a few bytes, `payloadDataSize` can still exceed the cap and decode rejects at `:333` — the case the comment just above says is now prevented.

I couldn't construct a realistic payload in that window, since zlib returns nil rather than a 1-byte saving, so I think this is a placement issue rather than a live bug. Codex flagged the same thing as P2 on this line. Would you be open to also checking `payloadDataSize` after compression, next to the width checks at `:171-172`?

localPackages/BitFoundation/Tests/BitFoundationTests/BinaryProtocolTests.swift
```diff
@@ -346,8 +346,29 @@ struct BinaryProtocolTests {
...
+    }
+
+    @Test("Round-trip a payload exactly at the framed file cap")
+    func framedFileCapRoundTrip() throws {
+        let targetSize = FileTransferLimits.maxFramedFileBytes
+        var payload = Data(repeating: 0x41, count: targetSize)
```
jozanek, line 357: `payload` is only read here — passed at `:364` and compared at `:371` — so this will raise a "never mutated; consider changing to `let`" warning in the BitFoundation job. Would you change it to `let`?

jozanek (commented):
Thanks for this — and for the duplicate-spotting across #1620/#1623/#1682.

**What I understood this to be for:** making the codec refuse to emit a frame it would itself reject on decode. That seems a reasonable instinct, so my comments are about shape rather than direction.

**My main concern** is the test change, left inline on `:351`. `encode` only ever runs on our own outbound data; `decode` runs on every untrusted inbound frame. Swapping the assertion moves coverage off the attacker-facing guard, and because both guards are live code, neither Periphery nor a green CI run will show it.

**Three questions:**

1. #1618/#1629 both describe an Android *sender* dropped by an iOS *decode*. This changes iOS *encode*, so I don't think it can affect either — is there a failure mode I'm missing? If not, would you be open to rewording the summary, so a later reader doesn't take this as fixing that drop?
2. `maxFramedFileBytes` is a whole-*frame* budget (`maxPayloadBytes + tlvEnvelopeOverhead + binaryEnvelopeOverhead`) applied here to *payload* bytes; `NoiseSecurityConstants.swift:24-30` subtracts the envelope before using the same constant. Would a payload-derived budget fit better?
3. A refusal would be silent at six of the seven `nil` call sites. Would you consider a `SecureLogger.warning`, the way the ratio guard at `:374` does?

**What I couldn't verify:** any app path that reaches the guard. Everything I traced is bounded below it — file TLV at 1,179,664, private-file ciphertext at 1,179,728, largest non-file payload at 65,555. If you know a caller that can exceed it, that would change my read of the whole PR.

One bit of context: on #1620, jackjackbits said #1618 is the venue for the shared ceiling and that #1623 lands first — worth confirming that sequencing, since this adds the first test pinning `maxFramedFileBytes` as a supported payload size.

**What would move me to approve:** re-target the guard at the emitted frame size, keep a test that still exercises decode's `:333`, log the refusal, and adjust the issue attribution. Happy to re-review.

Chessing234:
awaiting-review

PR #1705, opened 2026-09-03
Attribute reassembled sync replies to the link's bound peer

heyaim:
Gossip sync exists to carry the public history a peer missed. A message too long for one BLE write, at most 512 bytes, from an author two or more hops away did not arrive that way once it was more than two minutes old. A reply that fits in one write from the same author over the same link arrives.

A sync reply that arrives in one BLE frame is validated against the peer bound to the link it came in on: `BLEIngressLinkRegistry` attributes a flagged frame to the bound peer, and `RequestSyncManager` checks that we asked that peer for sync. That is the rule `docs/REQUEST_SYNC_MANAGER.md` states: a flagged reply is exempt from the timestamp check only when it answers a pending request sent to that specific peer. A reply too large for one frame took a different path: each fragment was judged against the bound peer and passed, but the dispatch handed `BLEFragmentHandler` only the fragment's claimed sender, and the handler validated the reassembled packet against its inner sender, the original author in both cases. With no open request to the author, the reply was dropped. A flagged packet never reaches the skew check. The reply's age made no difference.

Requests go only to direct neighbors: on the 30-second maintenance tick, one second after a neighbor connects, and on stalled-fragment resyncs, each against a 30-second window. Two-device testing therefore hides the drop: the author is always a neighbor. The fragment round narrows it further: re-served author fragments reassemble into a packet judged by its own timestamp, which still delivers trains under two minutes old. The requester never asks an author two or more hops away. Every multi-frame reply older than two minutes from such an author was lost, among them long public and group messages, file transfers and board posts.

## Changes

- **The dispatch passes the link's bound peer.** `attributeAndHandlePacket` carries it through `handleReceivedPacket` to `handleFragment`, and `BLEFragmentHandler.handle` takes it as an optional `boundTo:` parameter after the claimed sender: nil for an unbound link or a re-injected packet.
- **A flagged reassembled packet is judged against the bound peer, or against its own sender when there is none.** That is the rule `BLEIngressLinkRegistry` applies to a single flagged frame. The fragments' claimed sender is never the anchor: the signature and the direct path both refer to the packet's own sender, and anyone can name a peer we asked. An unflagged packet is still judged against its author.
- **Self-fragment suppression and re-injection are unchanged.** Both keep the fragments' claimed sender, which is the author for a planner-built train. Re-injection hands the packet on as that peer's, as it did before, and it refreshes that peer's last-seen time; moving it would change that bookkeeping for all fragment traffic.
- **One DEBUG-only test hook.** `_test_registerSyncRequest(to:)` records the pending request a `REQUEST_SYNC` would record, which lets a test present a solicited reply without the timer-driven send.

## Tests

- `BLEFragmentHandlerTests`: planner-built fragments of a stored broadcast message marked as a sync response, with the assembly buffer, `BLEIngressPacketGuard` and a `RequestSyncManager`. The accepted reply passes the ingress guard and reaches the handler with the same `from` and `boundTo` the dispatch passes, with a request to the serving peer only. The two rejection cases feed the handler directly: the same reply with no request registered is dropped, and an unflagged train from the same author is still judged on the author.
- `reassemblyFollowsTheDirectPathsAcceptanceRule` treats the direct path as the specification: across 256 runs covering 192 distinct combinations of flag placement, request state, age, author, link binding (bound or none) and the sender the fragments claim (the author or another peer), reassembly accepts a packet only when the same packet would pass as a single frame on the same link, and a fully admitted train is accepted exactly when that frame would be.
- `SimulatedMeshTests`: the production ingress path end to end, in a line of three peers where the author is two hops from the requester, known from a relayed announce and never asked. Her signed reply arrives as fragments on the serving peer's bound link, with a request held to that peer, and is delivered. An equivalent train on a never-announced link, with fragments naming the serving peer, is judged by its own sender and dropped. The tests split the reply with the planner, since the simulated mesh has no physical links.

## Validation

- `swift test --parallel --skip PerformanceBaselineTests` (the CI app job's test selection, run without coverage): 2024 Swift Testing tests, exit 0.
- iOS simulator suite, scheme "bitchat (iOS)", `-parallel-testing-enabled NO`, iPhone 17 Pro simulator: 2237/2237, 0 failed, 0 skipped.
- `swiftlint lint --strict`: no violations.
- Reverting the rule to the author fails three tests, a SimulatedMesh test among them. Letting an unbound link fall back to the fragments' claimed sender fails the 256-combination test and the unbound-link SimulatedMesh test. Dropping the binding at the dispatch fails the two-hop SimulatedMesh test; passing the received-from peer as the binding fails the unbound-link one.
- No hardware run; the SimulatedMesh tests are the end-to-end check of the receive path.

## What this does not change

- Unflagged fragment trains are judged on the author's timestamp, as before. The fragment round is untouched.
- The 30-second response window still applies to every flagged fragment and to the reassembled packet.
- The single-frame path's decisions and the wire format are unchanged.
- The assembly buffer keys a train on its sender and fragment ID and does not record links. A train whose fragments arrived over more than one link is judged by the link that completed it.


Chessing234 (commented):
test comment permission

Chessing234 (commented):

heyaim:
Merged main after #1717 and #1719. #1719's type check now runs before the bound-peer check, and its two tests now take the bound peer. Description updated for #1717 and the new test counts.

PR #1706, opened 2026-09-04
fix: accept Android-aged gift wraps; gate outer timestamps before decrypt

Chessing234:
## Summary
- Android stamps outer Nostr gift wraps up to 48h into the past (`randomizeTimestampUpToPast`). iOS subscribed with a 24h `since` filter, so relays that honor `since` never delivered those DMs.
- Add `nostrGiftWrapMaxAgeSeconds` (48h + 15min skew) and use it for gift-wrap subscriptions (geo DM + account mailbox). Keep the inner rumor window at 24h + skew.
- Reject implausible outer `created_at` values *before* decrypt so a hostile relay cannot force ECDH work with ancient or far-future wraps.

## Test plan
- [x] Extended `NostrInboundPipelineTimestampTests` (accept 48h-old outer; reject far-future / over-age; rumor window unchanged)
- [x] Standalone `swiftc` assertions of the same policy math (no Xcode on this machine)
- [ ] CI Swift / iOS simulator suites

### DIFF
--- a/bitchat/Services/MessageDeduplicationService.swift
+++ b/bitchat/Services/MessageDeduplicationService.swift
@@ -171,10 +171,11 @@ final class MessageDeduplicationService {
     private let nostrAckCache: LRUDeduplicationCache<Bool>
 
     /// Optional cross-launch persistence for the Nostr event cache. BitChat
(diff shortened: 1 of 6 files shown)


Chessing234:
awaiting-review

PR #1707, opened 2026-09-04
fix: randomize seal/gift-wrap timestamps up to 48h into the past

Chessing234:
## Summary
- iOS stamped seal and gift-wrap `created_at` with a ±15 minute jitter around now.
- Android samples uniformly from the past 48 hours (`randomizeTimestampUpToPast`), which is what the wider gift-wrap subscription window exists for.
- Match that past-only distribution so iOS envelopes get the same timing-correlation resistance.

Stacked on #1706 (receive-path lookback + outer timestamp gate). That lookback is required once iOS starts emitting wraps older than 24h; if #1706 merges first, rebase drops the shared commit cleanly.

## Test plan
- [x] `NostrEnvelopeTimestampRandomizationTests` — 200 samples stay in `[now−48h, now]`; zero/negative maxPast collapse to `now`
- [x] Standalone `swiftc` assertions of the same sampler
- [ ] CI Swift / iOS simulator suites

### DIFF
--- a/bitchat/Nostr/NostrProtocol.swift
+++ b/bitchat/Nostr/NostrProtocol.swift
@@ -745,25 +745,23 @@ struct NostrProtocol {
         return sharedSecretData
     }
     
+    /// Randomized `created_at` for seals and gift wraps. Samples uniformly
+    /// from `[now − maxPast, now]` (never future) so relays cannot correlate
+    /// envelope timing with the true send time carried on the inner rumor.
+    /// Matches Android `NostrCrypto.randomizeTimestampUpToPast`. Internal so
+    /// tests can pin `now` / `maxPast`.
+    static func randomizedEnvelopeTimestamp(
+        now: Date = Date(),
+        maxPast: TimeInterval = TransportConfig.nostrGiftWrapTimestampRandomizationSeconds
+    ) -> Date {
+        let clampedPast = max(0, maxPast)
+        // Inclusive of both endpoints, matching Android's nextInt(maxPast+1).
+        let offsetSeconds = Int.random(in: 0...Int(clampedPast))
+        return now.addingTimeInterval(-TimeInterval(offsetSeconds))
+    }
+
     private static func randomizedTimestamp() -> Date {
-        // Add random offset to current time for privacy
-        // This prevents timing correlation attacks while the actual message timestamp
-        // is preserved in the encrypted rumor
-        let offset = TimeInterval.random(in: -900...900) // +/- 15 minutes
-        let now = Date()
-        let randomized = now.addingTimeInterval(offset)
-        
-        // Log with explicit UTC and local time for debugging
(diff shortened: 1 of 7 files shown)


Chessing234:
awaiting-review

bitchat/Services/TransportConfig.swift
```diff
@@ -215,11 +215,22 @@ enum TransportConfig {
...
     static let nostrDMSubscribeLookbackSeconds: TimeInterval = 86400
-    // Tolerated clock skew for the client-side rumor-timestamp window on
-    // inbound Nostr DMs (senders stamp the inner rumor with real time; only
-    // the outer gift wrap is randomized per NIP-17).
+    // Tolerated clock skew for client-side Nostr DM timestamp windows.
     static let nostrDMMaxClockSkewSeconds: TimeInterval = 900
+    /// Outer gift-wrap / seal `created_at` randomization floor. Matches
+    /// Android `NostrCrypto.randomizeTimestampUpToPast` (up to 48h into the
+    /// past). The actual message time lives on the inner rumor.
+    static let nostrGiftWrapTimestampRandomizationSeconds: TimeInterval = 172_800
```
jozanek, line 225: Half a 48h draw falls below `now − 24h`, but every installed build subscribes `since: now − 86400` (GeohashSubscriptionManager.swift:167/:265/:393) — so ~half of DMs and acks to a not-yet-updated peer never arrive, silently.

Would you ship at 85,500 (24h − skew) for one release and raise it after #1706 saturates?

bitchat/Nostr/NostrProtocol.swift
```diff
@@ -745,25 +745,23 @@ struct NostrProtocol {
         return sharedSecretData
     }
     
+    /// Randomized `created_at` for seals and gift wraps. Samples uniformly
```
jozanek, line 748: The seal is encrypted inside the wrap (:507-518), and `decryptPrivateMessage` returns `rumor.created_at` (:169), never `seal.createdAt` — so randomizing the seal isn't observable to anyone. Narrow the comment to the wrap?

jozanek (commented):
Direction looks right to me — past-only avoids future-dated `created_at`, which is the bound relays actually enforce.

What stops me is sequencing. Half of a 48h uniform draw lands below `now − 24h`, but every installed build subscribes `since: now − 86400`, so ~half of DMs and acks to a peer who hasn't updated get dropped at the relay while the UI says "sent". #1706 fixes future receivers, not installed ones. Suggestion inline.

One question: `giftWrapsFor` sets `limit = 100` (NostrRelayManager.swift:1870), and NIP-01 returns the most recent `limit` *by `created_at`* — doesn't a wrap stamped 47h back then lose its place to older-stamped ones?

Also WHITEPAPER.md:87 still says "±15 minutes".

PR #39, opened 2025-07-08
Fixes Xcode Project so that test are running again

dasmer:
Replace `PasswordProtectedRoomTests.swift` with `PasswordProtectedChannelTests.swift` in Xcode project file so tests run

Fixes #36

Issue #939, opened 2026-01-12
[Cross-Platform Bug] Android → iOS communication fails due to packet serialization mismatch


Timi16:
## Problem Description

Android devices can receive messages from iOS, but iOS cannot receive messages from Android. This creates one-way communication and breaks the core mesh networking feature.

### Observed Behavior
- ✅ iOS → Android: **Works**
- ❌ Android → iOS: **Fails** (messages invisible or cause crashes)
- ✅ Bluetooth LE connection: **Establishes successfully**
- ❌ Message parsing: **Fails on iOS side**

### Affected Versions
- iOS: [specify version you're testing]
- Android: [specify version]

### Related Issues
- #143 (Android repo) - DM from Android to iOS crashes
- #355 - Cross-platform communication issues
- #272 - Messages not visible between platforms

---

## Root Cause Analysis

This is **not** a Bluetooth connectivity issue - the BLE stack connects fine. The problem is at the **protocol serialization layer**.

### Hypothesis
iOS and Android implementations serialize packets differently:

1. **Byte Ordering**: Swift vs Kotlin may use different endianness for multi-byte integers
2. **Length Prefixes**: String/array length encoding might differ (varint vs fixed-width)
3. **Field Alignment**: Struct padding or field ordering could be inconsistent
4. **Packet Framing**: Different approaches to MTU fragmentation/reassembly

### Evidence
- iOS can parse Android packets → Android serialization is more permissive
- Android cannot parse iOS packets → iOS uses stricter/different format
- No crash logs suggest silent parsing failures (packets dropped)

---

## Proposed Investigation Steps

1. **Capture BLE Traffic**
   - Use Wireshark/nRF Connect to sniff packets
   - Compare hex dumps of identical messages from iOS vs Android
   - Document byte-level differences

2. **Add Debug Logging**
   - Log raw packet bytes before encryption
   - Log decrypted payload on receiver side
   - Compare field-by-field deserialization

3. **Create Test Packets**
   - Generate known-good test packets
   - Verify both platforms can parse the same binary data
   - Add to integration test suite

---

## Proposed Fix

Once we identify the serialization mismatch, we need to:

1. **Standardize the Wire Format**
   - Document the canonical packet structure
   - Specify byte ordering, length encoding, and field alignment
   - Add this to `WHITEPAPER.md`

2. **Update Non-Compliant Implementation**
   - Fix whichever platform diverges from the spec
   - Ensure backward compatibility if possible

[...]

PR #1708, opened 2026-09-08
Bind a trust seal to the nickname it was earned under

shubhambhandari126:
## The gap

A vouched or verified key can rename itself onto a nickname the user trusts and keep rendering the seal beside the new name.

`VouchAttestation.signableBytes` covers `voucheeFingerprint | voucheeSigningKey | timestampMs` and says nothing about a name — which is right, the attestation should stay name-free. But the seal is *rendered* beside a self-claimed nickname, and nothing binds the two:

1. Eve announces `ravi`, meets someone who verifies her, and earns one legitimate vouch.
2. Eve announces `medic`. The rename is free and silent — the only trace is a `SecureLogger.debug` line in `BLEAnnounceHandler`.
3. Every device that trusts the real medic now shows a second trusted-looking medic.

This is the mirror of #1661 — that one is *same nickname, new key*; this is *same key, new nickname*.

## Where the seal actually comes from

Three sources, all now gated:

| Surface | Draws the seal from |
| --- | --- |
| Peer list, peer **offline** | `showsVerifiedBadgeWhenOffline` / `showsVouchedBadge` in `PeerListModel` |
| Peer list, peer **connected**, and the DM header | `EncryptionStatus.noiseVerified.icon` — the same `checkmark.seal.fill` glyph |
| Public timeline and DM rows | `isVerifiedSender` / `showsVerifiedSeal` |

The connected-peer path is the live impersonation and it is the one that runs through `getEncryptionStatus`, so the binding is applied there: `.noiseVerified` demotes to `.noiseSecured` when the key renamed since trust. The session really is encrypted; only the claim about *who* is withdrawn.

No *behaviour* branches on the distinction — `VerificationModel:33`, `ContentSheetViews:659` and `:760` all treat `.noiseVerified` and `.noiseSecured` alike. Four things the user perceives do, and all four are meant to move: the icon, the status caption, the accessibility description (so VoiceOver stops saying "verified" on a demoted DM header), and the green tint on the nickname at `FingerprintView:87` — the one site that does *not* treat the two statuses identically. An earlier revision of this description claimed "nothing moves"; @RodsF1 showed that was 3-for-4, and it is corrected here and in the code comment that repeated it.

The demotion is computed on read, not stored: the cached status is invalidated by verification and session changes but never by a rename, so a demotion written into the cache would be stale and sticky.

## The binding

`IdentityCache` gains `trustedNicknames` — the nickname a key was announcing when this device decided to trust it. Pinned:

- **On the first vouch only**, never moved by a later one. Otherwise the attacker renames and the next voucher silently re-anchors the seal to the new name.
- **Re-verifying does move it** — the user just checked this key in person, under whatever name it shows now.
- **On the first announce seen while trust already stands.** A vouch normally arrives for a peer we have not seen announce: it comes over Noise from someone else and the vouchee may be hops away, so there is no name to bind to at record time. The announce path deliberately does *not* pin before trust exists — binding to whatever name we happened to see first would suppress the seal of a peer who renamed *before* being vouched.
- **Never on an empty nickname, and a missing baseline never suppresses.** Peers trusted by earlier builds have none, and dropping their seals on upgrade would teach users to ignore the signal.

## Two different questions, two different folds

Getting these backwards is how a vouch for one name comes to cover another, so both are named, both are tested, and neither is reused for the other's job.

| Question | Fold | Used by |
| --- | --- | --- |
| Is this the **same name**? | `nicknameBindingKey` — NFC, case, NFC again | the trust binding |
| Do these two **look alike**? | `PeerDisplayNameResolver.collisionKey` — NFKC + case | the `#abcd` disambiguation suffix |

The binding key is deliberately NFC and **not** NFKC: a fullwidth `Ｍedic` merely *looks* like `Medic`, so it is a different name and must break the binding. It case-folds because recasing your own nickname is not a rename — and it re-normalises after folding, because case folding can itself emit decomposed sequences.

The collision key folds the opposite way, and that is @RodsF1's one-liner from review. Folding the **count** key (never the displayed name) closes five categories for free: case, fullwidth, mathematical bold, script `ℳ` and roman-numeral `Ⅿ`. It also makes explicit the canonical folding Swift's `String` keys were already giving incidentally. Cross-script homoglyphs (Cyrillic `М`, Greek `Ο`) still get through; that needs a UTS #39 confusables table, which is its own conversation and not proposed here. Both gaps are pinned as tests so whoever adds the table has something to flip.

Collisions are counted in **two** places — `PeerDisplayNameResolver` and `BLEPeerSenderDisplayName.collisionResolvedName`, which feeds sender names on file transfers. Both had the same exact-match blind spot; folding only the first would have left the second behind.

## It compares announced names, never rendered ones — except on a row

`peer.nickname` is already collision-resolved: `BLEPeerRegistry.transportSnapshots` runs names through `PeerDisplayNameResolver`, which appends `#` plus four hex of the peerID to **connected** peers whose nicknames collide. That is exactly this attack — two connected peers both claiming `medic` — so comparing the displayed string would turn the real medic into `medic#c3d4`, mismatch the pinned `medic`, and **drop the impersonated peer's seal at the worst possible moment.**

So the live check takes no rendered name at all: it compares the pinned baseline to the peer's own announced nickname. A local petname short-circuits it, since a petname outranks the claimed nickname wherever it is displayed and there is nothing to spoof.

A message row asks a different question, and an earlier revision got it wrong. The row renders `message.sender`, frozen at receipt, while the live check gates on the name announced *now* — so: verified as `ravi` → announce `medic` → post → announce `ravi` again, and the archived row renders `<@medic ✓>`, a seal beside a name the key was never trusted under. `sealAppliesToRow(fingerprint:renderedSender:)` therefore checks each row against its own sender and never consults the live name. Requiring both would have retroactively unsealed accurate history every time anyone changed their nickname: a historical `ravi ✓` stays sealed while that key is currently renamed, because that row really was ravi.

Comparing a rendered name needs the suffix off without `splitSuffix()`'s `@` problem — it strips every `@` in the string, which is right for parsing a mention and wrong for comparing a nickname, since `validateUserString` rejects only control characters and `ravi@hq` would compare unequal to itself. Hence `withoutCollisionSuffix`, which removes only a trailing `#` plus four **ASCII** hex digits: `Character.isHexDigit` is also true for fullwidth digits, so a nickname ending in `#ＡＢＣＤ` would otherwise be truncated into something that could match a baseline it is not.

There is a partial mitigation already in the tree, and it is worth being precise about rather than claiming none exists: `PeerDisplayNameResolver` does disambiguate namesakes — but it counts collisions only among *connected* peers and suffixes only connected ones, so it does nothing when the impersonated peer is offline or out of range, and never applies to offline favourite rows. The hole is exactly where this attack aims.

## Scope, and one thing this is not

No wire change, no protocol change, no new user-facing strings. `trustedNicknames` decodes with `decodeIfPresent` like the other post-v1 fields, so caches written by older builds still load.

**This is display-only, and that is a deliberate boundary rather than an oversight.** A renamed key loses its seals but keeps its authority: `recordVouch` still counts it as a voucher, and it is still advertised in outgoing batches. @RodsF1 argued that should change. The reason it does not is that `trustedNicknameMismatch` fires on **any** rename, not on impersonation — gating authority on it would mean your vouches silently stop counting on every device that verified you the moment you change your nickname, with nothing shown to you or to them. The trigger that would justify revoking authority is the narrow one, *this key renamed onto a name another trusted key already holds*, and BitChat has no cross-peer name comparison to detect it. So the honest statement of the finding is "there is no impersonation detector here, only a rename detector". If you read the false-positive budget the other way, say so and I will implement whichever trigger you want — it is your product.

(I checked my own project rather than only arguing, found the identical boundary there, and fixed it — but there a real impersonation detector exists to trigger it.)

**`UnifiedPeerService.getPeerID(for:)` is a second, separate problem**, and I should correct the first revision of this description, which called this purely a display concern. It matches `displayName` or `nickname`, first match wins, with no trust ranking and no ambiguity error, and it backs `/msg`, `/hug`, `/slap`, `/block`, `/unblock`, `/ping`, `/trace` and `/fav` — so a rename can mis-target a command, and `/fav` shares your Nostr key with whoever it resolves to. @RodsF1 has written that up as #1709; it is out of scope here.

**One surface is deliberately only half covered.** In `FingerprintView` the status *icon* flows through `getEncryptionStatus` and demotes like everywhere else, but the text badge at line 227 reads `fingerprintState.isVerified` directly and still says "verified", in green, a few lines below `peerNickname`. Suppressing that text is the wrong fix: the key *is* verified, the sheet is where that distinction should be explained rather than collapsed, and the same flag drives the verify/unverify toggle. The right fix is copy, and a new key has to ship in all 30 locales to pass `LocalizationCoverageTests` — I am not going to machine-translate a security warning into 30 languages in someone else's product.

@RodsF1 has offered the wording and the locales, and the plumbing is already in: `FingerprintPresentationState` carries `verifiedAsNickname` and `nameChangedSinceVerification`, so the two keys wire up with no further changes. The keys are deliberately **not** referenced from code yet — `everyCodeReferencedKeyExistsInACatalog` would fail until the locale patch lands, and pushing a red CI at the maintainer is worse than waiting. The vouched panel just below is gated rather than explained, for the same reason.

Suppressing the seal is the security fix. **Explaining** it to the user is a follow-up — and I originally wrote that it should hang off the notice plumbing in #1661, which I now think is wrong: that path cannot fire in production today (`migrateNoiseKeyUpdate`'s only non-test caller is gated on `userInfo["isKeyUpdate"]`/`["oldPeerPublicKey"]` and nothing posts either key). A follow-up hung on it would inherit the same deadness, so the notice wants wiring to what the transport layer already detects.

## The known limit

Pinned by a test so it is explicit rather than a surprise: if we never saw the key under its real name, its first announce **is** the baseline, so an impersonating first announce is not caught. Receiver-side binding cannot close that. Signing the nickname into the attestation can — that is a wire change and your call, so this PR does not attempt it.

## Tests

- `SecureIdentityStateManagerNicknameBindingTests`, 23 cases: the rename attack, the second-voucher laundering attempt, the vouch-before-first-announce path and its residual gap, the no-pin-before-trust rule, the rename-back on an archived row, a regression guard asserting `PeerDisplayNameResolver` really does decorate and that the binding ignores it, petname and NFC handling, recasing not breaking the binding while a look-alike does, the `@`-in-nickname and fullwidth-hex suffix cases, rebinding on re-verification, clearing on unverify only when no vouch remains, both fail-open cases, and persistence compatibility each way.
- `PeerDisplayNameResolverTests`, 7 cases, a new file — the resolver had none. Includes both known-gap rows as failing-when-fixed markers, a check that the *displayed* name is never folded, and one that the second collision counter folds the same way.
- `ChatPeerIdentityCoordinatorContextTests`, 2 added: the demotion, that it is *not* written into the cache, and that secured/handshaking states are left alone. The first of these caught the demotion being skipped on the cache-hit path.
- `TestIdentityManager`, `MockIdentityManager` and `MockChatPeerIdentityContext` all conform to the changed protocols and are updated. The mock's mismatch check folds exactly like the real one — a double that compared raw strings would call a recase a rename, and which fold applies is the whole subject of this change.

## How this was verified — please read

**@RodsF1 built and ran this on a simulator host at `a74fa10`: macOS and iOS both build, 2038 tests in 220 suites green in 67s**, including the 34 app-bundle suites I cannot execute. That is the only real-toolchain evidence this PR has. Three commits have landed since, so it wants repeating.

**CI has still never run on this PR.** All commits sit at `action_required` behind the first-time-contributor gate — @jackjackbits, the workflows need approving before any of this counts.

What I can run, on a machine with no Xcode, by supplying swift-testing as a package and stripping `#Preview` blocks in a throwaway copy:

- **201 suites / 1975 tests pass, and every difference from unmodified `main` is an addition.** I run each suite in its own process on `main` and on this branch and diff the results: `ChatPeerIdentityCoordinatorContextTests` 8 → 10, plus the two new suites at 7 and 23. Baseline 1943 → 1975.
- **The 17 XCTest-based suites now run too, which they did not in earlier revisions.** XCTest ships with Xcode, so they were previously skipped outright — including `SecureIdentityStateManagerTests`, which is the suite guarding the file this PR adds most to. A small XCTest surface implemented over swift-testing, plus a generated `@Test` per method, executes 210 of them. That harness is **not** in this branch and is evidence of a lower grade than a real XCTest run; I am citing it as "the assertions execute and pass", nothing stronger.
- **Mutation check:** with `liveNameMismatchLocked` stubbed to `return false`, 11 of the binding cases fail; with the `.noiseVerified` demotion stubbed to the identity, the coordinator's demotion case fails. The tests bite, and the ones that still pass under the first stub are the fail-open and persistence cases that should.
- Two suites are not clean in that harness and neither is a result about this change: `NostrRelayManagerTests` has one test that times out under the shim on `main` and on this branch alike, and `VerificationServiceTests` flaked once across four full sweeps and passes in isolation.

A CI run on the simulator remains the gap that matters.

### DIFF
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -164,10 +164,12 @@ final class ConversationUIModel: ObservableObject {
               !isSentByCurrentUser(message),
               let peerID = message.senderPeerID else { return false }
         guard let fingerprint = chatViewModel.getFingerprint(for: peerID) else { return false }
-        guard chatViewModel.peerIdentityStore.isVerified(fingerprint) else { return false }
-        // A verified key that renamed itself onto a name the user trusts must
-        // not carry the seal beside the new one.
-        return !chatViewModel.trustedNicknameMismatch(fingerprint)
+        // The row renders `message.sender`, frozen at receipt, so the question
+        // is whether THAT name is the one this key was verified under — not
+        // whether its current name is. Checking the current name would let a
+        // rename away, a post, and a rename back put a seal beside a name the
+        // key was never trusted under. One call, one lock.
+        return chatViewModel.sealAppliesToRow(fingerprint, renderedSender: message.sender)
     }
 
     func senderDisplayName(for peerID: PeerID, fallbackMessages: [BitchatMessage]) -> String? {
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -164,7 +164,11 @@ final class ConversationUIModel: ObservableObject {
               !isSentByCurrentUser(message),
               let peerID = message.senderPeerID else { return false }
         guard let fingerprint = chatViewModel.getFingerprint(for: peerID) else { return false }
-        return chatViewModel.peerIdentityStore.isVerified(fingerprint)
+        guard chatViewModel.peerIdentityStore.isVerified(fingerprint) else { return false }
+        // The seal is read next to `message.sender`, so it has to be bound to
+        // that name: a verified key that renamed itself onto a name the user
+        // trusts must not carry the seal beside the new one.
+        return !chatViewModel.trustedNicknameMismatch(fingerprint, displayedSender: message.sender)
     }
 
     func senderDisplayName(for peerID: PeerID, fallbackMessages: [BitchatMessage]) -> String? {
--- a/bitchat/App/ConversationUIModel.swift
+++ b/bitchat/App/ConversationUIModel.swift
@@ -165,10 +165,9 @@ final class ConversationUIModel: ObservableObject {
               let peerID = message.senderPeerID else { return false }
         guard let fingerprint = chatViewModel.getFingerprint(for: peerID) else { return false }
         guard chatViewModel.peerIdentityStore.isVerified(fingerprint) else { return false }
-        // The seal is read next to `message.sender`, so it has to be bound to
-        // that name: a verified key that renamed itself onto a name the user
-        // trusts must not carry the seal beside the new one.
-        return !chatViewModel.trustedNicknameMismatch(fingerprint, displayedSender: message.sender)
+        // A verified key that renamed itself onto a name the user trusts must
+        // not carry the seal beside the new one.
+        return !chatViewModel.trustedNicknameMismatch(fingerprint)
     }
 
     func senderDisplayName(for peerID: PeerID, fallbackMessages: [BitchatMessage]) -> String? {
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -232,9 +232,18 @@ final class PeerListModel: ObservableObject {
             let isMe = peer.peerID == myPeerID
             let fingerprint = isMe ? nil : chatViewModel.getFingerprint(for: peer.peerID)
             let isVerifiedFingerprint = fingerprint.map { peerIdentityStore.isVerified($0) } ?? false
-            let verifiedBadge = !peer.isConnected && isVerifiedFingerprint
+            // A seal is bound to the name it was earned under. Without that,
+            // a key that earned one vouch as "ravi" can rename itself to
+            // "medic" and keep rendering the seal beside the new name. A local
+            // petname already outranks the claimed nickname for display, so
+            // there is nothing to spoof and the seal stands.
+            let hasPetname = !(peer.localPetname ?? "").isEmpty
+            let renamedSinceTrust = !hasPetname && (fingerprint.map {
+                chatViewModel.trustedNicknameMismatch($0, claimedNickname: peer.nickname)
+            } ?? false)
+            let verifiedBadge = !peer.isConnected && isVerifiedFingerprint && !renamedSinceTrust
             // Vouched is subordinate to verified: never show both seals.
-            let vouchedBadge = !isVerifiedFingerprint
+            let vouchedBadge = !isVerifiedFingerprint && !renamedSinceTrust
                 && (fingerprint.map { chatViewModel.isVouchedFingerprint($0) } ?? false)
 
             return MeshPeerRow(
--- a/bitchat/App/PeerListModel.swift
+++ b/bitchat/App/PeerListModel.swift
@@ -234,13 +234,15 @@ final class PeerListModel: ObservableObject {
             let isVerifiedFingerprint = fingerprint.map { peerIdentityStore.isVerified($0) } ?? false
             // A seal is bound to the name it was earned under. Without that,
             // a key that earned one vouch as "ravi" can rename itself to
-            // "medic" and keep rendering the seal beside the new name. A local
-            // petname already outranks the claimed nickname for display, so
-            // there is nothing to spoof and the seal stands.
-            let hasPetname = !(peer.localPetname ?? "").isEmpty
-            let renamedSinceTrust = !hasPetname && (fingerprint.map {
-                chatViewModel.trustedNicknameMismatch($0, claimedNickname: peer.nickname)
-            } ?? false)
+            // "medic" and keep rendering the seal beside the new name. The
+            // check compares announced names, NOT `peer.nickname`, which is
+            // already collision-resolved: two connected peers both claiming
+            // "medic" are rendered "medic#a1b2" and "medic#c3d4", so comparing
+            // the displayed string would drop the real medic's seal during the
+            // very attack this defends against. Petnames are handled there too.
+            let renamedSinceTrust = fingerprint.map {
+                chatViewModel.trustedNicknameMismatch($0)
+            } ?? false
             let verifiedBadge = !peer.isConnected && isVerifiedFingerprint && !renamedSinceTrust
             // Vouched is subordinate to verified: never show both seals.
             let vouchedBadge = !isVerifiedFingerprint && !renamedSinceTrust
--- a/bitchat/App/VerificationModel.swift
+++ b/bitchat/App/VerificationModel.swift
@@ -16,6 +16,15 @@ struct FingerprintPresentationState: Equatable {
     let voucherCount: Int
     /// Display names of the (verified) vouchers, where known.
     let voucherNames: [String]
+    /// The nickname this key was announcing when it was verified or first
+    /// vouched, if anything was bound. Non-nil and different from the name on
+    /// screen is what `nameChangedSinceVerification` reports.
+    let verifiedAsNickname: String?
+    /// The key is trusted, but it now announces a different name than when
+    /// that trust was established. The sheet is the one place this should be
+    /// explained rather than collapsed into a glyph, so it is surfaced here
+    /// instead of silently suppressing the badge.
+    let nameChangedSinceVerification: Bool
 
     /// Vouched for by ≥1 peer the user verified (and not explicitly verified).
     var isVouched: Bool { voucherCount > 0 }
@@ -136,6 +145,13 @@ final class VerificationModel: ObservableObject {
         let isVerified = theirFingerprint.map { peerIdentityStore.isVerified($0) } ?? false
         let localPetname = theirFingerprint
             .flatMap { chatViewModel.identityManager.getSocialIdentity(for: $0)?.localPetname }
+        // A vouched seal is bound to the name it was earned under, exactly like
+        // the verified one. Missing baseline means nothing was bound, so this
+        // fails open like every other site.
+        let nameBound = theirFingerprint
+            .map { !chatViewModel.identityManager.trustedNicknameMismatch(fingerprint: $0) } ?? true
+        let verifiedAsNickname = theirFingerprint
+            .flatMap { chatViewModel.identityManager.trustedNickname(fingerprint: $0) }
 
         // Vouch state is recomputed on read: only vouchers still in the
         // verified set count, so removing a verification silently retires the
@@ -161,8 +177,12 @@ final class VerificationModel: ObservableObject {
             myFingerprint: chatViewModel.getMyFingerprint(),
             isVerified: isVerified,
             localPetname: localPetname,
-            voucherCount: vouchers.count,
-            voucherNames: voucherNames
+            // Suppressed together: a voucher list beside a name nobody
+            // vouched for is the same claim as the seal itself.
+            voucherCount: nameBound ? vouchers.count : 0,
+            voucherNames: nameBound ? voucherNames : [],
+            verifiedAsNickname: verifiedAsNickname,
+            nameChangedSinceVerification: !nameBound
         )
     }
 
--- a/bitchat/Identity/IdentityModels.swift
+++ b/bitchat/Identity/IdentityModels.swift
@@ -202,6 +202,15 @@ struct IdentityCache: Codable {
     // entries verified before this field exists sort as oldest)
     var verifiedAt: [String: Date]? = nil
 
+    /// The self-claimed nickname each fingerprint was presenting at the moment
+    /// trust in it was established here — set on verification and on a first
+    /// accepted vouch. A trust badge means "this key was the Medic I checked",
+    /// so it has to be pinned to the name it was earned under; without this
+    /// baseline a vouched key can rename itself onto a name the user trusts and
+    /// keep rendering the seal beside it. Absent for peers trusted by older
+    /// builds, which is why a missing entry never suppresses a badge.
+    var trustedNicknames: [String: String]? = nil
+
     // Stable Noise fingerprints that proved encrypted private-media support
     // inside an authenticated Noise session. Optional for decoding caches
     // written before this migration. Entries are monotonic until a panic wipe
@@ -240,6 +249,7 @@ struct IdentityCache: Codable {
         vouchesByVouchee = try container.decodeIfPresent([String: [VouchRecord]].self, forKey: .vouchesByVouchee)
         vouchBatchSentAt = try container.decodeIfPresent([String: Date].self, forKey: .vouchBatchSentAt)
         verifiedAt = try container.decodeIfPresent([String: Date].self, forKey: .verifiedAt)
+        trustedNicknames = try container.decodeIfPresent([String: String].self, forKey: .trustedNicknames)
         privateMediaCapableFingerprints = try container.decodeIfPresent(Set<String>.self, forKey: .privateMediaCapableFingerprints)
         authenticatedSigningKeysByFingerprint = try container.decodeIfPresent([String: Data].self, forKey: .authenticatedSigningKeysByFingerprint)
         cryptographicIdentities = try container.decodeIfPresent([String: CryptographicIdentity].self, forKey: .cryptographicIdentities) ?? [:]
--- a/bitchat/Identity/SecureIdentityStateManager.swift
+++ b/bitchat/Identity/SecureIdentityStateManager.swift
@@ -130,6 +130,9 @@ protocol SecureIdentityStateManagerProtocol {
     func setVerified(fingerprint: String, verified: Bool)
     func isVerified(fingerprint: String) -> Bool
     func getVerifiedFingerprints() -> Set<String>
+    /// Whether this peer is now claiming a different nickname than the one its
+    /// trust was earned under.
+    func trustedNicknameMismatch(fingerprint: String, claimedNickname: String) -> Bool
 
     // MARK: Vouching (transitive verification)
     @discardableResult
@@ -148,6 +151,17 @@ protocol SecureIdentityStateManagerProtocol {
     // MARK: Private-media downgrade protection
     func markPrivateMediaCapable(fingerprint: String)
     func hasObservedPrivateMediaCapability(fingerprint: String) -> Bool
+}
+
+extension SecureIdentityStateManagerProtocol {
+    /// `trustedNicknameMismatch` for a name as *rendered* in a message row,
+    /// which may carry a `#abcd` disambiguation suffix that an announced
+    /// nickname never has. Comparing the displayed string verbatim would read
+    /// every suffixed sender as a mismatch.
+    func trustedNicknameMismatch(fingerprint: String, displayedSender: String) -> Bool {
+        trustedNicknameMismatch(fingerprint: fingerprint,
+                                claimedNickname: displayedSender.splitSuffix().0)
+    }
 }
 
 /// Singleton manager for secure identity state persistence and retrieval.
@@ -428,6 +442,15 @@ final class SecureIdentityStateManager: SecureIdentityStateManagerProtocol {
                 } else if self.cache.socialIdentities[fingerprint] == nil {
                     self.cache.socialIdentities[fingerprint] = identity
                 }
+                // A vouch usually arrives for a peer we have not seen announce
+                // — it comes over Noise from someone else, and the vouchee may
+                // be several hops away. There was no name to bind to then, so
+                // bind on the first announce we see while the trust stands.
+                self.pinTrustedNicknameLocked(
+                    fingerprint: fingerprint,
+                    overwrite: false,
+                    requireExistingTrust: true
+                )
             }
 
             self.saveIdentityCache()
@@ -678,9 +701,15 @@ final class SecureIdentityStateManager: SecureIdentityStateManagerProtocol {
                 var verifiedAt = self.cache.verifiedAt ?? [:]
                 verifiedAt[fingerprint] = Date()
                 self.cache.verifiedAt = verifiedAt
+                // Re-verifying overwrites the baseline: the user just checked
+                // this key again, under whatever name it presents now.
+                self.pinTrustedNicknameLocked(fingerprint: fingerprint, overwrite: true)
             } else {
                 self.cache.verifiedFingerprints.remove(fingerprint)
                 self.cache.verifiedAt?.removeValue(forKey: fingerprint)
+                if self.cache.vouchesByVouchee?[fingerprint]?.isEmpty ?? true {
+                    self.cache.trustedNicknames?.removeValue(forKey: fingerprint)
+                }
             }
 
             // Update trust level if social identity exists
@@ -696,6 +725,52 @@ final class SecureIdentityStateManager: SecureIdentityStateManagerProtocol {
     func isVerified(fingerprint: String) -> Bool {
         queue.sync {
             return cache.verifiedFingerprints.contains(fingerprint)
+        }
+    }
+
+    // MARK: - Nickname binding
+    //
+    // A vouch signs `voucheeFingerprint | voucheeSigningKey | timestampMs` and
+    // deliberately says nothing about a name — the attestation is right to stay
+    // name-free. But the badge is *rendered* next to a self-claimed nickname,
+    // so the binding has to exist somewhere, and the receiver is the only party
+    // that can hold it: it is the one that decided to trust this key while it
+    // was presenting a particular name.
+
+    /// Requires `queue`. Records the peer's currently claimed nickname as the
+    /// name its trust is bound to. Empty nicknames are not pinned — a key
+    /// verified before its first announce has no name to bind to, and pinning
+    /// "" would then read as a mismatch against every later announce.
+    ///
+    /// `requireExistingTrust` is for the announce path, which runs for every
+    /// peer: pinning a name before there is any trust would bind the badge to
+    /// whatever name we happened to see first, so a peer who renamed *before*
+    /// being vouched would have its legitimate badge suppressed.
+    private func pinTrustedNicknameLocked(fingerprint: String,
+                                          overwrite: Bool,
+                                          requireExistingTrust: Bool = false) {
+        if requireExistingTrust {
+            let trusted = cache.verifiedFingerprints.contains(fingerprint)
+                || !(cache.vouchesByVouchee?[fingerprint] ?? []).isEmpty
+            guard trusted else { return }
+        }
+        guard let claimed = cache.socialIdentities[fingerprint]?.claimedNickname,
+              !claimed.isEmpty else { return }
+        var pinned = cache.trustedNicknames ?? [:]
+        if !overwrite, pinned[fingerprint] != nil { return }
+        pinned[fingerprint] = claimed
+        cache.trustedNicknames = pinned
+    }
+
+    /// True only when a baseline exists AND the peer now claims something else.
+    /// Fails OPEN on a missing baseline on purpose: peers trusted by builds
+    /// before this existed have none, and dropping their badges on upgrade
+    /// would train users to ignore the signal.
+    func trustedNicknameMismatch(fingerprint: String, claimedNickname: String) -> Bool {
+        queue.sync {
+            guard let pinned = cache.trustedNicknames?[fingerprint], !pinned.isEmpty,
+                  !claimedNickname.isEmpty else { return false }
+            return pinned != claimedNickname
         }
     }
     
@@ -760,6 +835,10 @@ final class SecureIdentityStateManager: SecureIdentityStateManagerProtocol {
             var vouches = self.cache.vouchesByVouchee ?? [:]
             vouches[voucheeFingerprint] = capped
             self.cache.vouchesByVouchee = vouches
+            // Pin on the FIRST vouch only. A later vouch for the same key must
+            // not move the baseline, or an attacker could rename and then have
+            // a second voucher silently re-anchor the badge to the new name.
+            self.pinTrustedNicknameLocked(fingerprint: voucheeFingerprint, overwrite: false)
             self.saveIdentityCache()
             return true
         }
--- a/bitchat/Identity/SecureIdentityStateManager.swift
+++ b/bitchat/Identity/SecureIdentityStateManager.swift
@@ -179,7 +179,7 @@ extension SecureIdentityStateManagerProtocol {
         else { return true }
         let shown = renderedSender.withoutCollisionSuffix
         guard !shown.isEmpty else { return true }
-        return pinned.normalizedNickname == shown.normalizedNickname
+        return pinned.nicknameBindingKey == shown.nicknameBindingKey
     }
 }
 
@@ -826,7 +826,7 @@ final class SecureIdentityStateManager: SecureIdentityStateManagerProtocol {
             // stand between the reader and a spoofed name on this row.
             let shown = renderedSender.withoutCollisionSuffix
             guard !shown.isEmpty else { return true }
-            return pinned.normalizedNickname == shown.normalizedNickname
+            return pinned.nicknameBindingKey == shown.nicknameBindingKey
         }
     }
 
@@ -839,9 +839,9 @@ final class SecureIdentityStateManager: SecureIdentityStateManagerProtocol {
         // is displayed, so there is nothing to spoof and the seal stands.
         if let petname = social?.localPetname, !petname.isEmpty { return false }
         guard let claimed = social?.claimedNickname, !claimed.isEmpty else { return false }
-        // NFC, matching `normalizedNickname` everywhere else nicknames are
-        // compared: a decomposed and a precomposed "café" are one name.
-        return pinned.normalizedNickname != claimed.normalizedNickname
+        // See `nicknameBindingKey`: NFC plus case, so a decomposed and a
+        // precomposed "café" are one name and recasing is not a rename.
+        return pinned.nicknameBindingKey != claimed.nicknameBindingKey
     }
     
     func getVerifiedFingerprints() -> Set<String> {
--- a/bitchat/Identity/SecureIdentityStateManager.swift
+++ b/bitchat/Identity/SecureIdentityStateManager.swift
@@ -130,9 +130,19 @@ protocol SecureIdentityStateManagerProtocol {
     func setVerified(fingerprint: String, verified: Bool)
     func isVerified(fingerprint: String) -> Bool
     func getVerifiedFingerprints() -> Set<String>
+    /// The nickname `fingerprint` was announcing when trust in it was
+    /// established, or nil if nothing was bound. Shown to the user, so a sheet
+    /// can say *which* name was verified rather than only that one changed.
+    func trustedNickname(fingerprint: String) -> String?
     /// Whether this peer now announces a different nickname than the one its
     /// trust was earned under.
     func trustedNicknameMismatch(fingerprint: String) -> Bool
+    /// Verified AND still presenting the name it was verified under — what a
+    /// seal beside a LIVE name is actually asserting. One lock, not two.
+    func isVerifiedAndNameBound(fingerprint: String) -> Bool
+    /// Verified AND `renderedSender` is the name it was verified under — what
+    /// a seal beside a name frozen on a message row is asserting.
+    func sealAppliesToRow(fingerprint: String, renderedSender: String) -> Bool
 
     // MARK: Vouching (transitive verification)
     @discardableResult
@@ -151,6 +161,26 @@ protocol SecureIdentityStateManagerProtocol {
     // MARK: Private-media downgrade protection
     func markPrivateMediaCapable(fingerprint: String)
     func hasObservedPrivateMediaCapability(fingerprint: String) -> Bool
+}
+
+extension SecureIdentityStateManagerProtocol {
+    // Default implementations so a conformance only has to provide the two
+    // primitives. `SecureIdentityStateManager` overrides both to answer in a
+    // single lock acquisition — these seal checks run per row, per render, and
+    // the format cache does not shield the one in ChatMessageFormatter because
+    // the answer is part of its cache key.
+    func isVerifiedAndNameBound(fingerprint: String) -> Bool {
+        isVerified(fingerprint: fingerprint) && !trustedNicknameMismatch(fingerprint: fingerprint)
+    }
+
+    func sealAppliesToRow(fingerprint: String, renderedSender: String) -> Bool {
+        guard isVerified(fingerprint: fingerprint) else { return false }
+        guard let pinned = trustedNickname(fingerprint: fingerprint), !pinned.isEmpty
+        else { return true }
+        let shown = renderedSender.withoutCollisionSuffix
+        guard !shown.isEmpty else { return true }
+        return pinned.normalizedNickname == shown.normalizedNickname
+    }
 }
 
 /// Singleton manager for secure identity state persistence and retrieval.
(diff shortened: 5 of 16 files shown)


RodsF1:
Reviewed this against `main` because it is the mirror of a finding I raised on #1661. The vulnerability is real and worth fixing — I confirmed the rename is free and silent (same key, new nickname is accepted; the only trace is a `SecureLogger.debug` line in `BLEAnnounceHandler`), that `verifyFingerprint` never records the name it verified, and that `displayName` falls back to the peer-claimed nickname. Four findings, roughly in order of how much they matter.

**1. The patch does not cover the case the attack actually needs.** `verifiedBadge` already carried `!peer.isConnected` before this PR, so `&& !renamedSinceTrust` is dead for a connected peer. And `EncryptionStatus.noiseVerified.icon` is *the same glyph*, `checkmark.seal.fill` — rendered for connected peers straight from `peer.encryptionStatus.icon` in `MeshPeerList`, and again in the DM header in `ContentSheetViews`. Neither is gated. So a verified key that renames while it holds a Noise session still shows a filled seal beside the new name, which is the live impersonation, not an edge case. This is the Codex P1 above, and I think it understates it: the patch currently only moves the *offline* badge.

**2. It can suppress the seal of the peer it is protecting, during the attack.** `PeerListModel` compares the pinned baseline against `peer.nickname` — but that field is already collision-resolved. `BLEPeerRegistry.transportSnapshots` runs names through `PeerDisplayNameResolver`, which appends `#` + the first four of the peerID whenever two *connected* peers share a nickname. So in your own scenario, with Eve and the real medic both connected, the real medic's `peer.nickname` becomes `medic#c3d4`, which no longer equals the pinned `medic` — and the real medic loses their seal. You wrote the `displayedSender` overload that strips exactly this suffix, but the peer-list site calls the raw `claimedNickname:` one. Related: `splitSuffix()` also strips **every** `@` anywhere in the string, and `@` is legal in a nickname, so a peer named `ravi@hq` mismatches its own baseline at the message sites.

**3. Your point (e) is not quite right, in a way that helps you.** There *is* duplicate-nickname disambiguation — the resolver above. But it counts collisions only among connected peers and suffixes only connected peers, so it does nothing when the real medic is offline or out of range, and never applies to offline favourite rows. That is a better argument for this PR than "no disambiguation exists": the mitigation exists and has a hole exactly where you are aiming.

**4. I would not call it display-layer only.** `UnifiedPeerService.getPeerID(for:)` matches `displayName` or `nickname`, first match wins, no trust ranking and no ambiguity error, and it backs `/msg`, `/hug`, `/slap`, `/block`, `/unblock`, `/ping`, `/trace` and `/fav`. So a rename can mis-target a command, and `/fav` shares your Nostr key with whoever it resolves to. Out of scope for this PR, but worth saying out loud rather than leaving it implied by "the seal is wrong".

**One thing before you build the follow-up.** You wrote that explaining this to the user "wants the notice plumbing #1661 is adding". That plumbing cannot fire in production today — `migrateNoiseKeyUpdate`'s only non-test caller is gated on `userInfo["isKeyUpdate"]` and `["oldPeerPublicKey"]`, and nothing in the tree posts either key; the writer was removed in #693, and the tests there pass because they call the method directly. Details in [my review on #1661](https://github.com/permissionlesstech/bitchat/pull/1661#pullrequestreview-5062434349). So a follow-up hung on that notice would inherit the same deadness.

Worth noting your option 1 there — surfacing the existing `previousNickname` detection instead of only logging it — is the same shape as the route I recommended to @jackjackbits: wire the notice to what the transport layer already detects. Two of us arriving at that independently is probably worth him knowing.

Good find, and the receiver-side TOFU framing is the right instinct — the binding genuinely has to live where the name was seen. Happy to re-check once the connected paths are covered.

shubhambhandari126:
All four confirmed against `main`, and the first two were bugs in the patch rather than gaps in it. Pushed in a74fa10.

**1 — the connected peer.** You are right that the patch was moving the wrong badge, and I do not think you were overstating it. `verifiedBadge` already carried `!peer.isConnected`, so `&& !renamedSinceTrust` was dead for exactly the peer that matters, and `.noiseVerified.icon` is the same `checkmark.seal.fill` drawn straight from `peer.encryptionStatus.icon`. The gate now sits in `ChatPeerIdentityCoordinator.getEncryptionStatus`, which both the connected peer row and the `ContentSheetViews` header read through, demoting `.noiseVerified` to `.noiseSecured`. The session is still encrypted; only the claim about who is withdrawn. I checked every `noiseVerified` reference first — `VerificationModel:24`, `ContentSheetViews:659` and `:760` all already treat it and `.noiseSecured` identically, so nothing behavioural moves.

Two things fell out of that. The demotion has to be computed on read rather than stored, because the cached status is invalidated by verification and session changes but never by a rename. And it has to run on the cache-hit path too — I missed that at first, and the new coordinator test caught it, which is a fair argument for the test existing.

**2 — the collision suffix.** This is the one I am glad you caught, because it inverted the patch: in the two-connected-namesakes case the real medic renders as `medic#c3d4`, mismatches the pinned `medic`, and loses their seal. The patch would have punished the victim in precisely the scenario I wrote it for.

I did not take the `displayedSender` fix, because it would have left me comparing decorated strings and reasoning about which decorations apply where. The check now takes no name at all — pinned baseline against the peer's own announced nickname. That kills the suffix problem, your `@` problem (`splitSuffix` strips every `@`, and nothing in `validateNickname` forbids one, so `ravi@hq` mismatched itself), and a third case I had not thought about, where a message row's historical sender name diverges from the live one. Both sides go through `normalizedNickname`, matching `getPeerID(for:)` and the rest of the nickname comparisons; a decomposed and a precomposed `café` should not read as a rename. There is a test asserting `PeerDisplayNameResolver` really does decorate and that the binding ignores it, so nobody reintroduces the rendered-name comparison later.

**3 — the mitigation exists and has a hole.** Taken, and it is a better framing than mine; the PR description now says it that way. Connected-only collision counting, connected-only suffixing, nothing for offline favourite rows.

**4 — not display-layer only.** Confirmed: `getPeerID(for:)` returns the first peer whose `displayName` or `nickname` matches, no trust ranking, no ambiguity error. I have corrected the PR description rather than leave the narrower claim standing. `/fav` sharing a Nostr key with whatever the name resolves to is the part I would not have wanted to leave implied. Out of scope here, but it seems worth its own issue — happy to open one, or leave it to you since you found it.

**On the #1661 plumbing.** You are right and I should not have leaned on it; I had read your review but still wrote the follow-up as hanging off that notice. A follow-up built on a path whose only writer was removed in #693 inherits the deadness. The PR description now says so.

@jackjackbits — flagging the convergence, since it seems worth knowing that two people got there independently: RodsF1's route 2 on #1661 and my option 1 here are the same recommendation, which is to wire the notice to what the transport layer already detects rather than to the removed favourites-layer trigger. `BLEAnnounceHandler` and `BLEPeerRegistry` already carry `previousNickname` and it currently only reaches a debug log.

One caveat that has not changed: no Xcode here, so `just test-ios` has not run and the 34 app-bundle-dependent suites (`ChatViewModel*`, `ViewSmokeTests`, `AppArchitectureTests`) never execute. That is why both halves of this live in `SecureIdentityStateManager` and `ChatPeerIdentityCoordinator`, which tests can reach. 186 suites / 1897 tests pass with no result differing from `main`; the only changes are additions. Re-check welcome.

shubhambhandari126:
One more surface, found while re-checking the connected paths end to end rather than assuming my own fix was complete.

Everything that draws the seal for a connected peer now routes through `getEncryptionStatus` and demotes: the peer row (via `peer.encryptionStatus.icon`), the vouched seal (which renders outside the connected/offline branch, so it needed the gate too), the `ContentSheetViews` DM header, and — I had not traced this one before — the `FingerprintView` status icon and the green tint on the nickname, both of which come from `VerificationModel:133` calling the same accessor.

The exception is `FingerprintView`'s text badge at line 227. It reads `fingerprintState.isVerified` directly, so after a rename the sheet shows a lock and the new nickname but still says "verified" in green. Suppressing it looks wrong to me — the key genuinely is verified, that sheet is the one place the distinction can be explained instead of collapsed into a glyph, and the same flag drives the verify/unverify toggle. It wants copy, and any new key needs all 30 locales to pass `LocalizationCoverageTests`, so I have left it and written it into the description rather than machine-translating a security string. Happy to add it if you want to supply the wording.

`isVerifiedNickname` is untouched on purpose: it means the announce was signature-authenticated, not that the user verified anyone, and a rename does not make the new name less authentically announced.

shubhambhandari126:
Adjacent finding, from taking @RodsF1's point 3 further — the disambiguation mitigation has a third blind spot, and I want to flag it here rather than quietly widen this PR.

`PeerDisplayNameResolver` counts collisions with `counts[p.nickname]`, a Swift `String` key, so it disambiguates more than it looks like it does: `String` hashing is Unicode canonical equivalence, so `José` precomposed and `Jose` + U+0301 already collide and both get suffixed, for free. What it does not fold is case, or anything from another script.

Running the resolver as it stands, with a verified peer announcing `Medic` and a second **connected** peer announcing:

| second peer | resolved names | `#abcd` fires |
| --- | --- | --- |
| `Medic` | `Medic#aaaa` / `Medic#bbbb` | yes |
| `José` vs `Jose`+U+0301 | `José#aaaa` / `José#bbbb` | yes |
| `medic` | `Medic` / `medic` | **no** |
| `Меdic` (Cyrillic М + е) | `Medic` / `Меdic` | **no** |
| `Ｍedic` (fullwidth) | `Medic` / `Ｍedic` | **no** |
| `𝐌edic` (math bold) | `Medic` / `𝐌edic` | **no** |

So two connected rows can render identically while the suffix that exists to tell namesakes apart stays silent. To be accurate about severity: this is **not** the impostor acquiring a seal — seals are keyed by fingerprint, so the look-alike gets none. It is that the reader has to notice which of two identical-looking rows carries the seal, having been given no `#abcd` to separate them. Lower severity than the case this PR fixes, and a different mechanism: cross-peer name collision, not per-key rename.

**I have not added this to this branch, deliberately.** The fix is a UTS #39 confusable skeleton, which means a generated Unicode table — mine is 6,355 mappings, about 60KB — living in `PeerDisplayNameResolver`, a file this PR does not touch. Putting that behind a review that is about trust seals would mean shipping a table into the app binary as a side effect of a different discussion, and whether BitChat wants that is a product call rather than a review comment. Happy to open it as its own PR if you want it, or to leave it as a note.

Two things I learned doing it elsewhere, in case it saves someone time:

**Case folding and confusable folding conflict, and neither order works.** Uppercase Cyrillic `М` is confusable with Latin `M`, but lowercase `м` is not confusable with `m` — it skeletons to `ʍ` — so folding case first destroys the resemblance being detected, which is why UTS #39 defines skeleton on the string as displayed. Fold case first and you catch `MEDIC` but miss `Меdic`; skeleton first and you catch `Меdic` but miss `MEDIC`. Iterating to a fixed point just picks a different subset. What worked was keeping the two forms separate and taking the union.

**The table has to be generated, not written.** A hand-maintained homoglyph list drifts, and one that folds `М` onto `M` on iOS but not on another client is worse than no table, because the two then disagree about who is impersonating whom. Also worth knowing if anyone pins the source: `https://www.unicode.org/Public/security/latest/` currently serves data labelled `Version: 17.0.0` from a directory that does not exist yet — only `16.0.0` is published — so `latest` is not reproducible.

And a caution in the other direction: UTS #39 maps `m` onto `rn`, so `medic` and `rnedic` collide. The table will occasionally flag an honest handle, which is fine when the outcome is a suffix and would need more thought if the outcome were suppressing something.

This PR is unchanged and still stands on its own.

RodsF1:
Answers to your three questions, and four things from re-checking `a74fa10`. The revision is a clear improvement — the demotion is confined to display (I enumerated all three `getEncryptionStatus` readers and all ten `.noiseVerified` references), and the announced-to-announced comparison genuinely kills all three false positives.

**First, the thing nobody has said: this PR has never had CI run.** Both commits show `action_required` — the first-time-contributor gate. So your test numbers are the only evidence anyone has, from a machine without Xcode, for a change that lives in the iOS render paths. @jackjackbits needs to approve the workflows before any of it counts.

Meanwhile I built and ran it for you. **macOS and iOS simulator both build, and the full suite is green: 2038 tests, 220 suites, 67s.** That includes the 34 app-bundle suites you could not execute — they run fine on a simulator host, so that gap is closable, not structural. Your `SecureIdentityStateManagerNicknameBindingTests` (15) and `ChatPeerIdentityCoordinatorContextTests` (10) pass individually too.

---

**1. The `/fav` routing problem — I have taken it: #1709.** I had the call sites traced, so it was cheaper for me to write up. It covers `getPeerID(for:)` returning first-match with no trust ranking and no ambiguity error, the seven commands that reach it, why `/fav` is the one that costs something irreversible, and four possible shapes without assuming one. Credited to your PR since that is where it came from.

**2. The fingerprint sheet copy — yes, and I will do the 30 locales.** That is my lane on this repo and you are right not to machine-translate a security string. Matching the existing register (`✓ VERIFIED` / `⚠️ NOT VERIFIED`, and `fingerprint.message.verified` = "you have verified this person's identity."):

```
fingerprint.badge.verified_renamed   =  ⚠️ VERIFIED, NAME CHANGED
fingerprint.message.verified_renamed =  you verified this key as "%@". it now announces a different name.
                                        the key has not changed; the name it presents has.
```

Keeping VERIFIED in the badge is deliberate — the key genuinely is verified, and dropping the word collapses exactly the distinction the sheet exists to preserve. The `%@` is the pinned baseline, which means surfacing `trustedNicknames[fingerprint]` through `FingerprintPresentationState`; if you would rather avoid that plumbing, the no-placeholder form is `this key is verified, but it now announces a different name than when you verified it.` — less useful, since naming the old name is what lets someone say "oh, that is ravi". Say which you want and I will send the two keys across all 30 locales as a patch you can drop in.

**3. The confusables table — I would not open it as a PR, but there is a one-liner worth taking first.** Your analysis is right; I reproduced your six-row table exactly, and the `latest/` reproducibility footnote checks out. But before a 60KB generated table, fold the collision **count** key (not the displayed name) through compatibility normalization and case:

```swift
counts[p.nickname.precomposedStringWithCompatibilityMapping.lowercased(), default: 0] += 1
```

Measured against your table, that closes three of your four misses — case, fullwidth, math-bold — for one line and zero bytes. Only the Cyrillic row still needs a real confusables table. That seems like the right split: take the free 3, and let the table be its own conversation with @jackjackbits, since shipping a Unicode table into the binary is a product call rather than a review comment.

---

**Four findings on the revision itself.**

**(a) The rename-back restores the seal, and this one is my fault.** I pushed you off comparing the rendered name; comparing live-to-live means a row is no longer checked against its own sender. `ChatMessageFormatter` renders `message.sender`, frozen at receipt, while `isVerifiedSender` now gates on the live claimed nickname. So: verified as `ravi` → announce `medic` → post to the public timeline → announce `ravi` again. Mismatch goes false, and the archived row renders `<@medic ✓>` — a seal beside a name the key was never trusted under, which is the attack this PR exists to stop. Your v1 would have caught it. The shape that holds both ways is to require the live name to match the baseline **and** the row's own sender to match it — or to carry the undecorated announced nickname on `BitchatMessage` at receipt, so a row can be checked against its own name without reintroducing `splitSuffix`.

**(b) The rename suppresses the seal but not the trust.** Every call site of `trustedNicknameMismatch` is a render path. `recordVouch` gates only on `cache.verifiedFingerprints.contains(voucherFingerprint)`, and outgoing batches come from `mostRecentlyVerifiedFingerprints` — neither consults the mismatch. So the renamed key keeps its authority to confer a vouched seal on other keys, and this device keeps advertising it as verified under whatever name it now claims. That is the difference between the impostor looking less trustworthy and the impostor losing the ability to launder trust onward.

**(c) The vouched seal in the fingerprint sheet is ungated** — `FingerprintView.swift:192-222`, the teal hollow seal with "vouched by N" and the voucher names, reads `voucherCount` and never consults the mismatch. That is the seal this PR is named for, in the one sheet you singled out as the place the distinction should be explained.

**(d) Your audit claim is 3-for-4.** `FingerprintView.swift:87` is a fourth `.noiseVerified` reference, and it is the one site that does **not** treat the two statuses identically — `== .noiseVerified ? Color.green : textColor`. Only a tint, so harmless, but it is what produces the contradiction you flagged separately: after a rename the sheet shows a plain lock and a "secured" caption above a bold green "verified". Also worth restating precisely: the accessibility description changes too, so VoiceOver stops saying "verified" on the DM header. Correct outcome, but not "nothing moves".

Minor, while you are in there: both `showsVerifiedSeal` and `isVerifiedSender` already took a `queue.sync` via `isVerified`, and now take a second — and in the formatter the check sits *before* the cache lookup because it is part of the cache key, so the format cache does not protect it. A single `isVerifiedAndNameBound(fingerprint)` would halve it.

Nothing above changes my view that the direction is right.

shubhambhandari126:
Thank you for building and running it — that closes the one gap I could not close myself, and it is a more useful contribution to this PR than anything I wrote today. 2038 tests on a simulator host is the evidence I did not have. Noted on the `action_required` gate: @jackjackbits, the workflows on this PR have never run, so nothing here has been through CI.

Fixed in a008850: (a) and (c) were bugs, (d) was my claim being wrong, and I took your resolver one-liner.

**(a) — the rename-back.** Confirmed and fixed. Do not take the blame for this one; you pushed me off comparing a decorated string, which was right, and I then compared the wrong thing instead of the undecorated right thing.

The fix is not "require both" — that was my first instinct and it is wrong in the other direction. A live row and a historical row are asking different questions:

| surface | question |
| --- | --- |
| peer list, DM header, encryption status | is the name this peer announces **now** the one it was verified under? |
| public timeline and DM rows | is the name **on this row** the one it was verified under? |

So `sealAppliesToRow(fingerprint:renderedSender:)` checks the row's own sender against the baseline and never consults the live name. That kills the rename-back, and it keeps the converse honest: a historical `ravi ✓` stays sealed while that key is currently renamed, because that row really was ravi. Requiring both would have retroactively unsealed accurate history every time anyone changed their nickname.

Comparing a rendered name again needed the suffix stripped without `splitSuffix`'s `@` problem, so there is a `withoutCollisionSuffix` that removes only a trailing `#` plus four hex digits. Reverting the row check to the live name fails 3 of the 5 new tests, including the sequence you described.

**(c) — the vouched panel.** A straight miss, and the worst-placed one: the seal this PR is named for, in the sheet I had just argued was where the distinction should be explained. `voucherCount` and `voucherNames` are now suppressed together — a voucher list beside a name nobody vouched for is the same claim as the seal.

**(d) — my claim was wrong and I have corrected it.** `FingerprintView.swift:87` is a fourth reference and it does not treat the two alike, and `accessibilityDescription` differs per case so VoiceOver stops saying "verified" on a demoted header. Right outcome, but "nothing moves" was not accurate and I should not have written it that confidently after enumerating only the switch statements.

**(3) — took the one-liner, and it does more than you measured.** Folding the count key with `precomposedStringWithCompatibilityMapping.lowercased()` closes five categories, not three: case, fullwidth, mathematical bold, **script ℳ and roman-numeral Ⅿ**. NFKC also subsumes the canonical folding Swift's `String` keys were already giving for free, so that behaviour is now explicit rather than incidental. Only cross-script homoglyphs are left. `PeerDisplayNameResolver` had no tests at all, so it has six now, including the two known-gap rows as failing-when-fixed markers for whoever adds the table. Agreed the table is its own conversation.

**(2) — the `%@` form, please, and the plumbing is already in.** `FingerprintPresentationState` now carries `verifiedAsNickname` and `nameChangedSinceVerification`, so the two keys can be wired with no further changes. I deliberately did **not** reference the keys from code yet: `everyCodeReferencedKeyExistsInACatalog` would fail until your locale patch lands, and I would rather not push a red CI at the maintainer. Send the patch and I will wire the view in the same commit. Your register is right, and keeping VERIFIED in the badge is the part I would have got wrong — I was thinking about suppressing it.

**(1) — thank you for taking #1709.** Better written than mine would have been, and four shapes without picking one is the right way to hand it over.

---

**(b) — I disagree, and I want to be specific about where, because I think you are right about the shape and wrong about the trigger.**

You are right that this is display-only and that display-only lets an impostor keep laundering trust. Where I part company is that `trustedNicknameMismatch` fires on **any** rename, not on impersonation. Gating authority on it would mean: change your nickname, and your vouches silently stop counting on every device that verified you, with nothing shown to you or to them. Renames are ordinary. That is a large blast radius for a signal that is not evidence of anything on its own.

The trigger that would justify revoking authority is the narrow one — this key renamed onto a name another trusted key already holds — and BitChat has no cross-peer name comparison to detect that. `PeerDisplayNameResolver` counts collisions for display and nothing consults it for trust. So the honest version of your finding is not "gate `recordVouch` on the mismatch"; it is "there is no impersonation detector here, only a rename detector", and building one is the conversation that starts with the confusables table.

I should also say that I checked my own project rather than only arguing: it has the identical boundary. A key flagged as impersonating still propagates trust to keys it vouched for — flagged 🔴 itself, while a peer it vouched for stays 🟣 through it. Same call, made twice, so I am not applying a standard to BitChat that I dodged at home. You have convinced me it is worth fixing **there**, where a real impersonation detector exists to trigger it, and I will do that.

For here, my position is that the display/authority split is the correct scope for this PR and that the authority question needs the detector first. If @jackjackbits reads it the other way I will implement whichever trigger he wants — it is his product and his false-positive budget, not mine.

Everything else is unchanged: 187 suites / 1908 tests here with no result differing from `main`, only additions.

shubhambhandari126:
Audited my own revision before asking you to look again, and found three more defects in it. Fixed in ff39744. Each has a test that fails without its fix, which is the only claim I will make unprompted now.

**1. Recasing your own nickname broke the binding.** The comparison was NFC only, so `Ravi` → `ravi` read as a rename and silently dropped the seal. A rename is meant to break the binding; a recase is not a rename. There is an explicit `nicknameBindingKey` now — NFC, case fold, NFC again, because case folding can itself emit decomposed sequences.

It is deliberately NFC and **not** NFKC, and this is the part I want on the record: a fullwidth `Ｍedic` merely *looks* like `Medic`, so it is a different name and must break the binding. Folding look-alikes is `collisionKey`'s job, which answers the different question of whether two peers need telling apart. Both directions are tested, because getting it backwards is how a vouch for one name comes to cover another.

**2. `withoutCollisionSuffix` stripped more than this device generates.** `Character.isHexDigit` is true for fullwidth digits, so a nickname literally ending in `#ＡＢＣＤ` was truncated and could then match a baseline it is not. ASCII-only now, matching how `splitSuffix()` recognises the suffix. Contrived, but it is a seal on a name that was never verified, which is the whole subject of this PR.

**3. Collisions are counted in two places and I fixed one.** `BLEPeerSenderDisplayName.collisionResolvedName` compares nicknames exactly and feeds sender names on file transfers, so it kept the blind spot the peer-list resolver had just lost. It folds through the same `collisionKey` now, so the two counters cannot drift. That one is on me — your finding was about the resolver, and I should have grepped for other collision counters before reporting the fix as done.

187 suites / 1912 tests here, no result differing from `main`; the same 34 still need an app bundle.

---

**And the BitTrust half of (b), since I said I would do it.** Worth reporting because the first attempt was wrong in a way that is directly relevant to this review.

I closed the authority side there: a key is a conduit only while it presents a name somebody vouched it under, so a flagged key stops conferring degrees on everyone it vouched for. Then, auditing it, I found I had used the look-alike comparison for that check — while my own spec table says a rename check must use the strict one, and says why. So it only half worked:

| Eve, vouched as `Ravi` | Eve's badge | Mallory, vouched by Eve |
| --- | --- | --- |
| announces `Ravi` | 2nd | 3rd |
| announces `Zebra` | unknown | unknown — correct |
| announces `Rаvi` (Cyrillic а) | unknown | **3rd — wrong** |

An attacker kept laundering trust onward by picking a look-alike instead of an unrelated name: badge gone, authority intact, which is precisely the hole the change claimed to close. Fixed, and the memoisation with it — deciding conduit status per edge scanned every edge for every edge, and that runs on every evaluate.

The reason this matters here rather than being my own housekeeping: it is the same mistake in both directions. Mixing up "is this the same name" with "does this look like that name" produced a false negative there and would have produced false positives here. The two questions want two different comparisons at every single site, and I have now got that wrong once in each direction.

Your (b) framing is what sent me looking, so the finding is yours as much as mine. My position on the BitChat side is unchanged — a vouch that signs only a key stays true across a rename, so the trigger there has to be detected impersonation rather than renaming — but I would not have found the BitTrust bug without your pushing on it.

**(2) still open on your side.** `%@` please; `verifiedAsNickname` and `nameChangedSinceVerification` are on `FingerprintPresentationState` and unused, waiting for your two keys. I have not referenced them from code so CI stays green until your locale patch lands.

shubhambhandari126:
Correcting myself again, because I told you the BitTrust authority side was closed and it was not.

I audited the whole of it rather than only the bug I had already found there, and §4.3 turned out to implement half of what it claimed. It said a key failing the handle check stops carrying trust. It only cut keys that had **renamed**. An attacker vouched under a look-alike name never renames at all, so the condition never fired:

| impostor's handle | its badge | the key it vouched |
| --- | --- | --- |
| Cyrillic `Меdic` | 🔴 | 🟣 third |
| `medic` | 🔴 | 🟣 third |
| fullwidth `Ｍedic` | 🔴 | 🟣 third |

Badge suppressed, authority intact — the precise thing you described in (b), surviving two attempts to fix it. All three now leave the vouched key unknown, and the two controls (an unrelated name, and a look-alike with no trusted namesake) still confer a degree, so it is not just suppressing harder.

Why it was missed is the part worth passing on. The impersonation condition depends on the graph — a namesake only counts if it is reachable — so evaluating it while building the graph is circular, and it is tempting to skip. The shape that works is two fixed passes rather than a fixed point:

```
base    = live edges from keys that have not renamed
flagged = { k : k impersonates, judged against base }
graph   = base minus flagged, as issuers
```

Verdicts decided against `base`, paths found in `graph`, so a badge and conduit status cannot disagree. A fixed point is actively wrong here: removing a key can make a namesake unreachable and so *un*flag someone, which oscillates. Two passes terminate and the second only ever removes trust.

That is the third time in this area I have confused "is this the same name" with "does this look like that name" — twice in my code here and once there, producing false positives in one direction and false negatives in the other. The two questions want different comparisons at every site that asks either, and I have stopped trusting myself to keep them apart without a test pinning each one.

None of this changes my position on the BitChat side: a vouch that signs only a key stays true across a rename, so the trigger here still has to be detected impersonation rather than renaming, and that needs the cross-peer comparison this repo does not have. But it does strengthen your (b) rather than weaken it — the display-only fix was incomplete in the one codebase where I *did* have the detector, and it took two passes at it to actually close.

The BitChat branch is unchanged since ff39744 and still waiting on the CI gate and on your locale patch.

shubhambhandari126:
Pushed `729ca27`. Nothing behavioural — three things from auditing the revision rather than the original, and one new piece of evidence that I think is the more useful half of this comment.

**`isVerifiedAndNameBound` had no production caller, so I removed it.** @RodsF1 suggested it to collapse the double `queue.sync`, and I added it without then checking that it could actually serve the sites in question. It cannot: `showsVerifiedSeal` and `isVerifiedSender` both went to `sealAppliesToRow`, which already answers in one lock, and the two sites that *do* still take two locks — `PeerListModel` and `VerificationModel` — need the two booleans separately, because `vouchedBadge` is `!isVerified && !renamed` and the conjunction cannot express that. So it was an unused requirement on a protocol three test doubles have to satisfy. Its cases are folded into a test for what the baseline is actually kept for, which is letting a sheet name the old nickname.

**The comment on `demotedIfRenamed` still said nothing branches on `.noiseVerified` beyond the glyph.** You corrected that and I agreed in the thread, then left the wrong version sitting in the code, which is where the next person reads it. It now says what is true: no *behaviour* branches on it, and four things the user perceives do — icon, caption, accessibility description, and the `FingerprintView:87` tint, the one site that does not treat the two statuses alike.

**`MockIdentityManager` compared nicknames with `!=`.** So the double called a recase a rename while the real one does not. Which fold applies is the entire subject of this PR, so a double that has it backwards is a trap for whoever writes the next view test.

---

**The 17 XCTest-based suites run here now, and they did not before.** XCTest ships with Xcode, so on this machine they were not skipped in the sense of being excluded from a count — they could not be compiled at all, and I had not said so plainly. That included `SecureIdentityStateManagerTests`, which is the suite guarding the file this PR adds the most to. Skipping the suite that covers the code under change is the worst one to skip, and I should have gone looking for a way around it three revisions ago rather than reporting a number that quietly omitted it.

The way around it is a small XCTest surface implemented over swift-testing plus a generated `@Test` per method, which executes 210 of them. **That harness is not in this branch and never will be**, and I want to be exact about its grade: it proves the assertions execute and pass, and nothing stronger. It is not an XCTest run.

At `729ca27`, with it: **201 suites / 1975 tests pass, and every difference from unmodified `main` is an addition** — `ChatPeerIdentityCoordinatorContextTests` 8 → 10, plus the two new suites at 7 and 23, baseline 1943 → 1975. Two suites are not clean and neither is a result about this change: `NostrRelayManagerTests` has one test that times out under the shim on `main` and on this branch alike, and `VerificationServiceTests` flaked once across four full sweeps and passes 3/3 in isolation. Mutation-checked both directions: stub `liveNameMismatchLocked` to `false` and 11 binding cases fail; stub the demotion to the identity and the coordinator case fails.

I rewrote the description to match the PR as it now stands rather than as it was on day one — the corrected `.noiseVerified` claim, the two-folds table, both collision counters, #1709 for the `/fav` routing, and the display-only boundary stated as a position with its reasoning rather than left implied.

Two things are still waiting on other people and neither is something I can close by pushing at it. @jackjackbits — the workflows on this PR have never run; four commits at `action_required`, so the only real-toolchain evidence is @RodsF1's simulator run at `a74fa10`, three commits ago. And the fingerprint-sheet copy is still the `%@` form with the plumbing in and deliberately unreferenced, so CI stays green until the locales land; no hurry from me, and I would rather it sit unfinished than machine-translate a security string.

shubhambhandari126:
Pushed `ecdea99`, and it is a correction rather than an addition. `sealAppliesToRow` was wrong in both directions, and it has been through two rounds of review here with me describing that surface as finished.

It stripped a trailing `#` plus four hex from the rendered sender unconditionally. But a peer announces whatever string it likes, and `#` plus four hex is a legal thing to announce — `splitSuffix` exists precisely because announced names carry them. So:

| pinned | row shows | was | should be |
| --- | --- | --- | --- |
| `ravi#cafe` | `ravi#cafe` | seal **lost**, with no rename at all | kept |
| `ravi` | `ravi#cafe` | seal **kept**, after a rename | withheld |

The second is the one that matters, and it is a better disguise than an unrelated nickname rather than a worse one: `#cafe` is indistinguishable from the decoration this app generates itself, so the row reads as "ravi, disambiguated".

**Trying the raw name before undecorating is not enough.** I wrote that version first, ran it, and watched the second row sail through — a row string on its own genuinely cannot distinguish the list decorating `ravi` with `#a1b2` from a peer announcing the name `ravi#a1b2`. The information is not in the string.

It is in the peerID. `PeerDisplayNameResolver` builds the decoration as `#` plus the first four hex of the peerID and nothing else does, so `matchesRenderedName(_:decoratedFor:)` removes that exact suffix and no other. Both call sites already had `message.senderPeerID` in hand, so this costs a parameter and no new lookup.

The live check is untouched and must stay that way: what a peer announces has no decoration to remove. That announced-versus-rendered split is the thing I had right at the top of this PR and then quietly collapsed inside one predicate — which is what produced both rows above.

**Where it came from, since it is worth saying:** I opened the Android mirror of this patch at permissionlesstech/bitchat-android#939, and Codex flagged exactly this on the Kotlin version. I then checked whether it applied here. It did. Two human reviews and my own audit pass had not found it, and I would not have gone looking without that prompt.

Evidence, same harness as before: 25 cases in `SecureIdentityStateManagerNicknameBindingTests`, up from 23; **201 suites / 1977 tests, with every difference from unmodified `main` an addition** (`ChatPeerIdentityCoordinatorContextTests` 8 → 10, plus the two new suites at 7 and 25). Mutation-checked — restore the unconditional strip and the new case fails in both directions.

Unchanged from before and still not mine to close: CI has never run on this PR (every commit at `action_required`, @jackjackbits), and the fingerprint-sheet copy is still the `%@` form with the plumbing in and deliberately unreferenced so CI stays green until the locales land.

Chessing234 (commented):
binding on first vouch / re-verify / first announce-while-trusted looks right, and demoting `.noiseVerified` → `.noiseSecured` on rename (instead of caching the demotion) avoids a sticky false negative. fail-open on missing baseline is the painful but correct upgrade call. no new defect from me on tip.

Chessing234 (commented):
follow-up: peer list / dm / timeline seals look gated, but fingerprintview still prints the green "verified" badge from `isVerified` after a rename (icon can demote, text doesn't). `nameChangedSinceVerification` / `verifiedAsNickname` are plumbed on the presentation state and unused in the view — drive the badge/copy off that, or stop claiming verified next to the new name. the sheet is the strongest trust surface and still endorses the rename.

shubhambhandari126:
@Chessing234 — you are right, and the reasoning I gave for the half-measure was wrong. Pushed `83eb89e`.

I kept `✓ VERIFIED` on the grounds that the key genuinely *is* verified and only the glyph should demote. But that badge sits a few lines under `peerNickname`, so it is not a claim about a bare key — it is a claim about **this name**, and that claim is false. And, as you put it, this is the screen someone opens *because* a seal vanished from a row: the one surface they consult to resolve the doubt was the one endorsing the rename.

The badge, the message and the verify button now all read `isVerified && !nameChangedSinceVerification` — which is what that field was plumbed onto the presentation state for and then left unused, exactly as you said.

**It needs no new string, which removes the objection I had been making.** The existing not-verified copy already names the current nickname and offers to verify it, and that offer *is* the recovery: re-verifying re-pins the baseline to the name on screen. The stored verification is untouched underneath, so nothing is lost.

One consequence I would rather state than have found: a renamed peer's sheet now offers "mark verified" instead of "remove verification", so unverify is one re-verification away rather than immediately to hand. I think that is the right trade against a green checkmark beside a name nobody verified, but it is a real change to the affordance and worth arguing with if you read it differently.

That leaves the locale question much smaller than I had it. The `%@` form @RodsF1 offered — *"you verified this key as `%@`. it now announces a different name."* — is still strictly better than reusing the not-verified copy, because naming the old nickname is what lets someone say "oh, that is ravi". But it is now an improvement rather than the thing standing between this PR and being correct, and it can land whenever the locales do.

26 cases in the binding suite; 201 suites / 1977 tests, every difference from unmodified `main` an addition.

Thanks for the read — your first note also confirmed the two calls I was least sure about (computing the demotion on read rather than caching it, and failing open on a missing baseline), which is genuinely useful to hear from someone else.

CI has still never run on this PR, @jackjackbits.

shubhambhandari126:
Pushed `e2e929b`. This one came out of a change in how I was looking, so the method is worth more than the diff.

Every fix on this branch so far arrived the same way: someone found a place the seal survived a rename, I fixed that place. That is a loop that ends when reviewers get tired, not when the bug class is gone. So I went the other way — enumerated every site in the app that *draws* a trust seal, working backwards from the drawing code to whatever state gates it, and asked of each one whether that state is bound to a name you actually verified.

The glyph sites are all clean, and each traces to gated state:

| site | gated by |
|---|---|
| `MeshPeerList.swift:133` | `showsVerifiedBadgeWhenOffline` |
| `MeshPeerList.swift:149` | `showsVouchedBadge` |
| `FingerprintView.swift:195` | `isVouched`, zeroed via `voucherCount` |
| `TextMessageView.swift:54`, `MediaMessageView.swift:52` | `conversationUIModel.showsVerifiedSeal` |
| `NoiseEncryptionService.swift:112` | demoted in `getEncryptionStatus` |
| `ChatMessageFormatter.swift:467` | in-string `✓`, same source |
| `AppInfoView.swift:182` | static legend, no peer |

One site was not clean, and it is not a glyph. `VerificationModel` built the fingerprint sheet's voucher list from `social.claimedNickname` — the voucher's *current* name. That list is a trust attribution one level out: "people you verified vouch for this peer". So Eve is verified as `ravi`, vouches for Mallory, renames to `medic`, and Mallory's sheet reads **vouched by medic**. The vouch is real and Eve's key is genuinely trusted; the *name* printed beside it was never verified by anyone. Same defect as the seal, one indirection removed, which is exactly why looking only at glyph sites would never have found it.

Fixed to prefer a local petname, then the pinned baseline, then the claimed nickname. Petname first because it is your own label and nothing on the network can touch it; claimed nickname last as the same fail-open used everywhere else here, for vouchers recorded before baselines existed.

On the test, plainly: the new case covers the *prerequisite* — that the baseline outlives the rename, so the sheet has `ravi` to show — not the view, which needs a `ChatViewModel` I cannot build outside Xcode. Reverting the `VerificationModel` change alone would not fail it. I would rather say that than let it read as coverage it does not have.

I ran the same pass over bitchat-android (#939) and it came back clean: `PeerAvatar.kt:120`, `SecurityVerificationSheet.kt:217/303`, `MeshPeerListSheet.kt:1958` gated at `:1801`, and five wear screens — four fixed in `077c82b`, and `PeopleScreen.kt:229` already reactive because it reads `snapshot` with no `remember` at all.

Full suite still green and additive against unmodified `main`: 237 suites, one process each, the only differences being `SecureIdentityStateManagerNicknameBindingTests` +27, `PeerDisplayNameResolverTests` +7, `ChatPeerIdentityCoordinatorContextTests` 8 → 10.

Separately, and not urgent: this branch has **zero check runs** across all eight commits — workflows on a first-time contributor's PR need a maintainer to approve them. @jackjackbits, one click would let CI say something about this rather than only me.

PR #1713, opened 2026-09-16
fix(transport): reconcile decompressed payload ceiling with Android (#1629)

### LINKED ISSUE #1628: Log silent rejection of oversized compressed payloads instead of dropping with no trace

### LINKED ISSUE #1629: iOS/Android disagree on max expanded (decompressed) payload size — 1.13 MiB vs 10 MiB, silent drop on iOS

vinayak509143:
### Summary
Addresses cross-platform media reception failures (#1518) by reconciling the iOS decompressed payload ceiling with Android (#1629) and logging previously silent compressed-payload rejects (#1628).

### Changes
1. **Payload Ceiling Parity:** Added `FileTransferLimits.maxExpandedPayloadBytes = 10_485_760` (10 MiB) matching Android's `AppConstants.Protocol.MAX_PAYLOAD_LENGTH`.
2. **Reconciled Bounds:** Updated `payloadLength` and `originalSize` in `BinaryProtocol.decodeCore` and BLE reassembly limits for `.fileTransfer` / `.noiseEncrypted` to use the 10 MiB ceiling.
3. **Structured Diagnostics (#1628):** `SecureLogger.warning(..., category: .security)` on compressed-path rejects (ceiling, truncated body, inflate/size mismatch) with `originalSize`, ceiling, `compressedSize`/`decompressedSize`, `codec=zlib`, and 16-hex wire peer ID. No payload bytes, no peer NACK, no UI.
4. **Defensive Allocation:** `CompressionUtil.decompress` refuses `originalSize` above the ceiling **before allocate**, and logs the bound refusal (zip-bomb / OOM guard).
5. **Unit Tests:** `BinaryProtocolTests` and `BLEFragmentAssemblyBufferTests` cover in-bound ~1.2 MiB accept, 10 MiB+ reject, fragment eviction, and DEBUG warning capture via `SecureLogger.testWarningSink`.

Closes #1629
Closes #1628
Addresses #1518


vinayak509143:
Closing. #1719 landed per-type payload caps and asked not to raise the global decompressed ceiling on top of that. This branch conflicts with those changes in BinaryProtocol.swift and BLEFragmentAssemblyBuffer.swift.

